PR #38, opened 2017-06-09
Navigation structure updates

subsymbolic:
Reviewer: Caine
Asana: https://app.asana.com/0/276208855376242/290598814185289

**Description**:
Updates app to use new navigational structure:
- simplifies home screen and adds animation
- consolidates separate omnibar designs into one
- adds browsing menu to omni bar & moves settings into menu
- updates tab switcher to new design
- updates app structure so that the "home" screen is no longer a tab

**Steps to test this PR**:
- Run the app
- Note thae home screen UI changes. In particular test activating and dismissing the search bar
- Check that a new menu appears in the browser view and that the refresh, copy, bookmark and settings options all work as expected
- Note the tab switcher UI changes. Check that tabs can be added, and individual tab can be closed and that the fire button now clears everything but remains on the switcher

###### Reviewer Checklist:

###### PR DRI Checklist:


DuckDuckGo/MainViewController.swift
```diff
@@ -17,10 +17,11 @@ class MainViewController: UIViewController {
     @IBOutlet weak var shareButton: UIBarButtonItem!
     @IBOutlet weak var backButton: UIBarButtonItem!
     @IBOutlet weak var forwardButton: UIBarButtonItem!
-    weak var omniBar: OmniBar?
+    weak var omniBar: OmniBar!
```
nilnilnil: What's the difference here?

subsymbolic: The `!` force unwraps an optional. The omniBar has to be declared as an optional/nullable variable because it isn't instantiated during construction. However it is instantiated during `viewDidLoad` and as it is now never touched again, it makes sense to force unwrap here rather than unwrap manually everywhere else. Force unwrapping is a bit of an anti-pattern but common in this case - you'll see it a lot for ViewControllers where items are instantiated by InterfaceBuilder or in the viewDidLoad phase.

subsymbolic:
Thanks for the review Caine! Apologies if it was a convoluted one. With the navigation changing so much a bunch of concepts no longer made sense and items needed to be renamed to better represent their new behaviour.

PR #70, opened 2017-07-22
Navigation updates

subsymbolic:
Reviewer: Caine
Asana: https://app.asana.com/0/361428290920652/384289442884654

**Description**:
• Adds the fire button to the browsing screen
• Renames the "Bookmark" option to "Add to Bookmarks"
• Adds a "New Tab" option to 3-dot menu
• Adds a new toolbar to the home and browsing screens to allow different button sets
• Moves "Settings" button
• Adds bookmark button to home omni bar
• Adds state machine to manage various state/button combinations for the omnibar
• Removes "Copy" option from 3-dot menu

**Steps to test this PR**:
Run app and check that the new navigation structure.

###### Reviewer Checklist:

###### PR DRI Checklist:


Core/WKWebViewExtension.swift
```diff
@@ -67,6 +67,7 @@ extension WKWebView {
         dataStore.fetchDataRecords(ofTypes: allData) { records in
             let externalRecords = records.filter { $0.displayName != Constants.cacheKey }
```
nilnilnil: This is what I was saying could probably become a method of some sort to take in a non const.

DuckDuckGo/Localizable.strings
```diff
@@ -34,15 +34,17 @@
 "action.title.pasteAndGo" = "Paste & Go";
 "action.title.save" = "Save";
 "action.title.cancel" = "Cancel";
-"action.title.newTab" = "Open in New Tab";
+"action.title.newTab" = "New Tab";
```
nilnilnil: Why is this file not enough to translate?

DuckDuckGo/OmniBar.swift
```diff
@@ -171,37 +183,48 @@ class OmniBar: UIView {
...
         omniDelegate?.onContenBlockerPressed()
     }
 }
 
 extension OmniBar: UITextFieldDelegate {
     
     func textFieldDidBeginEditing(_ textField: UITextField) {
-        onEditingMode()
+        refreshState(state.onEditingStartedState)
+        DispatchQueue.main.async {
```
nilnilnil: Why is this construct necessary?

DuckDuckGo/OmniBarState.swift
```diff
@@ -0,0 +1,127 @@
...
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+
+import Foundation
+import Core
+
+protocol OmniBarState {
```
nilnilnil: nice

DuckDuckGo/TabSwitcherViewController.swift
```diff
@@ -114,12 +111,12 @@ class TabSwitcherViewController: UIViewController {
         delegate.tabSwitcher(self, didRemoveTabAt: index)
         collectionView.reloadData()
         refreshTitle()
-    }
-    
-    fileprivate func dismiss() {
         if tabsModel.isEmpty {
             delegate.tabSwitcherDidRequestClearAll(tabSwitcher: self)
         }
+    }
```
nilnilnil: This was a confusing part of the diff to read at first; but I think I get it now. You want this clear all called on onClear, and dismiss just does dismiss.

DuckDuckGo/TabSwitcherViewController.swift
```diff
@@ -114,12 +111,12 @@ class TabSwitcherViewController: UIViewController {
...
+    
+    fileprivate func dismiss() {
```
nilnilnil: Why this? Seems like a loss in explicitness. Just for terseness?

Core/WKWebViewExtension.swift
subsymbolic: Definitely. We'd pass in a array of items to blow away based on top level domains we hit during a page load. All webkit exposes for matching is this `displayName` so it probably won't be a perfect match, but we'll hopefully catch most items added to the cache.

DuckDuckGo/Localizable.strings
subsymbolic: The strings in Interface Builder use a different system and aren't yet extracted.

DuckDuckGo/OmniBar.swift
subsymbolic: Do you mean the async? We're in the middle of starting editing and the ui stutters if we block to highlight.

DuckDuckGo/TabSwitcherViewController.swift
subsymbolic: Looks like you answered this in your other comment... Let me know if you want further clarification.

PR #78, opened 2017-07-27
Allow user to forget an individual page

subsymbolic:
Reviewer: Caine or Chris
Asana: https://app.asana.com/0/361428290920652/391002287401934 & https://app.asana.com/0/361428290920652/388642735958849

**Description**:
First pass of the "forget this page" functionality. In terms of the cache it clears some things but it has to guess the key so there are a lot of misses as the domains visited and the cache keys don't always match. No surprise really - we expected this.

I'll submit this for now but will iterate soon with more monitoring. My feeling is that if I natively spy on the http cookie headers and use javascript to interrogate the document cookies I should catch session domains and be able to clear more from the cache.

Interestingly, this is supported natively in iOS 11 where WKHTTPCookieStore and WKHTTPCookieStoreObserver are added https://developer.apple.com/documentation/webkit/wkwebsitedatastore.

**Steps to test this PR**:
1. Run app
2. Browse a page
3. Hit the fire button and note that the "Close this tab" now reads "Forget this page"
4. Note that hitting this button closes the current page and returns to the home page

###### Reviewer Checklist:

###### PR DRI Checklist:


Core/WebCacheManager.swift
```diff
@@ -0,0 +1,81 @@
...
+                completionHandler()
+            }
+        }
+    }
+    
+    /**
+     Clears the cache of external (non-duckduckgo) data belonging to provided hosts
+     - parameter hosts: the hosts to clear from the cache
+     */
+    public static func clear(forHosts hosts: [String], completionHandler: @escaping () -> Swift.Void) {
```
subsymbolic: I initially combined this with the method above (with default hosts=nil ) but I removed it as I felt that it detracted from readability and I want these methods to be really clear to anyone reading/updating them.

PR #86, opened 2017-07-31
Migrate legacy saved searches and favourites

brindy:
Reviewer: Mia
Asana: https://app.asana.com/0/392891325557410/390989879439961
CC: Caine

**Description**:
Migrates the users saved searches and stories from the Search and Stories app in to bookmarks in the new app. 

Tested with tagged version of 5.0.1 (delete the mogenerator target and dependency, comment out references to test flight, and update the bundle id to match the new app)

Tested with tagged version of 6.0.2 (update the bundle id to match the new app)

Tested with tagged version of 6.0.9 (update the bundle id to match the new app)

**Steps to test this PR**:
1. Checkout a version of the Search and Stories app to test against, as in the description above
2. Launch the Search and Stories app from Xcode 
3. Favourite one or more stories and make a note of what they are
4. Favourite one or more searches and make a note of what they are
5. Launch the new app from Xcode
6. Finish the on-boarding screens and check the favourited stories and searches are present in the bookmarks

###### Reviewer Checklist:

###### PR DRI Checklist:


DuckDuckGoTests/Migration/MigrationTests.swift
```diff
@@ -0,0 +1,225 @@
...
+            XCTAssertEqual("http://www.example.com", link.url.absoluteString)
+            XCTAssertNil(link.favicon)
+            
+            expectation.fulfill()
+        }
+        
+        wait(for: [expectation], timeout: 1)
+    }
+    
+    func testFavouriteStoriesDeletedAfterMigration() {
```
subsymbolic: This test fails if the app has bookmarks when it is run.

DuckDuckGo/Migration/Migration.swift
```diff
@@ -0,0 +1,72 @@
...
+            bookmarks.save(bookmark: Link(title: title, url: url, favicon: nil))
+            bookmarkCount += 1
+        }
+        
+        defaults.removeObject(forKey: Migration.oldBookmarksKey)
+        return bookmarkCount
+    }
+    
+    private func migrateStories(into bookmarks: BookmarksManager) -> Int {
+        let savedStories = self.container.savedStories()
```
brindy: Agreed.  That's because I extracted it in to a private method as I was TDDing.

DuckDuckGo/AppDelegate.swift
```diff
@@ -82,13 +82,22 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     }
     
     private func startOnboardingFlowIfNotSeenBefore() {
+        
         var settings = TutorialSettings()
         if !settings.hasSeenOnboarding {
+            startMigration()
```
brindy: Suggestions on where to put it?  I figured on-boarding was good because it would be complete by the time the user had got through it and already has logic for only doing this once.

subsymbolic: Similar to what you have now - just make a separate method to handle migration logic and call it before the `startOnboardingFlowIfNotSeenBefore()` method rather than within.

brindy: It should be a pretty quick operation, but my worry is that by decoupling it from the on-boarding there may be edge cases where the migration isn't finished by the time the user gets to their bookmarks.  Do we need to manage the UI/UX in this case, or are we OK with that potential edge case?

subsymbolic: I'm 80% sure that the onboaridng is getting deleted. From what I saw the migration was pretty fast so I'm ok with deferring any migration related ui (if you are too!) until we know we need it. Only caveat is that the migration needs to be non-destructive to any new bookmarks created while it is running.

subsymbolic: If we find we need to a UI we can show a simple banner telling the user that migration is in progress. There's hopefully no need for the migration to have any impact on their enjoyment of the app.

brindy: It deletes the old searches and favourites so that they don't get migrated more than once, but doesn't do anything to the bookmarks in the new app other than adding to them.

DuckDuckGoTests/Migration/MigrationTests.swift
brindy: It now clears existing book marks as part of the test setup - hopefully safe to assume people with bookmarks won't be running unit tests against their actual boomarks. :)

brindy:
@subsymbolic have updated PR to address comments :)

subsymbolic:
Looks good!

PR #92, opened 2017-08-05
Remove static disconnect list from project

subsymbolic:
Reviewer: Chris
Asana: https://app.asana.com/0/361428290920652/375879998104269

**Description**:
• removes disconnect list from the project, now downlods it from the server instead
• adds error handling flow to cover rare situation where content blocker is opened before trackers list have ever been downloaded

**Steps to test this PR**:

Ensure that error page is shown if content blocker not yet retrieved:
1. Turn on aeroplane mode and freshly install app
2. Wait for it to launch
3. Turn network back on and search for something
4. Open content blocker, it should show an error page
5. Press the refresh button to retrieve the trackers list and show standard panel

Ensure content blocker continues to work as expected:
1. Launch app
2. Navigate to a page with trackers, ensure tracker number is updated

###### Reviewer Checklist:

###### PR DRI Checklist:


DuckDuckGo/AppDelegate.swift
```diff
@@ -45,6 +45,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     
     func applicationDidBecomeActive(_ application: UIApplication) {
         startMigration()
+        TrackerLoader.shared.updateTrackers()
```
brindy: Note that this causes the trackers request to happen every time the application becomes active.  That means, for instance, if the user swipes up/away the control panel or notifications.  

That said, it looks like Alamofire uses the default URLCache configuration, but it's not entirely clear to me what that is.  Do we need to be more explicit about when the tracker data is downloaded to prevent too much data usage or, conversely, getting out of date by using a cached version?

brindy: Also note that this means it's difficult to test on a device where swiping up to turn off Airplane mode and then swiping away means the trackers are automatically downloaded.  Easier to test on the simulator though by putting your entire Mac in Airplane mode.

DuckDuckGo/ContentBlockerErrorDelegate.swift
```diff
@@ -0,0 +1,15 @@
+//
```
brindy: Non-standard copyright block?

DuckDuckGoTests/DisconnectMeRequestTest.swift
```diff
@@ -0,0 +1,132 @@
+//
+//  DisconnectMeRequestTests.swift
```
brindy: Non-standard copyright block

DuckDuckGo/ContentBlockerErrorViewController.swift
```diff
@@ -0,0 +1,62 @@
+//
+//  ContentBlockerRelaoder.swift
```
brindy: Wrong file name

DuckDuckGo/ContentBlockerErrorDelegate.swift
```diff
@@ -0,0 +1,15 @@
...
+//  ContentBlockerErrorViewControllerDelegate.swift
```
brindy: Wrong file name

Core/SFContentBlockerManagerExtension.swift
```diff
@@ -0,0 +1,38 @@
+//
+//  SFContentBlockerManagerExtension.swift
```
brindy: wrong file name

DuckDuckGo/ContentBlockerViewController.swift
```diff
@@ -0,0 +1,88 @@
+//
+//  ContentBlockerPopover.swift
```
brindy: wrong file name

Core/SFContentBlockerManagerExtension.swift
subsymbolic: This one looks ok to me

brindy: ah sorry, it's because the name of the extension is different, i.e. `SFContentBlockerManager`. is that deliberate?

DuckDuckGo/AppDelegate.swift
subsymbolic: > Do we need to be more explicit about when the tracker data is downloaded to prevent too much data usage or, conversely, getting out of date by using a cached version?
We're using a standard etag model for caching so it _should_ be ok. Have you noticed data being re-downloaded?

Core/SFContentBlockerManagerExtension.swift
```diff
@@ -0,0 +1,38 @@
...
+    }
+    
+    public static func reloadContentBlocker() {
+        reloadContentBlocker(withIdentifier: Constants.identifier) { error in
+            if let error = error {
+                Logger.log(text: "Could not reload content blocker in Safari due to \(error)")
+                return
+            }
+            Logger.log(text: "Content blocker rules for Safari reloaded")
+        }
```
brindy: This only ever seems to display the error condition for me when I run it: 

> Could not reload content blocker in Safari due to Error Domain=SFErrorDomain Code=1 "(null)"

DuckDuckGo/ContentBlockerErrorDelegate.swift
subsymbolic: Thanks!

DuckDuckGo/AppDelegate.swift
brindy: Cool, OK - that should be fine then - Alamofire handles that automatically?  

Well, it always seems fast for me, so I wasn't sure.  I'll watch it a bit closer and see what's going on, especially if there's a difference between the first and subsequent requests.

brindy: Just checked with Networking monitoring, first request uses 88kb of network, subsequent uses 0.5kb.

Core/SFContentBlockerManagerExtension.swift
subsymbolic: Yes, it is deliberate, I use that pattern for generic extensions.

Core/SFContentBlockerManagerExtension.swift
subsymbolic: I find the content blocking extension a real pain to debug. But if I open the safari dev console and connect it to the simulator's safari session I can see that the content blocker is working as expected.
safari_console

subsymbolic:
@brindy, thanks for the review. I *think* I've covered all your comment (I've ignored a couple of duplicates). Ready for rereview!

PR #96, opened 2017-08-09
localize attributed strings

brindy:
Reviewer: @subsymbolic 
Asana: https://app.asana.com/0/392891325557410/399963741350587
CC:

**Description**:

Convert attributed strings to plain labels.  

Unfortunately, they have to be plain labels because if they are attributed strings in interface builder and have a trait variant, they can't be updated from code (or I have not been able to work out how to do that).  This means they may get translated a couple of times, but it will be the text in the localisation strings file that takes priority. 

The instructions strings couldn't simply use HTML because when constructing an attributed string object with document type HTML, it overrides any style applied in interface builder resulting in a very bad/basic looking piece of text.  While it would be possible to be explicit about the format using HTML this shouldn't be passed to translators as it is likely to cause confusion/errors.  Instead, SwiftRichString library has been added.  This has a range of useful utilities for building attributed strings, in particular being able to define your own markup to which only given styles are applied to text that is wrapped in the defined markup, thus retaining any styles applied in IB.  For the instructions text I have defined `"<highlight />"` which simply makes the text color of the wrapped text #333.

Also note that having the localisation strings in core (requiring the use of NSLocalizationString initialiser which took a bundle) broke the localization export so these strings have had to be duplicated in the app and the share extension.

**Steps to test this PR**:
1. Launch the app
2. Navigate to the about screen, ensure the "What your search for your is your business statement" looks correct
3. Navigate to the Safari Search Instructions, ensure the text looks correct
4. Change language to Dutch and repeat the above.  For the Safari Search instructions the English will still be visible, but prefixed with a dutch flag to highlight the localisation is working and that they require translation
5. Retest the above on both a phone and tablet form-factor device to ensure that traits are applied correctly (i.e. text is slightly larger on the tablet, though it might not be immediately obvious as it's only 4pts different)
6. From Xcode it should be possible to successfully export the translations to xliff. (Note that markup will be encoded in the xliff file, i.e. < and > will be &lt; and &gt; respectively but fear not, they re-import correctly)

###### Reviewer Checklist:

###### PR DRI Checklist:


DuckDuckGo/en.lproj/Localizable.strings
```diff
@@ -57,3 +58,7 @@
 "feedbackemail.body" = "I'm running %@ on an %@ %@. Here's my feedback:\n";
 
 "navigation.title.edit" = "Edit";
+
+"safari.instructions.settings" = "Open <highlight>App Settings</highlight>";
```
subsymbolic: "App Settings" should be "Settings App"

DuckDuckGo/SafariSearchInstructionsViewController.swift
```diff
@@ -70,8 +88,12 @@ class SafariSearchInstructionsViewController: UIViewController {
         let minimumMargin = SafariSearchInstructionsViewController.minimumTopMargin
         topMarginConstraint.constant = marginForVerticalCentering > minimumMargin ? marginForVerticalCentering : minimumMargin
     }
-    
-    @IBAction func onDonePressed(_ sender: UIButton) {
-        dismiss(animated: true, completion: nil)
+
+    private func makeStyle() -> Style {
```
subsymbolic: Really cool library, this is a great solution 👍

DuckDuckGo/SafariSearchInstructionsViewController.swift
```diff
@@ -70,8 +88,12 @@ class SafariSearchInstructionsViewController: UIViewController {
...
+        return Style("highlight", {
+            $0.color = UIColor(hex: "#333333", alpha: 1.0)
+        })
```
subsymbolic: Totally optional style thing: consider moving the closure out of the brace
```
return Style("highlight") {
   $0.color = UIColor(hex: "#333333", alpha: 1.0)
 }
```

brindy: Agreed - I prefer this style myself. Probably because I copied and changed the example from the docs.

brindy:
Yep, I was planning to. Think I'll do that by default moving forward.

PR #103, opened 2017-08-14
Add a simple in-memory cache to maintain site score during navigation

subsymbolic:
Reviewer: Chris ot Caine

**Description**:
When the user is navigating backwards and forwards a full page load may not occur meaning that the site score may be artificially low. The cache allows us to return a better score

**Steps to test this PR**:
1. Browse to a page with bad site score e.g the daily mail
2. Browse to a another page with a better rating e.g duckduckgo
3. Navigate back (by swiping or with the back button). Ensure the site score stays the same

###### Reviewer Checklist:

###### PR DRI Checklist:

### DIFF
--- a/Core/SiteRating.swift
+++ b/Core/SiteRating.swift
@@ -39,7 +39,7 @@ public struct SiteRating {
         score += httpsScore
         score += trackerCountScore
         score += majorTrackerNetworkScore
-        return score
+        return SiteRatingCache.shared.register(domain: domain, score: score)
     }
     
     public var https: Bool {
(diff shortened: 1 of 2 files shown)


Core/SiteRating.swift
```diff
@@ -71,3 +71,30 @@ public struct SiteRating {
...
+    
+    private var cachedScores = [String: Int]()
+
+    
+    /**
+     Registers a site score for in-memory caching. Only updates the score
+     in the cache if the new score is higher.
+     - returns: The score maintaned in the cache
+     */
+    func register(domain: String, score current: Int) -> Int {
```
subsymbolic: I agree. I wonder if "update" works better as it's not uncommon to have the "result" of an update returned. In truth I'm struggling to name it. Let me know if you have any ideas! If not, it's  probably a sign that the method is doing too much and I should split it out

nilnilnil: You could break it into 2 things, update and get, or just update.

subsymbolic: I've gone with the former, let me know if that reads better now. If so, I'll squash the changes (rather than merge) them in

PR #111, opened 2017-08-17
suppress blank nav

brindy:
Reviewer: Mia
Asana: https://app.asana.com/0/0/402612726718857/f
CC:

**Description**:

After some investigation, there appears to be no official way to detect apps being launched as a result of universal links.

I've worked around this by observing a policy decision request (which still comes through for the destination URL).  If the app is backgrounded before any of the delegate methods are called (i.e. the web view has not processed the URL), then it assumes an app has been launched as a result of a universal link and tells the view to 'go back' to the previous page.

**Steps to test this PR**:

Test 1 - launch app via universal link
1.  Install twitter
2. Open the DuckDuckGo app and search for something that will give you a twitter links
3. Tap on a twitter link
4. Twitter will launch (unless you've previously suppressed this!)
5. When you return to the app it is on a blank page

Test 2 - standard browsing
1. Search for something that will not open in Twitter
2. Tap on a link
3. Observe the page loading in the app
4. Navigate to another app via pressing the home button
5. Return to the DuckDuckGo app and observe the same page is still displayed

###### Reviewer Checklist:

###### PR DRI Checklist:

### DIFF
--- a/Core/WebViewController.swift
+++ b/Core/WebViewController.swift
@@ -21,15 +21,21 @@ import UIKit
 import WebKit
 
 open class WebViewController: UIViewController, WKNavigationDelegate, WKUIDelegate {
-    
+
     private static let estimatedProgressKeyPath = "estimatedProgress"
-    
+
     public weak var webEventsDelegate: WebEventsDelegate?
     
     @IBOutlet weak var progressBar: UIProgressView!
     @IBOutlet weak var errorMessage: UILabel!
     
     open private(set) var webView: WKWebView!
+    private var waitingForLoadAfterPolicyAllowDecision = false
+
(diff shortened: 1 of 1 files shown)


subsymbolic:
Great work finding a workaround for this. I had assumed it was a issue with our code but I can see from your investigation that it's really an ios bug. I think the bug  is annoying but not really that bad and I'm wondering if we should just leave it until it gets fixed by apple rather than putting in a workaround that we're unhappy with. What do you think?

brindy:
Given this isn't a crashing issue and is likely to be fixed in iOS in the future, I would take the hardline of less code is better, so let's park it for now. I'll move it to a spike branch.

PR #112, opened 2017-08-17
Update animated fire prototype

subsymbolic:
Reviewer: Chris
Asana: https://app.asana.com/0/361428290920652/275938474052467

**Description**:
Replace previous static prototype animation with latest moving prototype.
NOTE: This this is NOT the final animation. We are waiting on final assets from the animator and need to implement a proper UIViewController transition.

**Steps to test this PR**:
1. Press the fire button
2. Ensure latest animation is displayed and fills the width of the screen irrespective of device

###### Reviewer Checklist:

###### PR DRI Checklist:


DuckDuckGo/FireAnimation.swift
```diff
@@ -30,22 +35,53 @@ struct FireAnimation {
         }
         
         let animationContainer = UIView(frame: window.frame)
```
brindy: It looks great, but if you rotate while it's in flight it doesn't adjust size.

If you add ` animationContainer.autoresizingMask = .flexibleWidth` here and the same for the fiewView it will then resize automatically on rotation.

subsymbolic: Good thinking!

PR #119, opened 2017-08-24
Feature/urls treated as search

brindy:
Reviewer: Mia
Asana: https://app.asana.com/0/414235014887631/312386120368203/f
CC:

**Description**:

Better handle determining if user entry is really a URL, or a search term.

**Steps to test this PR**:
1. Paste this link in to omni bar - http://www.veganchic.com/products/Camo-High-Top-Sneaker-by-The-Critical-Slide-Societ+80758-0180.html - it should load the page directly
2. Navigate to a domain using a user in the URL, e.g. example.com@samples.com - this should perform a search instead
3. Navigate to URLs using just the domain name
4. Navigate to URLs using the valid web schemes and a domain name
5. Perform regular searches

###### Reviewer Checklist:

###### PR DRI Checklist:

### DIFF
--- a/DuckDuckGoTests/URLExtensionTests.swift
+++ b/DuckDuckGoTests/URLExtensionTests.swift
@@ -21,7 +21,15 @@
 import XCTest
 
 class URLExtensionTests: XCTestCase {
-    
+
+    func testUrlWithUserIsFalse() {
+        XCTAssertFalse(URL.isWebUrl(text: "http://example.com@sample.com"))
+    }
+
+    func testBug1() {
+        XCTAssertTrue(URL.isWebUrl(text: "http://www.veganchic.com/products/Camo-High-Top-Sneaker-by-The-Critical-Slide-Societ+80758-0180.html"))
+    }
+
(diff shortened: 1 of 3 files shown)


DuckDuckGoTests/URLExtensionTests.swift
```diff
@@ -21,7 +21,15 @@
 import XCTest
 
 class URLExtensionTests: XCTestCase {
-    
+
+    func testUrlWithUserIsFalse() {
```
subsymbolic: How do you feel about the `When<condition>Then<expectedResult>` syntax for test names? I use it a fair bit - but not all the time as it doesn't always flow

brindy:
You're right, @subsymbolic, I should follow the established convention, sorry.

DuckDuckGoTests/URLExtensionTests.swift
brindy: Yep, I'm good with this. Will endeavour to do this automatically moving forward.

PR #148, opened 2017-09-16
Add tos to site score

subsymbolic:
Reviewer: Chris
Asana: https://app.asana.com/0/414235014887631/426432715344337

**Description**:
Adds TOS to the privacy grade

**Steps to test this PR**:
1. Visit amazon.com and notice the site score logs now contain an entry `tos: d_1` meaning the amazon has a terms score of d and this has added a point to the site score
2. Visit disqus.com and notice the site score logs now contain an entry `tos: a_-1`

NOTE: I've followed the approach used by extensions but this performs matching by dropping the url's tld, which isn't perfect. I've opened a task to discuss and update this separately https://app.asana.com/0/414235014887631/430675284089081.

###### Reviewer Checklist:

###### PR DRI Checklist:


Core/FileLoader.swift
```diff
@@ -27,15 +27,22 @@ enum FileError: Error {
 
 class FileLoader {
 
-    func load(name: String, ext: String) throws -> Data {
+    func load(fileName: String) throws -> Data {
```
brindy: This function isn't used, and is a little unclear about where files are loaded from as it uses the bundle this class was loaded from (i.e. Core). Delete it?

Core/FileLoader.swift
```diff
@@ -27,15 +27,22 @@ enum FileError: Error {
...
     
-    func load(bundle: Bundle, name: String, ext: String) throws -> Data {
-        guard let path = bundle.path(forResource: name, ofType: ext) else { throw  FileError.unknownFile }
-        guard let string = try? String(contentsOfFile: path) else { throw FileError.invalidFileContents }
-        guard let data = string.data(using: .utf16) else { throw FileError.invalidFileContents }
+    func load(fileName: String, fromBundle bundle: Bundle) throws -> Data {
+        
+        let fileUrl = URL(fileURLWithPath: fileName)
+        let baseName = fileUrl.deletingPathExtension().path
+        let ext = fileUrl.pathExtension
```
brindy: Seems excessive when you could just pass the extension in, but I understand the motivation.

Core/TermsOfService.swift
```diff
@@ -0,0 +1,50 @@
...
+//
+
+
+import Foundation
+
+
+public struct TermsOfService {
+
+    let classification: Classification
+    let goodTerms: [String]
```
brindy: Unused outside of tests - delete?

Core/TermsOfService.swift
```diff
@@ -0,0 +1,50 @@
...
+    let badTerms: [String]
```
brindy: Unused outside of tests - delete?

Core/FileLoader.swift
subsymbolic: It is used by the TermsOfServiceStore to load tosdr.json. I can rename is BundleFileLoader and make the bundle an explicit argument

brindy: `TermsOfServiceStore` uses the one that takes a bundle as a parameter.

Core/FileLoader.swift
subsymbolic: Caine gave me the opposite feedback - I sat on it for ages and finally put it in :-)

brindy: I'm not that fussed, but passing the extension is no surprise to an iOS dev and having to construct a URL just to get the extension (which is probably slightly less offensive than using a regex ;) ) still fees a bit OTT to me :)

Happy for it to stay.

Core/FileLoader.swift
subsymbolic: Good point! 🔥

PR #149, opened 2017-09-18
Fix score reporting

subsymbolic:
Reviewer: Chris
Asana: https://app.asana.com/0/414709148257752/420099613930289

**Description**:
This PR updates the cache to only match if the schemes are the same (as we score https better than http) and ensures the site rating is only updated for the current tab

**Steps to test this PR**:
TEST 1:
1. Visit http://duckduckgo.com
2. Note the url becomes https://duckduckgo.com
3. Ensure the site rating is still an A

TEST 2:
1. Open two tabs with very different ratings e.g http://duckduckgo.com which is an A and http://www.dailymail.co.uk/ which is a D
2. While the "D" rating is still loading, switch back to the "A" rated one
3. Ensure the rating stays an "A" during loading
4. When all loading is done flip between the tabs and ensure the rating is updated appropriately

###### Reviewer Checklist:

###### PR DRI Checklist:

### DIFF
--- a/Core/SiteRatingScoreExtension.swift
+++ b/Core/SiteRatingScoreExtension.swift
@@ -126,10 +126,10 @@ public class SiteRatingCache {
     }
     
     private func cacheKey(forUrl url: URL) -> String {
-        let scheme = url.scheme ?? URL.URLProtocol.http.rawValue
         guard let domain = url.host else {
             return url.absoluteString
         }
+        let scheme = url.scheme ?? URL.URLProtocol.http.rawValue
         return "\(scheme)_\(domain)"
     }
 }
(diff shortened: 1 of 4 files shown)


Core/SiteRatingScoreExtension.swift
```diff
@@ -103,24 +103,33 @@ public class SiteRatingCache {
...
+        let key = cacheKey(forUrl: url)
+        return cachedScores[key]
     }
     
     func reset() {
         cachedScores =  [String: Int]()
     }
     
+    private func cacheKey(forUrl url: URL) -> String {
+        let scheme = url.scheme ?? URL.URLProtocol.http.rawValue
```
brindy: scheme is unused until after the guard - move the guard up?

PR #154, opened 2017-09-20
first party url logic

brindy:
Reviewer: Mia
Asana: https://app.asana.com/0/392891325557410/433193852567931
CC:

**Description**:

* Use TLDs to determine if the domains of two URLs match in order to check that a URL is loaded from the same domain or not.

* Updated the logic so that all URLs are processed every time (instead of caching what happened to them previously.

* Load the scripts even if content blocking is disabled in settings.

**Steps to test this PR**:

Test 1 - First Party URLs
1. Go to criteo.com and ensure the site works as expected

Test 2 - blocking setting
1. Disable content blocking in settings
2. Visit a tracker heavy site
3. Ensure tracker detection messages are being received by the app

###### Reviewer Checklist:

###### PR DRI Checklist:


Core/tlds.js
```diff
@@ -0,0 +1,7921 @@
...
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+// Based on https://github.com/131/node-tld
+var duckduckgoTLDParser = function() {
+
+	var tlds = {
```
subsymbolic: I'm a bit worried about updating this for two reasons:
• it's been modified so we can't just drop in a new version to update it
• we have another copy of the list (in the official format) in the app. It would be easy to update one file and forget the other

Can we use the tld file already in the codebase and parse it into the format you need for the js layer?

subsymbolic: Happy for to be capture this in a separate Asana task if this isn't a good time to tackle it.

Core/contentblocker.js
```diff
@@ -129,10 +123,15 @@ var duckduckgoContentBlocking = function() {
 	}
 
 	// private
-	function shouldSkip(event) {
+	function isFirstParty(event) {
 		var topLevelUrl = getTopLevelURL()	
 		var url = toURL(event.url, topLevelUrl.protocol)
-		return url != null && (isDuckDuckGo(url) || isFirstPartyRequest(url.hostname, topLevelUrl.hostname))
+		if (url != null && domainsMatch(url, topLevelUrl)) {
+			console.log("skipping, is first party")
```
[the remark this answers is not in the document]
brindy: planning to remove console.log calls anyway

Core/tlds.js
```diff
@@ -0,0 +1,7921 @@
+//
```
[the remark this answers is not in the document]
brindy: I wrote it *based* on their example.  The JSON is theirs though.  Thoughts?

Core/tlds.js
brindy: We should also be downloading a new version of it from somewhere IMHO. Suggest we defer this to later, it's not going to update often.

Core/contentblocker.js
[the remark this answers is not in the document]
subsymbolic: Cool, removing it works for me too!

Core/tlds.js
subsymbolic: Cool, I see your new Asana task capturing this.

subsymbolic:
Loog great!

PR #160, opened 2017-10-02
universal app link fix

brindy:
Reviewer: Mia 
Asana: https://app.asana.com/0/414235014887631/430999069520960
CC:

**Description**:

This fixes a problem which occurs when a page redirects to a URL that has an associated universal link, launching the associated app and causing the page to finish with an error (load interrupted) which causes the error message UI to be shown.

(This does not solve the problem in which clicking on a search result from DDG results in a blank page as this is being fixed server side)

**Steps to test this PR**:

1. With the Twitter app installed, navigate to private.brindy.org.uk/twitter2.html
2. Click the link
3. The Twitter app will launch
4. Go back to the app and the previous page should be visible without any error message UI (note that the page may have scrolled up due to a separate issue, https://app.asana.com/0/414235014887631/416322909657029 )

###### Reviewer Checklist:

###### PR DRI Checklist:


Core/UniversalAppLinkRecovery.swift
```diff
@@ -0,0 +1,47 @@
+//
+//  UniversalAppLinkRecovery.swift
```
subsymbolic: This sounds like it specifically responds to universal app link recovery - but it really reloads any errors that happen while backgrounding. Can we rename? Maybe BackgoundErrorRecovery?

Core/AppUrls.swift
```diff
@@ -23,14 +23,16 @@ import Foundation
 public struct AppUrls {
 
     private struct Url {
+
+        // You can change the base to use a subdomain for testing, but the unit tests will fail
```
subsymbolic: Can we update this so that the unit test don't fail? It would be good if a dev working off a test server could still run the tests. If you'e worried about them inadvertently checking in the wrong base url, we could have a single unit test which fails on that.

Core/UniversalAppLinkRecovery.swift
brindy: Actually, that's a good point.  We don't want to do anything if the user backgrounds the app after an error under normal circumstances.  Need to think about this.

Core/WebViewController.swift
```diff
@@ -21,16 +21,19 @@ import UIKit
...
+
     public weak var webEventsDelegate: WebEventsDelegate?
     
     @IBOutlet weak var progressBar: UIProgressView!
     @IBOutlet weak var errorMessage: UILabel!
     
     open private(set) var webView: WKWebView!
 
+    private let universalAppLinkRecovery = UniversalAppLinkRecovery()
+    private var goBackWhenAppDidBecomeActive = false
```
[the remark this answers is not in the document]
brindy: Nope, good sot.

brindy:
Thanks for the feedback.  

I've simplified the reload scenario.  There's a remote possibility the user could background the app while a request is in progress which then fails, but I think that's going to be a very rare edge case.

I've also updated the tests to support testing with a subdomain, as suggested.

PR #161, opened 2017-10-03
use entire disconnect list to look up associated sites

brindy:
Reviewer: Mia
Asana: https://app.asana.com/0/414235014887631/438261419283368
CC:

**Description**:

Passes two structures to the Javascript - trackers that are banned and trackers that are allowed.  If a url is in the banned structure, it will be blocked, unless it is an associated first party url.  The logic for first party url checking checks both structures to give complete coverage of associated domains.

**Steps to test this PR**:
1. Visit a tracker heavy site and confirm that tracking is still working
2. Visit a site like criteo.com and ensure that first party URLs are still working

###### Reviewer Checklist:

###### PR DRI Checklist:

### DIFF
--- a/Core/contentblocker.js
+++ b/Core/contentblocker.js
@@ -132,7 +132,7 @@ var duckduckgoContentBlocking = function() {
 		}
 
 		var result = DisconnectMe.parentTracker(urlToCheck)
-		if (result && domainsMatch(result.parent, topLevelUrl)) {
+		if (result && domainsMatch(new URL(topLevelUrl.protocol + result.parent), topLevelUrl)) {
 			return true
 		}
 
--- a/Core/contentblocker.js
+++ b/Core/contentblocker.js
@@ -19,6 +19,9 @@
 
 var duckduckgoContentBlocking = function() {
 
+	var parentEntityUrl = null
+	var topLevelUrl = null
+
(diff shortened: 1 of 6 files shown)


Core/contentblocker.js
```diff
@@ -114,12 +119,22 @@ var duckduckgoContentBlocking = function() {
...
+		if (urlToCheck == null) {
+			return false
+		}
+
+		if (domainsMatch(urlToCheck, topLevelUrl)) {
+			return true
+		}
+
+		var result = DisconnectMe.parentTracker(urlToCheck)
+		if (result && domainsMatch(result.parent, topLevelUrl)) {
```
subsymbolic: This is comparison throws and error and never completes

subsymbolic: Do the extension do this in both directions? e.g
• instagram is a child of facebook, so allow an instagram tracker on facebook
• instagram is a child of facebook, so allow a facebook tracker on instragram

We currently cover the top case not the second, just want to make sure we're consistent.

brindy: Thanks.  I've updated the logic to make sure we cover both cases and improved error handling for security errors thrown in iframes.   We should probably run the report after this is merged ... :)

PR #165, opened 2017-10-05
Feature/use unchanged tracker lists

brindy:
Reviewer: Mia
Asana: https://app.asana.com/0/414235014887631/442414558181571
CC:

**Description**:

Tracks etags returned by the server and if it matches, doesn't pass the data to be persisted (and subsequently invalidating any caches, etc).

Alamofire caches via URLSession so returns 200 even if the response from the server is a 304 (confirmed with Charles).  

There's a setting to disable this when creating a URLRequest, but then you have to manage more of the caching yourself (expiries, etc). Comparing etags directly regardless of (positive response code) does the job.

**Steps to test this PR**:
1. Clean install - run the app
2. Visit a tracker heavy site and observe blocking for easylist and disconnect me
3. Kill the app
4. Launch the app - observe that etags have been matched when the downloader runs
5. Visit a tracker heavy site, observe blocking 

###### Reviewer Checklist:

###### PR DRI Checklist:

### DIFF
--- a/Core/APIRequest.swift
+++ b/Core/APIRequest.swift
@@ -57,7 +57,7 @@ public class APIRequest {
 
                 Logger.log(text: "Request for \(self.url) completed with response code: \(String(describing: response.response?.statusCode)) and headers \(String(describing: response.response?.allHeaderFields))")
 
-                guard nil == etag, etag != response.response?.headerValue(for: HeaderNames.etag) else {
+                if etag != nil && etag == response.response?.headerValue(for: HeaderNames.etag) {
                     Logger.log(text: "Using cached version of \(self.url) with etag: \(String(describing: etag))")
                     _ = completion(nil, nil)
                     return
@@ -86,7 +86,7 @@ class UserDefaultsETagStorage: APIRequestETagStorage {
 
     func etag(for url: URL) -> String? {
         let etag = defaults?.string(forKey: url.absoluteString)
-        print("*** etag for ", url, etag)
+        Logger.log(items: "stored etag for ", url, etag as Any)
         return etag
     }
 
--- a/Core/APIRequest.swift
+++ b/Core/APIRequest.swift
@@ -21,16 +21,17 @@
 import Foundation
 import Alamofire
 
-public enum APIRequestCompleteionResult {
 
-    case dataPersisted
-    case errorHandled
-
-}
-
-public typealias APIRequestCompletion = (Data?, Error?) -> APIRequestCompleteionResult
+public typealias APIRequestCompletion = (APIRequest.Response?, Error?) -> Void
 
 public class APIRequest {
+
+    public struct Response {
+
+        var data: Data?
+        var etag: String?
+
+    }
 
     struct HeaderNames {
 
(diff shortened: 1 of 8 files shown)


Core/BlockerListsLoader.swift
```diff
@@ -54,32 +54,38 @@ public class BlockerListsLoader {
...
                 self.easylistStore.persistEasylistPrivacy(data: data)
-            }
+            })
 
-            Logger.log(items: "EasylistPrivacyRequest", "\(String(describing: error))")
-            semaphore.signal()
         }
 
-        return 3
+        return 1
```
subsymbolic: Do we still need this return value?

brindy: forgot to push the correct number, it's there now.  still needed as we have to wait for all three to finish when updating from the content blocker error popover

Core/APIRequest.swift
```diff
@@ -21,26 +21,87 @@
...
             .responseData(queue: DispatchQueue.global(qos: .utility)) { response in
-                Logger.log(text: "Request for \(self.url) completed with result \(response.result)")
-                completion(response.data, response.error)
-        }
+
+                Logger.log(text: "Request for \(self.url) completed with response code: \(String(describing: response.response?.statusCode)) and headers \(String(describing: response.response?.allHeaderFields))")
+
+                if etag != nil && etag == response.response?.headerValue(for: HeaderNames.etag) {
+                    Logger.log(text: "Using cached version of \(self.url) with etag: \(String(describing: etag))")
+                    _ = completion(nil, nil)
```
subsymbolic: Does returning `nil` instead of data here essentially break the cache? I think returning an optional etag in the completion would work better than doing all this magic here and suppressing the data. Then the caller can store the etag if they care about it and choose how to proceed. It also avoids brittle url / etag lookups.

Core/BlockerListsLoader.swift
```diff
@@ -54,32 +54,38 @@ public class BlockerListsLoader {
...
 
-            Logger.log(items: "EasylistPrivacyRequest", "\(String(describing: error))")
-            semaphore.signal()
         }
 
         return 3
     }
 
+    private class func handleResponseAndSignalCompletion(data: Data?, error: Error?, semaphore: DispatchSemaphore, dataHandler: (Data) -> Void) -> APIRequestCompleteionResult {
+        var result: APIRequestCompleteionResult = .errorHandled
```
subsymbolic: Why is the valid case of a cached response being treated as an error?

Core/APIRequest.swift
brindy: There’s no magic here. As an API request it seems reasonable for this class to handle any caching. The caller doesn’t need to know about that and should just handle data if there is some. But the completion handler must be called so nils are appropriate as that’s the same as saying “no new data” and “no error”. Ie the lack of error does not imply the presence of data.

As long as the caller doesn’t persist nil (why would it?) it should be fine.

Are you seeing a problem running the code?

Core/BlockerListsLoader.swift
brindy: I didn’t want to store he etag if the handler hadn’t done something with the data (reinforcing the data completion handler’s role regarding data storage), and I didn’t want to use a bool as that was ambiguous. Let me think about this one.

Core/APIRequest.swift
brindy: I think the problem might be that this is an API” request” but can handle any URL. It might be better if this was more explicitly a content blocker list request which takes an enum case to build the URL and use that as the key for the etag cache.

I could even add that as an extra layer of abstraction between APIRequest and the BlockerListDownloader but my preference would just be to rename it and make the change as we don’t have a strong requirement for a generic downloading interface that I’m aware of and YAGNI.




I’ll uodate the code to be more specific.

brindy: Actually, I can see you're using it - I'll add the abstraction. :)

subsymbolic: Sounds good Chris!

Core/EasylistStore.swift
```diff
@@ -34,16 +34,25 @@ class EasylistStore {
 
     }
 
-    var hasData: Bool {
-        return easylist != "" && easylistPrivacy != ""
+    var hasData = false
```
subsymbolic: One small buglet here. This variable has slowly taken on two different meanings
1. That there is data from the server
2. That processed data from the js layer was received

This means if I kill the first easylist request (using charles) and then try and reload via the content blocking reload button, it doesn't seem to work. hasData continues to return false until the js response is received.

brindy: Gah, I was trying to avoid load those files in to memory is all.  I'll check against their presence on the file system instead.

PR #166, opened 2017-10-05
Add atb retention statistics

subsymbolic:
Reviewer: Chris
Asana: https://app.asana.com/0/414235014887631/416402065999825

**Description**:
Add additional calls required to track retention over time

**Steps to test this PR**:
Ensure that the app is making the calls described in the pinned comment in https://app.asana.com/0/414709148257752/407328226239763 (other then step 9)

###### Reviewer Checklist:

###### PR DRI Checklist:


Core/AppUrls.swift
```diff
@@ -114,11 +126,14 @@ public struct AppUrls {
     }
 
     public func applyStatsParams(for url: URL) -> URL {
-        let searchUrl = url.addParam(name: Param.source, value: ParamValue.source)
-            .addParam(name: Param.appVersion, value: appVersion)
+        var searchUrl = url.addParam(name: Param.source, value: ParamValue.source)
+        searchUrl = searchUrl.addParam(name: Param.appVersion, value: appVersion)
```
brindy: I prefer the chained approach myself, it's designed so that you can assign to a let in a single statement... not fussed though.

Core/StatisticsLoader.swift
```diff
@@ -22,26 +22,78 @@ import Foundation
 
 public class StatisticsLoader {
     
+    public typealias Completion =  (() -> ())
```
brindy: This feels redundant. Personally, I would avoid type aliases for functions unless they are more complex than this simple case or if they are going to be stored as a variable.

Either way, should return `Void` rather than `()`

Core/AppUrls.swift
subsymbolic: Me too! I just switched for consistency within the method as there's a spot where we can't chain. If you're not fussed, I'll keep it. Last chance to speak now if you want it changed...

Core/StatisticsLoader.swift
subsymbolic: I've switched to Void. I'm with you on the Completion being a potential overkill. I pulled it out as the code was getting a bit cumbersome with the escaping prefix and the default return parameter but would normally leave it.

Core/AppUrls.swift
subsymbolic: Me too! I just switched for consistency within the method as there's a spot where we can't chain. If you're not fussed, I'll keep it. Last chance to speak now if you want it changed...

PR #183, opened 2017-11-03
Merge live and model links

subsymbolic:
Reviewer: Caine and Chris
Asana: https://app.asana.com/0/414235014887631/469637695297538

**Description**:
Stop TabViewController from updating the model with blank fields in cases where the model has better data.

It is possible for the model to have a title from a previously persisted session and for the TabViewController's title to be blank while it is loading - so even though the TabViewController is most up-to-date we still want to fallback on the model data at times.

**Steps to test this PR**:
1. Open the app, launch a few tabs
2. Ensure they are fully loaded (check that all details are populated in the tab switcher)
3. Kill and open the app again
4. Open the tab switcher and ensure that all title data etc is populated.
5. Kill some tabs and ensure that the tab switcher data is maintained as items are reloaded from the cache.

###### Reviewer Checklist:

###### PR DRI Checklist:

### DIFF
--- a/Core/Link.swift
+++ b/Core/Link.swift
@@ -59,4 +59,26 @@ public class Link: NSObject, NSCoding {
         guard let other = other as? Link else { return false }
         return title == other.title && url == other.url && favicon == other.favicon
     }
+    
+    /**
+     Where links share the same url, uses the seconday link to plug any missing data in the
+     primary link.
+     */
+    public static func mergeData(primary: Link?, secondary: Link?) -> Link? {
+        guard let primary = primary else {
+            return secondary
+        }
+        
+        guard let secondary = secondary else {
+            return primary
+        }
+        
+        if primary.url != secondary.url {
+            return primary
+        }
+    
+        let title = (primary.title == nil || primary.title!.isEmpty) ? secondary.title : primary.title
+        let favicon = primary.favicon ?? secondary.favicon
+        return Link(title: title, url: primary.url, favicon: favicon)
+    }
 }
--- a/Core/Link.swift
+++ b/Core/Link.swift
@@ -61,24 +61,16 @@ public class Link: NSObject, NSCoding {
     }
     
     /**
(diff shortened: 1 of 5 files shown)


DuckDuckGo/TabViewController.swift
```diff
@@ -48,6 +48,15 @@ class TabViewController: WebViewController {
         webEventsDelegate = self
     }
     
+    public var link: Link? {
+        guard let url = url else {
+            return Link.mergeData(primary: nil, secondary: tabModel.link)
```
brindy: Given this function just returns the secondary if the primary is nil anyway, could just return the `tabModel.link` in this case and simplify `mergeData`?

DuckDuckGo/TabViewController.swift
```diff
@@ -48,6 +48,15 @@ class TabViewController: WebViewController {
...
+        }
+        
+        let activeLink = Link(title: name, url: url, favicon: favicon)
+        return Link.mergeData(primary: activeLink, secondary: tabModel.link)
+    }
```
brindy: Given primary will never be nil in this case, `mergeData` could just be an instance method which returns a new Link when merged, e.g.

`return activeLink.merge(with: tabModel.link)`

Core/Link.swift
```diff
@@ -59,4 +59,26 @@ public class Link: NSObject, NSCoding {
...
+     */
+    public static func mergeData(primary: Link?, secondary: Link?) -> Link? {
+        guard let primary = primary else {
+            return secondary
+        }
+        
+        guard let secondary = secondary else {
+            return primary
+        }
+
```
brindy: This should probably be an instance method (see comments below) in which case these guards would also be unnecessary.

subsymbolic:
@brindy, I've switched the static over to an instance method. Ready for rereview!

brindy:
LGTM, thanks @subsymbolic

PR #193, opened 2017-11-20
fix site grade when redirecting

brindy:
Reviewer: Mia or Caine
Asana: https://app.asana.com/0/413877547933097/395487348379600/f
CC:

**Description**:

Site grade should reflect currently URL, not the one at the beginning of request which might redirect.

**Steps to test this PR**:
1. Visit ddg.gg
1. When redirected to DuckDuckGo.com check site grade is A

###### Reviewer Checklist:

###### PR DRI Checklist:

### DIFF
--- a/Core/SiteRating.swift
+++ b/Core/SiteRating.swift
@@ -24,8 +24,8 @@ public class SiteRating {
     
     public var url: URL
     public var hasOnlySecureContent: Bool
-    public var domain: String {
-        return url.host ?? ""
+    public var domain: String? {
+        return url.host
     }
     public var finishedLoading = false
     private var trackersDetected = [Tracker: Int]()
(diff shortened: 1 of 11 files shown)


Core/SiteRating.swift
```diff
@@ -24,7 +24,9 @@ public class SiteRating {
     
     public var url: URL
     public var hasOnlySecureContent: Bool
-    public let domain: String
+    public var domain: String {
```
subsymbolic: The domain used to be guaranteed (e.g it was generated off a non-changing url and couldn't be optional as init would fail). Now that the url mutates, I would suggest deleting the domain altogether or as a minimum making it optional. Then the items which use it are forced to consider how to respond to no data rather than accidentally assuming that the non-optionlal means there is data.

PR #194, opened 2017-11-21
fix site grade being carried over

brindy:
Reviewer: Mia
Asana: https://app.asana.com/0/413877547933097/483017302281705/f
CC: Caine

**Description**:

Sites appeared to be carrying over the grade.

Additionally, entering search query or a url would be temporarily overwritten by the current site's url after submitting until the load had made progress.

**Steps to test this PR**:

Test 1
1. Bookmark DuckDuckGo.com
1. Visit cnn.com - observe D rating
1. Without giving CNN too much time to load, select DDG from the bookmarks - observe A rating
1. Repeat on a few sites
1. Observe text in omni reflects what is submitted until the page load begins and makes progress.

###### Reviewer Checklist:

###### PR DRI Checklist:

### DIFF
--- a/DuckDuckGo/TabViewController.swift
+++ b/DuckDuckGo/TabViewController.swift
@@ -332,6 +332,7 @@ extension TabViewController: WKScriptMessageHandler {
     }
     
     struct TrackerDetectedKey {
+        static let pageUrl = "pageUrl"
         static let blocked = "blocked"
         static let networkName = "networkName"
         static let url = "url"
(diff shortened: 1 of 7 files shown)


DuckDuckGo/TabViewController.swift
```diff
@@ -374,6 +385,8 @@ extension TabViewController: WebEventsDelegate {
     func webpageDidStartLoading() {
         Logger.log(items: "webpageLoading started:", Date().timeIntervalSince1970)
         resetSiteRating()
+        reloadScripts(with: siteRating!.protectionId)
```
subsymbolic: Is there any chance the SiteRating could be nil here? Consider a guar check.

brindy: If the page has started loading then there must be a url and in that case there’s a site rating, but yep, can wrap in a guard in case that changes later.

subsymbolic: My concern with non-local force unwraps is that you lose the compiler check forever more and bugs can creep in as the code gets refactored. You'd be hard pressed to find a swift style guide that doesn't explicitly call it out. I hear what you're saying about the site rating definitely existing in this instance though. If you want to avoid a guard, you could refactor to grab the instance as it is created and pass it to this method as a non-optional param.

What you've got is technically correct, so I'm approving the PR but do think on it!

brindy: Yep, I agree in this case. It’s wrong to assume the previous methods will behave the same way forever (especially when we don’t have unit tests). I’ll make the change before I merge.

PR #201, opened 2017-11-27
explicitly launch itunes protocol links

brindy:
Reviewer: Mia
Asana: 
CC:

**Description**:

Launch iTunes links explicitly.

**Steps to test this PR**:

On a device (will just get error page on simulator)

1. Search for 
> cnet best iOS apps
1. Tap on the CNET Best iPhone Apps of 2017 link
1. Tap on an iOS app link
1. The App Store app will launch

### DIFF
--- a/Core/SupportedExternalURLScheme.swift
+++ b/Core/SupportedExternalURLScheme.swift
@@ -26,9 +26,11 @@ public enum SupportedExternalURLScheme: String {
     case mailto
     case maps
     case sms
+    case itunesapps = "itms-appss"
     
     public static func isSupported(url: URL) -> Bool {
         guard let scheme = url.scheme else { return false }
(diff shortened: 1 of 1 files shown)


Core/WebViewController.swift
```diff
@@ -209,6 +209,12 @@ open class WebViewController: UIViewController, WKNavigationDelegate, WKUIDelega
             return
         }
 
+        if url.absoluteString.starts(with: "itms-appss://") {
```
subsymbolic: We have implemented this functionality in `SupportedExternalURLScheme`

brindy:
There's still a flicker of the error page, but we've solved the immediate problem so I think we should go ahead and see if we get any feedback about it.

subsymbolic:
Yep agree, the user can now get to itunes connect, so it's still worth getting this in

PR #204, opened 2017-11-28
fix major network detection

brindy:
Reviewer: Mia, Caine or Craig 
Asana: https://app.asana.com/0/414235014887631/488551667050388
CC:

**Description**:

Since we switched to using the network name we are excluding Amazon and Yahoo as major trackers.

**Steps to test this PR**:

Test 1

1. Visit CNN.com
1. Open the major tracker networks detail
1. Observe Amazon in the list

Test 2

1. Visit developer.yahoo.com
1. Turn off privacy protection
1. Open major networks detail
1. Observer Yahoo in the list

### DIFF
--- a/Core/MajorTrackerNetwork.swift
+++ b/Core/MajorTrackerNetwork.swift
@@ -23,12 +23,14 @@ import Foundation
 public protocol MajorTrackerNetworkStore {
 
     func network(forName name: String) -> MajorTrackerNetwork?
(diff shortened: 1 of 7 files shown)


subsymbolic:
@brindy, in your second test you suggest that yahoo should show up as blocked on a yahoo subdomain. Is that correct? I would expect it to be exempt.

brindy:
@subsymbolic if you turn off privacy protection it is shown as detected (and obviously not blocked).

PR #205, opened 2017-11-28
add itunes to list of supported external schemes

brindy:
Reviewer: Mia
Asana: https://app.asana.com/0/413877547933097/312386120368401/f
CC:

**Description**:

Add another iTunes specific url scheme

**Steps to test this PR**:
1. Access https://itunes.apple.com/gb/author/marc-cameron/id470262913?mt=11&ign-itsct=1227297889-470262913&ign-itscg=0178&ign-mpt=uo%3D4
1. Tap on "View in iTunes" for one of the links
1. iTunes is launched

### DIFF
--- a/Core/SupportedExternalURLScheme.swift
+++ b/Core/SupportedExternalURLScheme.swift
@@ -26,6 +26,7 @@ public enum SupportedExternalURLScheme: String {
     case mailto
     case maps
     case sms
+    case itunes = "itmss"
     case itunesapps = "itms-appss"
     
     public static func isSupported(url: URL) -> Bool {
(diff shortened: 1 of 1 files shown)


Core/SupportedExternalURLScheme.swift
```diff
@@ -26,6 +26,7 @@ public enum SupportedExternalURLScheme: String {
     case mailto
     case maps
     case sms
+    case itunes = "itmss"
```
subsymbolic: Did this used to be "itms"? Do we still need to support those too?

brindy:
`itmss` (with the extra s) is the url scheme causing problems. Do you know of any iTunes urls that use `itms`?

subsymbolic:
> Do you know of any iTunes urls that use itms?

Exactly my question :) I'm interested to know if "itms" links are still valid. If they are, then we should fix them too

PR #219, opened 2017-12-05
show is major tracker network status on the score card

brindy:
Reviewer: Caine
Asana: https://app.asana.com/0/414235014887631/487711692211751
CC:

**Description**:

Adds an entry to the score card to show if a site is a major tracker network or not (e.g. to clarify why sites like Google get a D grade)

**Steps to test this PR**:
1. Visit Google, view score card, note Site is Major Tracker Network text and red cross
1. Visit DuckDuckGo, view score card, note Site is not Major Tracker Network text and green check

### DIFF
--- a/Core/SiteRatingScoreExtension.swift
+++ b/Core/SiteRatingScoreExtension.swift
@@ -80,6 +80,12 @@ public extension SiteRating {
         if let network = majorTrackerNetworkStore.network(forDomain: domain) { return network.score }
         return 0
     }
+
+    var isMajorTrackerNetwork: Bool {
+        get {
+            return isMajorTrackerScore > 0
+        }
+    }
     
     private var ipTrackerScore: Int {
         return containsIpTracker ? 1 : 0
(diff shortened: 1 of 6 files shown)


Core/SiteRatingScoreExtension.swift
```diff
@@ -80,6 +80,12 @@ public extension SiteRating {
         if let network = majorTrackerNetworkStore.network(forDomain: domain) { return network.score }
         return 0
     }
+
+    var isMajorTrackerNetwork: Bool {
```
subsymbolic: I would move this around a bit. Add the logic for working out the `isMajorTrackerNetwork` condition here and then get `isMajorTrackerScore` to use this conditional to work out the score.

OPTIONAL: It would be nice to move the conditionals into the SiteRating itself then we have all the interesting info about a site there and this extension is just about the "score"

brindy: I don't think that works - the score returned by `isMajorTrackerScore` comes from `network.score` and is used by the site grade directly (divided by 10).

subsymbolic: Ah yes ignore, I forgot that it was used as a multiplier.

PR #229, opened 2017-12-12
show correct plurality of detected/blocked network(s)

brindy:
Reviewer:
Asana:
CC:

**Description**:

Show the correct plurality for the networks.  

The correct way to do this is to use a strings dictionary rather than branching logic and using the localised strings file, but that is overkill at this stage.  Will need to revisit when we localise for multiple languages.

**Steps to test this PR**:
1. Visit a site and open the privacy dashboard immediately
1. Observe the text update for blocked network(s)
1. Turn off protection 
1. Observe the text update for detected network(s)

Note, nytimes.com is good for observing the singular.  cnn.com is good for the plural.

### DIFF
--- a/DuckDuckGo/SiteRatingPrivacyProtectionExtension.swift
+++ b/DuckDuckGo/SiteRatingPrivacyProtectionExtension.swift
@@ -61,11 +61,19 @@ extension SiteRating {
     }
 
     func majorNetworksBlockedText() -> String {
-        return String(format: UserText.privacyProtectionMajorTrackersBlocked, uniqueMajorTrackerNetworksBlocked)
+        if 1 == uniqueMajorTrackerNetworksBlocked {
+            return UserText.privacyProtectionMajorTrackerBlocked
+        } else {
+            return String(format: UserText.privacyProtectionMajorTrackersBlocked, uniqueMajorTrackerNetworksBlocked)
+        }
     }
 
     func majorNetworksDetectedText() -> String {
-        return String(format: UserText.privacyProtectionMajorTrackersFound, uniqueMajorTrackerNetworksDetected)
+        if 1 == uniqueMajorTrackerNetworksDetected {
+            return UserText.privacyProtectionMajorTrackerFound
+        } else {
+            return String(format: UserText.privacyProtectionMajorTrackersFound, uniqueMajorTrackerNetworksDetected)
+        }
     }
 
     func networksText(contentBlocker: ContentBlockerConfigurationStore) -> String {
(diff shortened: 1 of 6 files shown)


DuckDuckGo/SiteRatingPrivacyProtectionExtension.swift
```diff
@@ -61,11 +61,19 @@ extension SiteRating {
     }
 
     func majorNetworksBlockedText() -> String {
-        return String(format: UserText.privacyProtectionMajorTrackersBlocked, uniqueMajorTrackerNetworksBlocked)
+        if 1 == uniqueMajorTrackerNetworksBlocked {
```
subsymbolic: Can we use plural dictionaries to avoid writing additional code here? They also translate better https://developer.apple.com/library/content/documentation/MacOSX/Conceptual/BPInternational/LocalizingYourApp/LocalizingYourApp.html

DuckDuckGo/SiteRatingPrivacyProtectionExtension.swift
```diff
@@ -61,11 +61,19 @@ extension SiteRating {
...
+            return UserText.privacyProtectionMajorTrackerBlocked
+        } else {
+            return String(format: UserText.privacyProtectionMajorTrackersBlocked, uniqueMajorTrackerNetworksBlocked)
+        }
     }
 
     func majorNetworksDetectedText() -> String {
-        return String(format: UserText.privacyProtectionMajorTrackersFound, uniqueMajorTrackerNetworksDetected)
+        if 1 == uniqueMajorTrackerNetworksDetected {
```
subsymbolic: Not a biggie but for readability reasons I'd suggest switching these:
`if uniqueMajorTrackerNetworksDetected equals 1`
reads better than
`if 1 equals uniqueMajorTrackerNetworksDetected`

In the first example the variable provides context. In the second, the reader must wait longer before getting to salient information they need to assess the statement.

brindy: Agree.  It's an old habit from my first job I've never been able to shake "always test against the constant".

DuckDuckGo/SiteRatingPrivacyProtectionExtension.swift
brindy: That's what I was referring to in the description.  I'll have a look, but it doesn't look like quick job. If it is, I'll do it now but otherwise will have to do the release tomorrow.

subsymbolic: I think it's worth doing. The reason we put strings in `Localizable.strings` is so we can translate easily later it would be a shame to add breaking changes now. Happy to pair on it too if that makes the process less painful.

brindy:
Done, please re-review.

DuckDuckGo/UserText.swift
```diff
@@ -70,6 +70,8 @@ public struct UserText {
     public static let privacyProtectionEncryptionGoodConnection = NSLocalizedString("privacy.protection.encryption.good.connection", comment: "Encrypted conection")
     public static let privacyProtectionEncryptionBadConnection = NSLocalizedString("privacy.protection.encryption.bad.connection", comment: "Unencrypted conection")
     public static let privacyProtectionEncryptionMixedConnection = NSLocalizedString("privacy.protection.encryption.mixed.connection", comment: "Mixed encryption conection")
+
+    // Plurals
```
subsymbolic: Delete old comment

PR #233, opened 2017-12-15
handle blocker list download error

brindy:
Reviewer: Mia
Asana: https://app.asana.com/0/414235014887631/470949914705548
CC:

**Description**:

Show an error in privacy protection when there's no blocker lists data and allow user to try and reload

**Steps to test this PR**:
1. Clean install of app then quickly break the network connection (breakpoint in start func of `BlockerListsLoader` class if using simulator, turn off network, then resume)
1. Tap site rating view
1. Error message about content blocker lists with retry button should appear
1. Tap retry - activity indicator will appear and button will disappear
1. Shortly the button will reappear and the activity indicator will disappear
1. Turn on network
1. Tap retry - activity indicator will appear, button will disappear
1. Privacy protection will close and web page will reload
1. Open privacy protection to see site grade

### DIFF
--- a/DuckDuckGo/PrivacyProtectionController.swift
+++ b/DuckDuckGo/PrivacyProtectionController.swift
@@ -77,8 +77,8 @@ class PrivacyProtectionController: UIViewController {
     private func showLoadBlockerListTryAgain(_ controller: PrivacyProtectionErrorController) {
         controller.onTryAgain {
             let blockerListLoader = BlockerListsLoader()
-            blockerListLoader.start { newData in
-                self.handleLoadBlockerListResult(controller, blockerListLoader.hasData)
+            blockerListLoader.start { [weak self] newData in
+                self?.handleLoadBlockerListResult(controller, blockerListLoader.hasData)
             }
         }
     }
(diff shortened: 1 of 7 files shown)


DuckDuckGo/PrivacyProtectionController.swift
```diff
@@ -63,6 +67,33 @@ class PrivacyProtectionController: UIViewController {
...
+        controller.errorText = UserText.privacyProtectionReloadBlockerLists
+        embeddedController.pushViewController(controller, animated: true)
+        showLoadBlockerListTryAgain(controller)
+    }
+
+    private func showLoadBlockerListTryAgain(_ controller: PrivacyProtectionErrorController) {
+        controller.onTryAgain {
+            let blockerListLoader = BlockerListsLoader()
+            blockerListLoader.start { newData in
+                self.handleLoadBlockerListResult(controller, blockerListLoader.hasData)
```
subsymbolic: The strong reference to self will likely cause a memory leak. I ran the profiler and the privacy protection dashboard is never cleared from memory. There are a few strong references to self that probably need a review to make sure they are safe.
screen shot 2017-12-15 at 12 37 23

brindy: Ah, amazing. Thanks @subsymbolic

brindy: I don't think this will leak as the closure is not retained, but the user might dismiss the view controller while the update is progress which could cause a crash.

PR #239, opened 2017-12-18
externalise MPL dependencies in a submodule

brindy:
Reviewer: Mia or Caine
Asana: https://app.asana.com/0/414235014887631/466646695795380
CC:

**Description**:

Move the MPL dependencies in to their own project and include them via a submodule.

Note that after doing a pull/check out/clone, you'll now have to run `git submodule update --remote --merge`

**Steps to test this PR**:

* On an iOS 9 device, visit evansycles.com and ensure tracking/blocking looks correct
* On an iOS 10 device, visit evansycles.com and ensure tracking/blocking looks correct
* On an iOS 11 device, visit evansycles.com and ensure tracking/blocking looks correct


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "submodules/ios-js-support"]
+	path = submodules/ios-js-support
+	url = https://github.com/duckduckgo/ios-js-support.git
```
subsymbolic: How do you feel about tagging the other repo and then pointing to a specific tag here?

subsymbolic: Ok scrap that since we can't do that with submodules! Let's just also tag the other repo so we have a version record there.

PR #242, opened 2018-01-04
implement correct first party lookup in app

brindy:
Reviewer: Caine or Mia
Asana: https://app.asana.com/0/414235014887631/414510321511365
CC:

**Description**:

Some trackers were being categories as Amazon just because their domain name ended with amazon.com, e.g. images-na.ssl-images-amazon.com yet were still blocked by our JavaScript.  This meant we appeared to be blocking Amazon on their site.

Even though they probably are Amazon trackers, to make the above domain appear as an Amazon tracker (and not be blocked on Amazon) the correct thing to do would be to upgrade our related URL lookup source (ie Disconnect), once confirmed it really is related.

This PR ensures consistency between JavaScript and iOS.

**Steps to test this PR**:
1. Visit Amazon.com
1. Note that items not in Disconnect appear in the list of trackers

### DIFF
--- a/Core/SiteRatingScoreExtension.swift
+++ b/Core/SiteRatingScoreExtension.swift
@@ -120,7 +120,7 @@ public extension SiteRating {
     }
 
     public func networkNameAndCategory(forDomain domain: String) -> ( networkName: String?, category: String? ) {
-        if let tracker = disconnectMeTrackers.first(where: { domain.hasSuffix($0.key) } )?.value {
+        if let tracker = disconnectMeTrackers.first(where: { domain == $0.key || domain.hasSuffix(".\($0.key)") } )?.value {
             return ( tracker.networkName, tracker.category?.rawValue )
         }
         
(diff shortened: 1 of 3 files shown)


Core/SiteRatingScoreExtension.swift
```diff
@@ -120,7 +120,7 @@ public extension SiteRating {
     }
 
     public func networkNameAndCategory(forDomain domain: String) -> ( networkName: String?, category: String? ) {
-        if let tracker = disconnectMeTrackers.first(where: { domain.hasSuffix($0.key) } )?.value {
+        if let tracker = disconnectMeTrackers.first(where: { domain == $0.key || domain.hasSuffix(".\($0.key)") } )?.value {
```
subsymbolic: Should this one be lowercased too? It be worth extracting this into a separate method as it happens in a few places.

brindy: Thanks, probably. Will update, test and merge.

PR #246, opened 2018-01-10
use equality and an index when fetching domains for https upgrade

brindy:
Reviewer: Caine or Mia
Asana: https://app.asana.com/0/488551667048375/514330959360383
CC:

**Description**:

Adds an index to the database and uses equality to look up domains (instead of like).  Note that the smaller list is still in use at this point until the other PR iOS merged.

Also fixes a crash bug (I think on iOS or slower devices, only observed on iPhone 5s running iOS 10).

**Steps to test this PR**:
1. Browse about - performance appear normal especially when accessing
1. Use the fire button - shouldn't crash.


nilnilnil (approved):
👍 nice

PR #250, opened 2018-01-16
improved privacy grade accuracy after navigation

brindy:
Reviewer: Caine or Mia
Asana: 

* https://app.asana.com/0/414235014887631/523426024528729

CC:

**Description**:

Fixes a couple of bugs with the site grade showing incorrectly because of back/forward button and swipe navigation.

Also enhances privacy grade caching so that the privacy dashboard is more accurate when performing back and forward navigation without a page reload

**Steps to test this PR**:

***Test 1***:

1. Clean launch of the app
1. Go directly to huffingtonpost.com
1. Perform a search: DDG Grade should be A

Repeat with privacy protection off

***Test 2***:

1. Clear data
1. Search for cnn
1. Navigate to cnn.com 
1. Swipe back to DDG
1. DDG should remain an A

(repeat a few times to be sure)

***Test 3***

1. Go to cnn.com directly
1. Go to ddg.gg directly 
1. Redirects to DuckDuckGo.com which has an A grade

### DIFF
--- a/DuckDuckGo/TabViewController.swift
+++ b/DuckDuckGo/TabViewController.swift
@@ -103,8 +103,13 @@ class TabViewController: WebViewController {
     }
 
     override func goBack() {
-        resetSiteRating()
+        siteRating = nil
         super.goBack()
+    }
+    
+    override func goForward() {
+        siteRating = nil
+        super.goForward()
     }
 
     private func addContentBlockerConfigurationObserver() {
@@ -147,10 +152,6 @@ class TabViewController: WebViewController {
     
     fileprivate func updateSiteRating() {
         if isError {
-            siteRating = nil
-        } else if let url = url {
-            siteRating?.url = url
-        } else {
             siteRating = nil
         }
         onSiteRatingChanged()
--- a/DuckDuckGo/TabViewController.swift
+++ b/DuckDuckGo/TabViewController.swift
@@ -125,6 +125,14 @@ class TabViewController: WebViewController {
         }
     }
     
+    func webView(_ webView: WKWebView, didReceiveServerRedirectForProvisionalNavigation navigation: WKNavigation!) {
+        guard let url = webView.url else { return }
+        
+        siteRating = SiteRating(url: url)
+        reloadScripts(with: siteRating!.protectionId)
+        updateSiteRating()
+    }
+    
     private func resetNavigationBar() {
         chromeDelegate?.setBarsHidden(false, animated: false)
     }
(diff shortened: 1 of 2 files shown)


DuckDuckGo/PrivacyProtectionHeaderController.swift
```diff
@@ -52,9 +52,9 @@ class PrivacyProtectionHeaderController: UIViewController {
     private func update() {
         guard isViewLoaded else { return }
 
-        let grades = siteRating.siteGrade()
+        // let grades = siteRating.siteGrade()
```
subsymbolic: Delete unused code

DuckDuckGo/TabViewController.swift
```diff
@@ -486,6 +487,12 @@ extension TabViewController: WebEventsDelegate {
         updateSiteRating()
     }
 
+    func webView(_ webView: WKWebView, didReceiveServerRedirectForProvisionalNavigation navigation: WKNavigation!) {
```
subsymbolic: Good change!

DuckDuckGo/SiteRatingView.swift
```diff
@@ -71,8 +71,8 @@ public class SiteRatingView: UIView {
         guard BlockerListsLoader().hasData else { return }
         guard let siteRating = siteRating else { return }
         
-        let grades = siteRating.siteGrade()
-        let grade = contentBlockerConfiguration.protecting(domain: siteRating.domain) ? grades.after : grades.before
+//        let grades = siteRating.siteGrade()
```
subsymbolic: Delete

subsymbolic (commented):
@brindy I think it's better but I can still reproduce it
bug_redroduced

brindy:
@subsymbolic thanks again for your help - can you re-review when you get a chance, please? :)

subsymbolic:
@brindy, I'm afraid I can still reproduce this issue. There's a fair amount of unrelated code in this PR I think it's worth going back to "basics" and resetting this branch with just the related changes (the refactoring is good maybe it can go in after the release) so we can better see and tweak the fix.

ddb_with_tracker

brindy:
I've reset this branch to just the changes we were discussing last night without any of the extra site rating caching logic.  Am just re-testing it locally.

brindy:
I've revised the tests to exclude the site rating cache and this tests OK for me which fixes the initial bug that was raised. If you're still able to reproduce can you please say which test and/or the steps you took (so we can add them as tests).

Thanks!

subsymbolic (commented):
Looks good @brindy, I can no longer reproduce it.

In case it's useful for testing when you later reapply the refactor, I was reproducing it by:
• switching off content blocking
• loading a heavy site e.g cnn.com
• while that was still in progress, I would perform an omnibar search to get us back to ddg and then load another heavy site
• swiping between all the pages for a while

subsymbolic (approved):

PR #251, opened 2018-01-16
improve stability of tab switcher

brindy:
Reviewer: Caine or Mia
Asana: https://app.asana.com/0/414235014887631/523274217653540
CC:

**Description**:

The index associated with a tab in the switcher could easily get out of sync and result in weird UI or worse, a crash.

**Steps to test this PR**:
1. Add a bunch of tabs
1. Close a few
1. Switch between them
1. Close them all
1. Repeat a few times, try at speed (ie to break it)

### DIFF
--- a/DuckDuckGo/TabViewCell.swift
+++ b/DuckDuckGo/TabViewCell.swift
@@ -22,9 +22,18 @@ import UIKit
 import Core
 import Kingfisher
 
+protocol TabViewCellDelegate: class {
+    
+    func deleteTab(tab: Tab)
+    
+}
+
 class TabViewCell: UICollectionViewCell {
     
     static let reuseIdentifier = "TabCell"
+    
+    weak var delegate: TabViewCellDelegate?
+    weak var tab: Tab?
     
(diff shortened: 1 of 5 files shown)


DuckDuckGo/TabViewCell.swift
```diff
@@ -22,21 +22,45 @@ import UIKit
...
+        self.tab = tab
+        isHidden = false
         title.text = tab.link?.title ?? ""
         link.text = tab.link?.url.absoluteString ?? ""
         configureFavicon(tab.link?.favicon)
     }
     
+    @IBAction func deleteTab() {
+        
+        guard let tab = tab else { return }
```
nilnilnil, line 53:
```
guard let tab = tab else { return }
```

instead of 

```
guard let tab = self.tab else { return }
```

nilnilnil (approved):

DuckDuckGo/TabViewCell.swift
subsymbolic: Strangely, it's correct. It's unwrapping the optional, if it fails it returns.

Issue #257, opened 2018-01-24
Drop SwiftyJSON Dependency


Marcocanc:
SwiftyJSON is only used in 3 places ([1](https://github.com/duckduckgo/iOS/blob/develop/Core/AtbParser.swift), [2](https://github.com/duckduckgo/iOS/blob/develop/Core/TermsOfServiceListParser.swift), [3](https://github.com/duckduckgo/iOS/blob/develop/Core/DisconnectMeTrackersParser.swift)).
Swift 4 offers built in [JSON Decoding](https://developer.apple.com/documentation/swift/encoding_decoding_and_serialization).
Would be nice to have one less dependency.

brindy:
Good point! Thanks for the suggestion.

brindy:
This was done in PR #271 - thanks for your input and contribution. :)

PR #258, opened 2018-01-24
Replace SwiftyJSON with Codable (2/3)

Marcocanc:
Partially adopted Codable. (#257)
Left out `DisconnectMeTrackersParser` because it seemed a little complex.

With Codable there's no need for the `*Parser` structs. If you still want to throw your own errors I recommend writing an extension on `Decodable` like this:

```swift
extension Decodable {
    init(fromJsonData data: Data) throws {
        do {
            let decoder = JSONDecoder()
            self = try decoder.decode(Self.self, from: data)
        } catch DecodingError.dataCorrupted {
            throw JsonError.invalidJson
        } catch {
            throw JsonError.typeMismatch
        }
    }
}
```

### DIFF
--- a/Core/Atb.swift
+++ b/Core/Atb.swift
@@ -20,7 +20,7 @@
 
 import Foundation
 
-public struct Atb {
+public struct Atb: Decodable {
     
     public static let variant = "mi"
 
--- a/Core/AtbParser.swift
+++ b/Core/AtbParser.swift
@@ -19,20 +19,17 @@
 
 
 import Foundation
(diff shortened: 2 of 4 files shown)


brindy:
Thanks for this and thanks for your patience. We'll probably pick this up next week. If you want to try and remove the dependency from `DisconnectMeTrackersParser` as well that would be good, but if not we can do that. :)

brindy (approved):
This is looking good, thanks for your contribution!

PR #265, opened 2018-01-30
Hotfix/7.0.5.896

brindy:
Reviewer: Mia
Asana: 

https://app.asana.com/0/414235014887631/540217967080411
https://app.asana.com/0/414235014887631/487711692211901

**Description**:

Defensive code to prevent a crash.
Fix colours for the settings, bookmarks and whitelist screens on iOS 9 only

**Steps to test this PR**:
1. On an iOS 9 device confirm the above screens look good
1. On an iOS 10 and 11 device confirm the above screens have not been affected
1. Browse the web and use the privacy dashboard - should be no errors introduced

### DIFF
--- a/DuckDuckGo/BookmarksDataSource.swift
+++ b/DuckDuckGo/BookmarksDataSource.swift
@@ -46,13 +46,16 @@ class BookmarksDataSource: NSObject, UITableViewDataSource {
     }
     
     private func createEmptyCell(_ tableView: UITableView) -> UITableViewCell {
-        return tableView.dequeueReusableCell(withIdentifier: NoBookmarksCell.reuseIdentifier)!
+        let cell = tableView.dequeueReusableCell(withIdentifier: NoBookmarksCell.reuseIdentifier)!
+        updateColor(cell)
+        return cell
     }
     
     private func createBookmarkCell(_ tableView: UITableView, forIndex index: Int) -> UITableViewCell {
         let bookmark = bookmarksManager.bookmark(atIndex: index)
         let cell = tableView.dequeueReusableCell(withIdentifier: BookmarkCell.reuseIdentifier) as! BookmarkCell
         cell.update(withBookmark: bookmark)
+        updateColor(cell)
         return cell
     }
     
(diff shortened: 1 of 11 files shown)


DuckDuckGo/BookmarksDataSource.swift
```diff
@@ -74,4 +77,9 @@ class BookmarksDataSource: NSObject, UITableViewDataSource {
     func tableView(_ tableView: UITableView, moveRowAt sourceIndexPath: IndexPath, to destinationIndexPath: IndexPath) {
         bookmarksManager.move(itemAtIndex: sourceIndexPath.row, to: destinationIndexPath.row)
     }
+    
+    private func updateColor(_ cell: UITableViewCell) {
+        if #available(iOS 10, *) { return }
```
subsymbolic, line 82: We used to have the color in IB but it's easy to see it there and delete it thinking it's not needed. I like this better, it makes the reason apparent.

fastlane/Fastfile
```diff
@@ -14,7 +14,7 @@ lane :release do
   #       app_identifier: "com.duckduckgo.mobile.ios",
   #       version: "7"
   #     ) + 1
-  buildNumber = 895
+  buildNumber = 896
```
subsymbolic, line 17: Are you bumping the minor version too?

Optional: Since this is no longer automated why not delete it from here and update it in the General panel in XCode (or Info.plist) directly? Saves the confusion of having a duplicate.

subsymbolic (approved):
Feel free to merge when you're ready

fastlane/Fastfile
brindy: Thanks @subsymbolic - Good spot - I thought I had done that and pushed it.

Moving forward I will probably reset the build number to 1 and then not bother with it unless we are iterating over a release. I think we discussed something like this before.  However, it's easier to bump the build number via Fastlane as Fastlane will update any extensions, etc as well.  Though since we don't really have any extensions, I guess that's kind of moot.

PR #266, opened 2018-01-31
Feature/surrogates

brindy:
Reviewer: Mia or Caine
Asana: https://app.asana.com/0/414235014887631/520419619790473
CC:

**Description**:

Uses surrogates when tracking scripts are blocked which break page functionality.

Note that raspberrypi.org is whitelisted so can't be used for testing, but the ones below are valid tests even with the whitelist.

**Steps to test this PR**:
1. Visit https://www.xxl.fi/urheilu-ja-pallopelit/pelit/lelut-ja-pelit/c/161404 - select an item and view the details
1. Visit https://www.trovaprezzi.it/ - site works
1. Visit https://www.eff.org/files/pbtest/ga_js_surrogate_test.html - the message 'it works!' should appear 
1. Visit other sites and confirm they still work


Core/SurrogateStore.swift
```diff
@@ -0,0 +1,77 @@
...
+                functionName = line.components(separatedBy: " ")[0]
+                jsFunction = ""
+                continue
+            }
+            
+            guard jsFunction != nil else { continue }
+            
+            jsFunction = "\(jsFunction!)\(line)\n"
+            
+            if line == "})();" {
```
subsymbolic: This parser scares me, slightly different spacing here would break stuff. Have there been any conversations about the format?

Core/WKWebViewConfigurationExtension.swift
```diff
@@ -48,6 +48,14 @@ extension WKWebViewConfiguration {
 
 fileprivate class Loader {
     
+    // Need to refactor so that that easylist and disconnect use the cache here instead of in their loader/parsers
```
subsymbolic: Can we put a TODO in front so the IDE flags it and a link to the Asana ticket scheduling it so it's not forgotten?

subsymbolic (commented):
Works well! A couple of observations

Core/WKWebViewConfigurationExtension.swift
brindy: Yep, was unsure if you were OK with TODOs being there.

DuckDuckGoTests/MockFiles/surrogates.txt
```diff
@@ -0,0 +1,33 @@
...
+
+example.com/script2.js application/javascript
+(function() {
+    console.log("Sample function 2")
+    //
+    // This one has more lines
+    //
+    var ga = function() { }
+    //
+}) ();
```
subsymbolic: Tests fail if there is a newline at the end of the file.

subsymbolic (commented):
Looks good! I think we should also accept a newline at the end of the file

Core/WKWebViewConfigurationExtension.swift
subsymbolic: Preferably not - but this is comment is a todo so we may as well label it as such

subsymbolic (approved):
Looks good!

PR #267, opened 2018-02-02
adblock usage fixes

brindy:
Reviewer: Mia or Caine
Asana: https://app.asana.com/0/414235014887631/476166083909434
CC:

**Description**:

Fixes two problems reconstructing the cached adblock structures:

* Using the wrong data structure to reconstruct the exception bloom filter
* Re-create Set objects on the raw filters

**Steps to test this PR**:

1. Visit evanscycles.com - observe trackers blocked (quite a few) and note that https://cdn.tagcommander.com/1705/tc_abtest.js is not blocked
1. Refresh the page and check again (using cached version)
1. Browse a few sites and confirm trackers blocked and privacy grade look reasonable

Additional tests to check for regressions:

***Test 1***:

1. Clean launch of the app
1. Go directly to huffingtonpost.com
1. Perform a search: DDG Grade should be A

Repeat with privacy protection off

***Test 2***:

1. Clear data
1. Search for cnn
1. Navigate to cnn.com 
1. Swipe back to DDG
1. DDG should remain an A

(repeat a few times to be sure)

***Test 3***

1. Go to cnn.com directly
1. Go to ddg.gg directly 
1. Redirects to DuckDuckGo.com which has an A grade


brindy:
@subsymbolic @nilnilnil - this is ready for review when you get a chance.  Definitely refer to the Asana task for background information on some of the changes but obviously any questions, just fire away.

Core/SiteRating.swift
```diff
@@ -21,12 +21,21 @@
 import Foundation
 
 public class SiteRating {
-    public let protectionId: String
-    public var url: URL
-    public var hasOnlySecureContent: Bool
+    
+    public var url: URL {
+        didSet {
```
subsymbolic: What is your motivation for reusing the SiteRating instance here? If the url changes would it be better to create a new instance? I always worry about "reset" logic. It would be easy to add a new property to the SiteRating and forget to reset it.

js/test.html
```diff
@@ -0,0 +1,28 @@
+<!--
+//
+//  test.js
```
subsymbolic: Where is this file used?

IntegrationTests/ContentBlockerTests.swift
```diff
@@ -26,6 +26,8 @@ class ContentBlockerTests: XCTestCase {
         static let iFrames = "http://localhost:8000/iframetrackers.html"
         static let resources = "http://localhost:8000/resourcetrackers.html"
         static let requests = "http://localhost:8000/requesttrackers.html"
+        static let imageSrc = "http://localhost:8000/image_src.html"
+        static let xhr = "http://localhost:8000/xhr.html"
```
subsymbolic, line 30: Nice!

subsymbolic: Great to see the integration tests up and running again. The skipOnboarding() method is broken though, can that get fixed as part of this PR too?

Core/SiteRating.swift
brindy: Yeah I know. But a new instance means a new protection id and that means re injecting the scripts.  So the motivation is trying to speed things up.

IntegrationTests/ContentBlockerTests.swift
brindy: Yep, will have a look.

js/test.html
brindy: It’s there to help debugging and testing of the content blocking during develop. I can get rid of this folder, actually.

brindy: -deleted, commented in wrong reply-

Core/SiteRating.swift
brindy: Oh and the goForward/goback calls which set the site rating to nil we’re contributing to the crash that was being reported a lot before 7.0.5 - i never saw it crash personally but after changing the code to handle a nil site rating and nil error I saw the (?) in the address bar quite a lot.

subsymbolic: What's the exact use case here? In what circumstance do you want the same protectionId but to reset everything else?

subsymbolic: If you do need it you could inject it into a new SiteRating instance.

Core/SiteRating.swift
```diff
@@ -22,19 +22,13 @@ import Foundation
...
-            hasOnlySecureContent = url.isHttps()
-        }
-    }
-    
     public var domain: String? {
         return url.host
     }
-
+    
+    public let url: URL
```
subsymbolic: Nice!

subsymbolic (approved):
Great work.

PR #270, opened 2018-02-07
hot fix: missing trackers in embedded documents

brindy:
Reviewer: Mia or Caine
Asana: https://app.asana.com/0/414235014887631/551796273628075
CC:

**Description**:

Fixed a bug that was allowing trackers through in embedded documents.

**Steps to test this PR**:

1. Visit http://holywarsoo.net/viewtopic.php?id=16&p=6002
1. Check the privacy dashboard shows that Google trackers have been blocked
1. Perform exploratory testing as usual

### DIFF
--- a/Core/contentblocker.js
+++ b/Core/contentblocker.js
@@ -63,7 +63,8 @@ var duckduckgoContentBlocking = function() {
 	// private 
 	function getTopLevelURL() {
 		try {
-			// FROM: https://stackoverflow.com/a/7739035/73479		
+			// FROM: https://stackoverflow.com/a/7739035/73479
+			// FIX: Better capturing of top level URL so that trackers in embedded documents are not considered first party
 			return new URL(window.location != window.parent.location ? document.referrer : document.location.href)
 		} catch(error) {
 			return new URL(location.href)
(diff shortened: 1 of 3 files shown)


Core/contentblocker.js
```diff
@@ -62,9 +62,9 @@ var duckduckgoContentBlocking = function() {
 
 	// private 
 	function getTopLevelURL() {
-		// can throw a security exception if called from a frame with document loaded from different url to top (mainly when blocking is disabled)
 		try {
-			return new URL(top.location.href)
+			// FROM: https://stackoverflow.com/a/7739035/73479		
+			return new URL(window.location != window.parent.location ? document.referrer : document.location.href)
```
subsymbolic: This makes sense, although I'm concerned that there may be edge cases where document.referrer does not give you the top level url. I'm not a javascript expert, let's get someone who is to double-check. Maybe @nilnilnil?

nilnilnil: I'd like to see another comment about the why. I think it's OK, or something to try.

bbraithwaite: Just something to throw into the hat, there *could* be some trial and error here with regard to how the referrer policy is being set across services/sites: https://w3c.github.io/webappsec-referrer-policy/#referrer-policies i.e. there may be cases where document.referrer is undefined, but I guess that would be no different than what's in place without this fix?

nilnilnil (approved):

PR #273, opened 2018-02-09
text updates

brindy:
Reviewer: Caine or Mia
Asana:

* https://app.asana.com/0/414235014887631/521587462118443
* https://app.asana.com/0/414235014887631/552319851838328

CC:

**Description**:

Updates truncated text on smaller screens
Updates the mixed encryption state message

**Steps to test this PR**:

1. Check the privacy dashboard screens on devices of various sizes and check for consistency ensuring text is not truncated
1. Visit a secure site and check the encryption state message is correct
1. Visit a mixed encryption site (e.g. https://english.sogou.com/ ) and check the mixed encryption state text is correct
1. Visit a site with no encryption and check the encryption state message is correct

### DIFF
--- a/DuckDuckGo/PrivacyProtectionEncryptionDetailController.swift
+++ b/DuckDuckGo/PrivacyProtectionEncryptionDetailController.swift
@@ -82,10 +82,12 @@ class PrivacyProtectionEncryptionDetailController: UIViewController {
         unencryptedLabel.isHidden = true
         mixedContentLabel.isHidden = true
 
+        messageLabel.text = UserText.ppEncryptionStandardMessage
         if !siteRating.https {
             unencryptedLabel.isHidden = false
         } else if !siteRating.hasOnlySecureContent {
             mixedContentLabel.isHidden = false
+            messageLabel.text = UserText.ppEncryptionMixedMessage
         } else {
             encryptedLabel.isHidden = false
         }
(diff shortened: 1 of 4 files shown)


DuckDuckGo/PrivacyProtectionEncryptionDetailController.swift
```diff
@@ -82,10 +82,12 @@ class PrivacyProtectionEncryptionDetailController: UIViewController {
         unencryptedLabel.isHidden = true
         mixedContentLabel.isHidden = true
```
subsymbolic, line 83: Would the various overlapping encryption labels in the storyboard that are shown / hidden in code be better as a single label that gets set to the correct text according to state?
• encryptedLabel
• unencryptedLabel
• mixedContentLabel

subsymbolic (approved):
One suggestion otherwise looks good. Feel free to merge when you're ready.

DuckDuckGo/PrivacyProtectionEncryptionDetailController.swift
brindy: Probably, it's a nightmare to look at it interface builder.  There's a similar situation in the privacy dashboard header, but that's because there's images embedded in the text. 

There's going to be a bit of UI work soon so I'll create a task for this as part of that.

PR #274, opened 2018-02-20
Make AppDeepLinks quick link detection methods not case sensitive.

timonus:
Reviewer:
Asana:
CC:

**Description**:
This pull request resolves an issue I found when trying to integrate DuckDuckGo into my app [Opener](http://www.opener.link/), which relies on URL schemes to work with other apps. I was pleased to find DuckDuckGo was open source so I could diagnose and resolve the issue 😊.

Here's a video of what I was seeing (also demos the testing steps) https://db.tt/ISduGbXZOv

**Steps to test this PR**:
1. Install and launch DuckDuckGo.
2. Launch Safari.
3. Enter the URL `ddgQuickLink://apple.com` and press the 'Go' key.
4. See that DuckDuckGo launches and navigates to apple.com.

Prior to this PR the iOS normalization of URL schemes to lowercase caused this link detection to not work, so the app would launch but not navigate to the correct URL.

###### Reviewer Checklist:

###### PR DRI Checklist:

### DIFF
--- a/Core/AppDeepLinks.swift
+++ b/Core/AppDeepLinks.swift
@@ -26,7 +26,7 @@ public struct AppDeepLinks {
     
     public static let quickLink = "ddgQuickLink://"
     
-    public static let aboutLink = URL(string: "\(AppDeepLinks.quickLink)https://duckduckgo.com/about")!
+    public static let aboutLink = URL(string: "\(AppDeepLinks.quickLink)duckduckgo.com/about")!
     
     public static func isLaunch(url: URL) -> Bool {
         if let scheme = url.scheme {
--- a/Core/AppDeepLinks.swift
+++ b/Core/AppDeepLinks.swift
@@ -37,12 +37,15 @@ public struct AppDeepLinks {
     
     public static func isQuickLink(url: URL) -> Bool {
         if let scheme = url.scheme {
-            return AppDeepLinks.quickLink.contains(scheme)
+            return AppDeepLinks.quickLink.lowercased().contains(scheme.lowercased())
         }
         return false
     }
     
     public static func query(fromQuickLink url: URL) -> String {
-        return url.absoluteString.replacingOccurrences(of: quickLink, with: "")
+        let components = NSURLComponents.init(url: url, resolvingAgainstBaseURL: true)!
+        components.scheme = nil // remove scheme
+        let string = String(components.url!.absoluteString.dropFirst(2)) // strip off '//' prefix
+        return string
     }
 }
--- a/Core/AppDeepLinks.swift
+++ b/Core/AppDeepLinks.swift
@@ -43,7 +43,7 @@ public struct AppDeepLinks {
     }
     
     public static func query(fromQuickLink url: URL) -> String {
-        let components = NSURLComponents.init(url: url, resolvingAgainstBaseURL: true)!
+        let components = NSURLComponents(url: url, resolvingAgainstBaseURL: true)!
         components.scheme = nil // remove scheme
         let string = String(components.url!.absoluteString.dropFirst(2)) // strip off '//' prefix
         return string
--- a/DuckDuckGo.xcodeproj/project.pbxproj
+++ b/DuckDuckGo.xcodeproj/project.pbxproj
@@ -7,6 +7,7 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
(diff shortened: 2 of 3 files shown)


timonus:
And I apologize in advance for any Swift issues, I'm versed with Obj-C but don't know much Swift. If I've done anything wrong I'd happily update the PR with the feedback.

brindy:
Thanks for this, you definitely found a bug. :)

I notice we don't have any unit tests for this code - if this is something you feel comfortable adding, please feel free to go ahead,  otherwise I'll do it when I get chance. 

Thanks again!

Core/AppDeepLinks.swift
```diff
@@ -37,12 +37,15 @@ public struct AppDeepLinks {
...
         if let scheme = url.scheme {
-            return AppDeepLinks.quickLink.contains(scheme)
+            return AppDeepLinks.quickLink.lowercased().contains(scheme.lowercased())
         }
         return false
     }
     
     public static func query(fromQuickLink url: URL) -> String {
-        return url.absoluteString.replacingOccurrences(of: quickLink, with: "")
+        let components = NSURLComponents.init(url: url, resolvingAgainstBaseURL: true)!
```
brindy: The convention here is not to use `.init(` ... just use `NSURLComponents(` directly.

Core/AppDeepLinks.swift
```diff
@@ -37,12 +37,15 @@ public struct AppDeepLinks {
...
+        components.scheme = nil // remove scheme
+        let string = String(components.url!.absoluteString.dropFirst(2)) // strip off '//' prefix
+        return string
     }
```
brindy: Unfortunately this code breaks the link we have to our about page.  If a link comes in with a scheme then it appears to mangle it slightly, e.g. `https://duckduckgo.com/about` becomes `https//duckduckgo.com/about` (ie no colon after https) and then it doesn't go to the expected page.

brindy (changes requested):
Thanks again for this - a couple of comments for you.

brindy:
Hi @timonus - when you're happy you've finished making changes, can you please remove the "Work in Progress" label and leave a comment @ me and I'll re-review.  Thanks! :)

Core/AppDeepLinks.swift
timonus: @brindy Good catch. There's a few solutions to this I could see. I think the right thing to do is to change the value of `aboutLink` to match the expectations of the new implementation of `-queryFromQuickLink`. The two I'm thinking of are

- Changing `aboutLink` to `ddgQuickLink://duckduckgo.com/about`, which implicitly transforms it to `http` later on.
- Changing `aboutLink` to `ddgQuickLink://https%3A%2F%2Fduckduckgo.com%2Fabout`. Having the host be a fully URL encoded URL makes it so that the scheme isn't dropped, which preserves the `https`.

Which would you prefer?

timonus: Ah wait, just kidding, URL encoding doesn't seem to work. I'd opt for the first solution here, if there's something else you'd prefer please let me know.

Core/AppDeepLinks.swift
timonus: I believe this is now resolved in https://github.com/duckduckgo/iOS/pull/274/commits/695e8ca0380dd1872fdc360f2cd519f2886b2931.

Core/AppDeepLinks.swift
timonus: I believe this is resolved in https://github.com/duckduckgo/iOS/pull/274/commits/a06d9c1f8ced82ce56657162d2a5b721aedd30d2.

timonus:
Unit tests added in https://github.com/duckduckgo/iOS/pull/274/commits/bbc5dcda1957de8159c20187d2f1b217f16a30c3 😄

timonus:
@brindy I'm not sure how to remove the "Work in progress" label but I believe this it ready for re-review.

timonus:
@brindy alright, think I've handled the recent feedback in the last two commits I pushed.

timonus:
Turns out this isn't strictly needed for Opener support since the OS only normalizes URL schemes to lowercase in some situations, so I'm gonna push that out not, but I still think this is a good change to have!

timonus:
@brindy just a quick reminder on this, I _think_ this is ready to land now unless I've missed something.

brindy (approved):

brindy:
Thanks for your contribution @timonus :)

timonus:
Awesome, thanks @brindy!

PR #275, opened 2018-02-23
Core Data concurrency tweaks

bwaresiak:
Reviewer:
Asana:
CC:

**Description**:
While looking at the code, I spotted that in some places Core Data seems to be used in an unsafe way (disregarding context queue). I confirmed it by enabling Core Data debug flag on a startup. 

This PR is supposed to solve above problem by:
1. Adding mentioned above argument to scheme, so it can be easily enabled if needed (you may want to include this as a part of a development cycle, or enable it by default).
2. Perform all Core Data fetches on appropriate queues.
3. Confine ManagedObjects to queues managed by ManagedObjectContext 

One tricky case to solve was how NetworkLeaderboard is used by the UI. I decided to make it bound to the Main queue, as I don't see it being used outside of UI related code. An alternative is to introduce Model layer and return Model objects instead of ManagedObjects, but that seems to be an overkill at the moment.

I tried to check every functionality while going through the whole app, but mind that I am not really familiar with it, so I might've missed something.

Another thing to consider is to whether enable Core Data debug flag for tests or not - it may require further tweaking, this time in the tests.

**Steps to test this PR**:
1. Run application with Core Data concurrency debug flag enabled. No assertions should be fired, all functionality should work as expected.

###### Reviewer Checklist:

###### PR DRI Checklist:

### DIFF
--- a/Core/DDGPersistenceContainer.swift
+++ b/Core/DDGPersistenceContainer.swift
@@ -27,7 +27,7 @@ public class DDGPersistenceContainer {
     public let persistenceStoreCoordinator: NSPersistentStoreCoordinator
     public let managedObjectContext: NSManagedObjectContext
 
-    public init?(name: String) {
+    public init?(name: String, concurrencyType: NSManagedObjectContextConcurrencyType = .privateQueueConcurrencyType) {
         let mainBundle = Bundle.main
         let coreBundle = Bundle(identifier: "com.duckduckgo.mobile.ios.Core")!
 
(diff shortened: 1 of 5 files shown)


brindy:
This is great, thanks. I am about to go on leave for a few days, but will pick this up for review next week.  Thanks again.

brindy:
Thanks again for this @bwaresiak

bwaresiak:
@brindy You're welcome! :)

PR #277, opened 2018-02-28
minor speed improvements

brindy:
Task/Issue URL: https://app.asana.com/0/392891325557410/561256581080450
Tech Design URL:
CC: 

**Description**:

* When browsing on the same site there is no need to re-inject the scripts
* Show progress at the earliest opportunity after the user decides to navigation (ie start showing progress as soon as the decision is considered allowed)
* Drop content blocking scripts for duckduckgo.com 

**Steps to test this PR**:

***Test 1***
1. Search for cnn https://duckduckgo.com?q=cnn 
1. Expected: site should be more responsive to load
1. Expected: content blocking scripts have not been injected (use Web Inspector to confirm)
1. Expected: Site has an A grade
1. Expected: Tab has correct favicon in Tabs view

***Test 2***
1. Follow link to cnn.com
1. Expected: Progress bar starts immediately
1. Expected: Appropriate site grade is displayed 

***Test 3***
1. Click on a link on cnn.com
1. Expected: loading seems more responsive to load
1. Expected: site grade and privacy details are appropriate 

---
###### Internal references:

### DIFF
--- a/Core/WebViewController.swift
+++ b/Core/WebViewController.swift
@@ -147,7 +147,7 @@ open class WebViewController: UIViewController {
         switch(keyPath) {
 
         case webViewKeyPaths.estimatedProgress:
-            progressBar.progress = Float(webView.estimatedProgress)
+            progressBar.progress = max(0.1, Float(webView.estimatedProgress))
 
         case webViewKeyPaths.hasOnlySecureContent:
             webEventsDelegate?.webView(webView, didUpdateHasOnlySecureContent: webView.hasOnlySecureContent)
@@ -188,6 +188,7 @@ open class WebViewController: UIViewController {
     
     private func showProgressIndicator() {
         progressBar.alpha = 1
+        progressBar.progress = 0.1
     }
     
     private func hideProgressIndicator() {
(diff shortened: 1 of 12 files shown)


bwaresiak:
@brindy Speaking of small performance improvements, have you considered tweaking JavascriptLoader's `load(script:withReplacements:into:forMainFrameOnly:injectionTime:)` method? What I have in mind is:
```
for (key, value) in replacements {
    js = js.replacingOccurrences(of: key, with: value)
}
```
where, how I understand, you process string and replace all pre-set (and pre-determined) values. I figured out that adding `.literal` option wouldn't hurt:
```
js = js.replacingOccurrences(of: key, with: value, options: .literal)
```
That change reduced time to process CNN site by 100ms on my iPhone SE. I briefly re-checked the output of that code and everything seems to be correctly replaced. 

Just my $0.02 :)

brindy:
Thanks @bwaresiak - I'll take a look against the 5S. Sounds like it wouldn't do any harm anyway.

Core/WebViewController.swift
```diff
@@ -295,55 +302,56 @@ extension WebViewController: WKNavigationDelegate {
...
     
     public func webView(_ webView: WKWebView, decidePolicyFor navigationAction: WKNavigationAction, decisionHandler: @escaping (WKNavigationActionPolicy) -> Void) {
+        let decision = decidePolicyFor(navigationAction: navigationAction)
+        if decision == .allow && navigationAction.isTargettingMainFrame() {
+            showProgressIndicator()
+        }
+        decisionHandler(decision)
+    }
+    
+    private func decidePolicyFor(navigationAction: WKNavigationAction) -> WKNavigationActionPolicy {
```
subsymbolic, line 319: This logic has grown over time, can we extract this method and unit test it? Happy to capture it in another PR if that makes more sense.
Optional: Does it make sense to return a boolean here?

subsymbolic (approved):

Core/WebViewController.swift
brindy: Since there's already constants I think they are better than booleans anyway - boolean return types can often seem ambiguous, imho.

I had thought of writing tests for this but I think it's a rabbit hole in terms of dependencies. I really want to refactor this whole class in to a bunch of unit testable components as part of the work to consolidate the TabViewController/WebViewController code so would rather do it as part of that task: https://app.asana.com/0/414709148257752/542099048519218

PR #288, opened 2018-03-27
network leaderboard tweaks

brindy:
Task/Issue URL: https://app.asana.com/0/414235014887631/494552239039546
Tech Design URL:
CC:

**Description**:

Minor optimisation to the network leaderboard code and vertical centering of the top offenders.

**Steps to test this PR**:
1. Fresh install (or reset the network tracker) 
1. Visit DuckDuckGo.com
1. Open dashboard, observe vertical centering and new 'no stats' icon with still gathering message
1. Visit thehill.com
1. Visit cnn.com
1. Open dashboard, observe vertical centering of the top offenders title and pills

---
###### Internal references:


subsymbolic:
The latest changes look good. There is still an issue in landscape whereby scrolling dismisses the dashboard view.

subsymbolic:
@brindy I see that padding is still being added on smaller screens
iphone_se

DuckDuckGo/PrivacyProtectionFooterController.swift
```diff
@@ -30,14 +30,16 @@ class PrivacyProtectionFooterController: UIViewController {
     var contentBlocker: ContentBlockerConfigurationStore = ContentBlockerConfigurationUserDefaults()
 
     override func viewDidLoad() {
+        super.viewDidLoad()
         leaderboard.didLoad()
-        update()
+        // update()
```
subsymbolic: Commented out code

subsymbolic (commented):
This is getting close but still runs into an issue of overlapping views when rotating from portrait to landscape. You could create two separate views for the leaderboard states rather than grouping them in the same view:
• one for when the leaderboard is shown
• one for when "collecting data" is shown

Then you could update the footer to contain the correct one as the state changes rather than showing/hiding elements. This *may* also allow you to use the intrinsic size for the footer cell and remove other manual resizing currently performed when swapping states.

footer_view_overlap

brindy:
Thanks @subsymbolic - I've already restructured it in to a couple of different views and then add/remove the right one, but I'll see if I can do something with sizes.

DuckDuckGo/PrivacyProtectionFooterController.swift
```diff
@@ -30,24 +30,33 @@ class PrivacyProtectionFooterController: UIViewController {
...
     }
-
+    
     @IBAction func toggleProtection() {
         let contentBlockingOn = privacyProtectionSwitch.isOn
         self.contentBlocker.enabled = contentBlockingOn
         update()
     }
 
+    override func viewWillTransition(to size: CGSize, with coordinator: UIViewControllerTransitionCoordinator) {
```
subsymbolic, line 48: ViewWillAppear is getting called on rotation so I think you can delete this method.

subsymbolic (approved):
Great work. Just one cleanup, merge when you're ready.

DuckDuckGo/PrivacyProtectionFooterController.swift
brindy: Thanks @subsymbolic 

It is needed because if you reset the tracker stats and then go back to the dashboard it needs to do another update to go from the score view to the gathering view. :)

subsymbolic: In case it was unclear, I'm referring to the `viewWillTransition` method not being needed as `viewWillAppear` already does the job.

brindy: Actually, scratch that, you're right.

PR #294, opened 2018-04-10
New header design: icons, app colors and edit text mode

subsymbolic:
**Task/Issue URLs**
Icons: https://app.asana.com/0/72649045549333/614787299259405
Colors: https://app.asana.com/0/72649045549333/614787299259403
SearchBar: https://app.asana.com/0/72649045549333/614787299259402

**Description:**
Adds new header design

**Steps to test this PR:**
Run the app and check that
• The icons and colors are updated
• The edit text mode hides other icons and shows new background when in edit mode

NOTE: There is a separate task to add a "count" to the tabs button.


brindy:
Looking good!

Can the following buttons also be made the right kind of grey to make them consistent:

* Settings -> Done
* Bookmarks -> Done
* Bookmarks -> Edit

Also, on iPad (checked 9.7inch and 12.9inch) the up arrow isn't quite aligned with the centre of the privacy grade. 

screen shot 2018-04-11 at 11 44 49

subsymbolic:
@brindy good thinking regarding the button colors, they should be consistent so I have updated those. I'll let design know too. The up arrow positioning is weird, I tweaked it but iOS does some magic to decide the final positioning. In the end, I gave in to the system and just simplified the code to provide the source rect and let it decide from there. Let me know if you have more luck!

brindy:
@subsymbolic - the stack view needs to be trailing/leading relative to the safe area. 

1. in OmniBar.xib click on Search Container
1. under Size Inspector turn on Safe Area Layout Guide
1. Update the StackView's leading and trailing to be relative to the Safe Area with a constant of 0

screen shot 2018-04-11 at 20 27 07
screen shot 2018-04-11 at 20 27 39
screen shot 2018-04-11 at 20 29 03
screen shot 2018-04-11 at 20 29 10

subsymbolic:
@brindy that example works as it has dropped the leading / trailing margin constants but as soon as I add them things are a tiny bit off centre again. I think this is just iOS choosing the "ideal" positioning, the same happens with UIBarButton.

brindy (approved):
Good points on the margins.  I don't think there's much more we can do if iOS is doing this to views it controls as well.

PR #295, opened 2018-04-11
Release/7.3.0.3 - XHR fix

brindy:
Task/Issue URL: https://app.asana.com/0/392891325557410/628429840777099
Tech Design URL:
CC:

**Description**:

If an XHR resource should be blocked and we try to abort it then previously we could break the page if the abort function throws an error.

Also, XHRs need to be applied or aborted otherwise the page can get in to a weird state waiting for resources that never appear.

This fix simply wraps an abort in a try-catch to prevent page breakages.

**Steps to test this PR**:
1. Go to DuckDuckGo.com
1. Start a Javascript debugging session
1. Navigate to freep.com
1. Check the console - there should be no errors related to XHRs
1. Check a few other sites to ensure no regressions

Note that freep.com may not work entirely properly until some resources we are blocking are either whitelisted or added to the Disconnect list.  This will come later.

---
###### Internal references:

### DIFF
--- a/Core/detection.js
+++ b/Core/detection.js
@@ -86,8 +86,12 @@
             duckduckgoMessaging.log(body)
             if (duckduckgoContentBlocking.shouldBlock(this.trackerUrl, "xmlhttprequest", function(url, block) { } )) { 
                 duckduckgoMessaging.log("blocking xhr")
-                xhr.abort()
-                return 
+                try { 
+                    xhr.abort() 
+                    return 
+                } catch(error) { 
+
+                }
             }
 
             duckduckgoMessaging.log("xhr ok: " + this.trackerUrl)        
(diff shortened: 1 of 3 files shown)


Core/detection.js
```diff
@@ -86,8 +86,12 @@
             duckduckgoMessaging.log(body)
             if (duckduckgoContentBlocking.shouldBlock(this.trackerUrl, "xmlhttprequest", function(url, block) { } )) { 
                 duckduckgoMessaging.log("blocking xhr")
-                xhr.abort()
-                return 
+                try { 
+                    xhr.abort()
```
subsymbolic, line 90: I think this errors every time so I'm not sure that we are blocking XHR requests. Also, let's only send the success message back to the app when we know we have blocked something - in the error scenario we haven't blocked but we tell the app we have.

brindy:
Killing this as going to release separate to 7.3.0

PR #309, opened 2018-05-17
Address home page feedback

subsymbolic:
Task/Issue URL: https://app.asana.com/0/361428290920652/675439631468541

**Description**:
• Updates the menu icon as the hamburger was clashing on many webpages including our own SERP
• Fixes issues where home page was not showing correct UI components (for example the editing background was shown when the user was not editing)
• Fixes issue where launching a new tab from the "New Tab" menu option did not display keyboard
• Only shows Bookmark in omnibar when keyboard up and omnibar is empty to avoid redundancy of two identical buttons on screen
• Moves settings button into omnibar and updates icon to better match other icons in app

**Steps to test this PR**:
**TEST ONE:** 
1. Launch the app
1. Ensure it is initially on the home page, in a non-editing state (keyboard down, dark header)
1. Tap into the search bar and note that we switch to empty-editing mode with the keyboard, editing background and bookmark buttons showing
1. Start entering text and note that the bookmarks and settings buttons disappear
1. Tap out of the text field (by pressing anywhere non-active on the page) and note that you are back to the initial state

**TEST TWO:** 
1. Launch a new tab in each of these ways:
- from the home page by opening the TabSwitcher and pressing plus
- from the browser by opening the TabSwitcher and pressing plus
- from the browser by pressing the menu and then "New Tab"

2. Ensure that in each case a New Tab is launched, the omnibar text is cleared and the keyboard is showing

**TEST THREE:**
Go to the browser page. ensure that the three-dot menu (rather than hamburger shows)

**TEST FOUR:**
Switch between tabs and ensure that the omnibar text is updated

---
###### Internal references:


nilnilnil (approved):
I had to do some updating; but got it running on an iPhone SE emulator. My go to emu tester. Seems fine.

subsymbolic:
Thanks for the review!

PR #312, opened 2018-05-17
only update if there is a site rating

brindy:
Task/Issue URL: https://app.asana.com/0/413877547933097/487656199561602
Tech Design URL:
CC:

**Description**:

Fixes a crash bug when privacy dashboard is open.

**Steps to test this PR**:
1. Search for a site
1. Open the privacy dashboard
1. Go in to airplane mode
1. Change privacy protection
1. Drill down in to one of the detail views - should not crash

---
###### Internal references:

### DIFF
--- a/DuckDuckGo/PrivacyProtectionController.swift
+++ b/DuckDuckGo/PrivacyProtectionController.swift
@@ -111,8 +111,9 @@ class PrivacyProtectionController: UIViewController {
 
     func updateSiteRating(_ siteRating: SiteRating?) {
         self.siteRating = siteRating
+        guard let siteRating = siteRating else { return }
         omniBar.updateSiteRating(siteRating)
-        omniBar.refreshText(forUrl: siteRating?.url)
+        omniBar.refreshText(forUrl: siteRating.url)
         updateViewControllers()
     }
 


DuckDuckGo/PrivacyProtectionController.swift
```diff
@@ -111,8 +111,9 @@ class PrivacyProtectionController: UIViewController {
 
     func updateSiteRating(_ siteRating: SiteRating?) {
         self.siteRating = siteRating
```
subsymbolic, line 113: What's the nullability contract of the SiteRating in the PrivacyProtectionController? It is currently unclear and this is the source of these kind of bugs. For example, this `updateSiteRating` method accepts an optional value and sets it to the `self.siteRating` however `self.siteRating` is force unwrapped which suggests it should never be null.

Imho the force unwrap in this class is bad (Apple's fault!) as we've lost all the compiler checks that would have caught this crash. If the class level siteRating can be optional then that's great we can change it and get the checks back (and setting the optional value like you do in this line will then be valid) but if it is meant to be non-optional we should see if we can massage this class to not force unwrap somehow. Just some ideas, you may have some better ones, happy to discuss.

brindy: I'm not seeing the ambiguity - siteRating is optional, therefore can be nil (and it doesn't get force unwrapped).  

What am I missing?

brindy: We discussed this and we're happy, but I'm going to raise a feedback item regarding the fact that the privacy grade screen might not reflect the underlying web view if it errors, etc.

subsymbolic (approved):

PR #316, opened 2018-05-21
Fix layout issue in Settings screen

tiagomartinho:
The footer in the Security section would overlap when in landscape for
the iPhone X. 
This would only happen in the device and not the
simulator. 

I fix it by removing the dot at the end of the sentence.

Xcode also applied some auto corrections to the Storyboard possibily due
to previous merging conflicts.

Before:
img_2649

After:
img_2650


brindy:
Hi - thanks for your interest, it's much appreciated.

I'm afraid we've been unable to reproduce this issue.  I originally thought it might be to do with the commercial font we're using (Proxima Nova - which we can't distribute for licensing reasons) but I tried a clean build without that font and I still don't see the issue.

As such, I'm going to close this off.  

Thanks again!

tiagomartinho:
Hello @brindy!

No problem, the first screenshot was done with the app directly from the Store, so I believe the font is not a problem.

I'm running the app in a iPhone X, with iOS 11.3.1, region Spain and language Spanish.

Thanks for the answer!

brindy:
Hi @tiagomartinho - same as me, except my language is English.  I've just switched to Spanish and still seems OK.  I can see from the screenshot you're on a fairly recent version of that, so that's really strange.  I'll reach out to some of our other non-English users and see if they are having the same problem.

Thanks for the update! :)

tiagomartinho:
Thanks @brindy!

PR #320, opened 2018-05-28
xhr improvements

brindy:
Task/Issue URL: https://app.asana.com/0/569473074191537/632684929109461
Tech Design URL:
CC:

**Description**:

Lowes website was broken because we were aborting XHR requests.  This approach redirects blocked XHR requests to about:blank instead and does not interfere with the "send" function at all.  

I've been running this locally for the weekend and had no problems.

**Steps to test this PR**:

1. Visit lowes.com - it should function correctly (bug was that it was a blank screen)
1. Login in to social media to check regression in XHR (Twitter, Facebook, etc)
1. Browse other sites to check for XHR regressions

---
###### Internal references:

### DIFF
--- a/Core/detection.js
+++ b/Core/detection.js
@@ -80,13 +80,14 @@
         var xhr = XMLHttpRequest.prototype
         var originalOpen = xhr.open
 
-        xhr.open = function(method, url) {
-            var trackerUrl = null
+        xhr.open = function() {
+            var args = arguments
+            var url = arguments[1]
             duckduckgoContentBlocking.shouldBlock(url, "xmlhttprequest", function(url, block) {                                                  
-                trackerUrl = block ? "about:blank" : url 
+                args[1] = block ? "about:blank" : url 
             })
-            duckduckgoMessaging.log("sending xhr " + url + " to " + trackerUrl)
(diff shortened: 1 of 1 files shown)


Core/detection.js
```diff
@@ -79,29 +79,16 @@
 
         var xhr = XMLHttpRequest.prototype
         var originalOpen = xhr.open
-        var originalSend = xhr.send
 
         xhr.open = function(method, url) {
```
subsymbolic: I'll let Dave review this as he's the js expert, I'll jump in and test it once it is reviewed. One question that goes beyond this PR, do we need to cover other xhr open methods https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/open?

brindy: Good point. It's actually a single function with variable args, but we should use the arguments array directly, rather than reconstructing it. I'll update that.

dharb (approved):
This looks good to me. I do wonder what's different about that xhr request on lowes.com that breaks things when aborted.. Could it be that the request was synchronous, and aborting it blocked the main thread from loading? Either way, what you have here should take care of the issue.

dharb (approved):

subsymbolic (approved):
LGTM

PR #321, opened 2018-05-28
use alternative data store remove data method and retain cookies for duckduckgo

brindy:
Task/Issue URL: https://app.asana.com/0/392891325557410/681801063131984
Tech Design URL:
CC:

**Description**:

Clears all data and retains cookies for DuckDuckGo.

Data expunging should work on both iOS 10 and 11, but retaining the cookies will only work on iOS 11.

**Steps to test this PR**:

Test 1

1. Visit https://cdn.cliqz.com/browser-f/fun-demo/test-webworker-indexed-db.html
1. Refresh the page and observe counter incrementing
1. Forget all 🔥 
1. Visit https://cdn.cliqz.com/browser-f/fun-demo/test-webworker-indexed-db.html and observe counter back at 1

Test 2

1. Visit DuckDuckGo.com, apply some settings (e.g. location, safe search off)
1. Forget all 🔥 
1. Visit DuckDuckGo.com, settings should still be applied

Test 3

1. Visit DuckDuckGo.com, apply some settings (e.g. location, safe search off)
1. Forget all 🔥 
1. Stop the app
1. Launch the app
1. Visit DuckDuckGo.com, settings should still be applied

Test 4

1. Login to a site that uses cookies
1. Forget all 🔥 
1. Return to previous site
1. Should now be logged out

Test 5
1. Visit DuckDuckGo.com, apply some settings (e.g. location, safe search off)
1. Forget all 🔥 
1. Visit DuckDuckGo.com, and change a setting
1. Kill the app
1. Launch the app (latest settings should apply)

---
###### Internal references:

### DIFF
--- a/Core/WebCacheManager.swift
+++ b/Core/WebCacheManager.swift
@@ -22,45 +22,61 @@ import WebKit
 
 public class WebCacheManager {
 
+    public static var instance = WebCacheManager()
+    
     private struct Constants {
-        static let internalCache = "duckduckgo.com"
+        static let cookieDomain = "duckduckgo.com"
     }
     
-    private static var allData: Set<String> {
+    private var cookies = [HTTPCookie]()
+    
+    private var allDataTypes: Set<String> {
         return WKWebsiteDataStore.allWebsiteDataTypes()
     }
     
-    private static var dataStore: WKWebsiteDataStore {
+    private var dataStore: WKWebsiteDataStore {
         return WKWebsiteDataStore.default()
     }
     
-    /**
-     Provides a summary of the external (non-duckduckgo) cached data
-     */
-    public static func summary(completionHandler: @escaping (_ summary: WebCacheSummary) -> Void) {
-         dataStore.fetchDataRecords(ofTypes: allData, completionHandler: { records in
-            let count = records.reduce(0) { (count, record) in
-                if record.displayName == Constants.internalCache {
-                    return count
-                }
-                return count + record.dataTypes.count
-            }
-            Logger.log(text: String(format: "Web cache retrieved, there are %d items in the cache", count))
-            completionHandler(WebCacheSummary(count: count))
-        })
+    private init() {
+        self.cookies = HTTPCookieStorage.shared.cookies ?? []
     }
-    
-    /**
-     Clears the cache of all external (non-duckduckgo) data
-     */
-    public static func clear(completionHandler: @escaping () -> Void) {
-        dataStore.fetchDataRecords(ofTypes: allData) { records in
-            let externalRecords = records.filter { $0.displayName != Constants.internalCache }
-            dataStore.removeData(ofTypes: allData, for: externalRecords) {
-                Logger.log(text: "External cache cleared")
-                completionHandler()
+
+    public func injectCookies(dataStore: WKWebsiteDataStore) {
+        
+        if #available(iOS 11, *) {
+            for cookie in cookies {
+                dataStore.httpCookieStore.setCookie(cookie) {
+                    // no-op
+                }
             }
         }
     }
-}
     
+    /**
+     Clears the cache of all data, except duckduckgo cookies
+     */
+    public func clear() {
+        if #available(iOS 11, *) {
+            extractAllowedCookies(in: dataStore.httpCookieStore)
+        }
+        
+        dataStore.removeData(ofTypes: allDataTypes, modifiedSince: Date.distantPast) {
+            // no-op
+        }
+    }
+
+    @available(iOS 11, *)
+    func extractAllowedCookies(in cookieStore: WKHTTPCookieStore) {
+        
+        cookieStore.getAllCookies { (cookies) in
+            self.cookies = cookies.filter({ $0.domain == Constants.cookieDomain })
+            let storage = HTTPCookieStorage.shared
+            for cookie in self.cookies {
+                storage.setCookie(cookie)
+            }
+        }
+        
+    }
+    
+}
--- a/Core/WebCacheManager.swift
+++ b/Core/WebCacheManager.swift
@@ -22,24 +22,19 @@ import WebKit
 
 public class WebCacheManager {
 
-    public static var instance = WebCacheManager()
-    
     private struct Constants {
         static let cookieDomain = "duckduckgo.com"
     }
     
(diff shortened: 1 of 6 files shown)


Core/WebCacheManager.swift
```diff
@@ -22,45 +22,61 @@ import WebKit
...
+    private init() {
+        self.cookies = HTTPCookieStorage.shared.cookies ?? []
+    }
+
+    public func injectCookies(dataStore: WKWebsiteDataStore) {
+        
+        if #available(iOS 11, *) {
+            for cookie in cookies {
+                dataStore.httpCookieStore.setCookie(cookie) {
+                    // no-op
```
subsymbolic: The completion has a default value of nil, you can remove it and just call `dataStore.httpCookieStore.setCookie(cookie)`

Core/WebCacheManager.swift
```diff
@@ -22,45 +22,61 @@ import WebKit
...
+        
+        dataStore.removeData(ofTypes: allDataTypes, modifiedSince: Date.distantPast) {
+            // no-op
+        }
+    }
+
+    @available(iOS 11, *)
+    func extractAllowedCookies(in cookieStore: WKHTTPCookieStore) {
+        
+        cookieStore.getAllCookies { (cookies) in
```
subsymbolic: No brackets needed in `(cookies)`

Core/WebCacheManager.swift
```diff
@@ -22,45 +22,61 @@ import WebKit
...
-            dataStore.removeData(ofTypes: allData, for: externalRecords) {
-                Logger.log(text: "External cache cleared")
-                completionHandler()
+    public func clear() {
+        if #available(iOS 11, *) {
+            extractAllowedCookies(in: dataStore.httpCookieStore)
+        }
+        
+        dataStore.removeData(ofTypes: allDataTypes, modifiedSince: Date.distantPast) {
+            // no-op
```
subsymbolic: Why is this completion empty? Would it make sense to simplify the code by storing cookies here  and then putting them back during completion? My guess is that you've already tried this, if you have ignore!

**Optional:** I would suggest killing the no-op, with `{}` on the same line for an empty non-optional completion.

DuckDuckGo/MainViewController.swift
```diff
@@ -568,3 +577,26 @@ extension MainViewController: BookmarksDelegate {
     }
 }
 
+// This preloads a webview so that the datastore can then receive cookies, otherwise nothing happens until subsequent webviews are shown
```
subsymbolic: Apple is making us do some interesting workarounds today! There's a comment on the 9th of May in your link that mentions a workaround by accessing the data store to for lazy creation. Is that something we can do instead of this? We could add that to the WKWebViewConfiguartionExtension.

subsymbolic (commented):
Great stuff, works well. Just a couple of questions/comments.

Core/WebCacheManager.swift
brindy: Yep that didn’t work sadly :(

DuckDuckGo/MainViewController.swift
brindy: I tried that too, didn’t work for me :(

brindy: Gonna try it again though to be 💯 sure

brindy: Thanks for the zoom call. I've managed to clean this up so it's not needed, at least.

Core/WebCacheManager.swift
```diff
@@ -22,45 +22,57 @@ import WebKit
...
      */
-    public static func clear(completionHandler: @escaping () -> Void) {
-        dataStore.fetchDataRecords(ofTypes: allData) { records in
-            let externalRecords = records.filter { $0.displayName != Constants.internalCache }
-            dataStore.removeData(ofTypes: allData, for: externalRecords) {
-                Logger.log(text: "External cache cleared")
-                completionHandler()
+    public func clear() {
+        if #available(iOS 11, *) {
+            extractAllowedCookies(in: dataStore.httpCookieStore)
```
subsymbolic: I think this could be problematic because we move straight onto the clearing before the completion in `extractAllowedCookies` method is done

subsymbolic (commented):
@brindy I've just noticed that when I press fire the cookies are saved and they overwrite future cookie changes that I make. This means that if I press fire, change my cookies and then kill the app, my recent changes are lost.

Core/WebCacheManager.swift
brindy: Good spot - I've updated the tests above to cover this as well.

subsymbolic (approved):
Looks good!

nilnilnil:
Why singleton?

brindy:
That's a good point @nilnilnil - it doesn't need to be. (I had introduced a singleton when I was holding on to some in memory state, but that's gone)

subsymbolic (approved):
LGTM

brindy:
Just found a bug in it - seems to lose settings when opening a new tab.

brindy:
Seems to be something flakey with the use of HTTPCookieStorage 🙄  ... am switching to storing them in UserDefaults instead.

Core/WebCacheManager.swift
```diff
@@ -23,44 +23,49 @@ import WebKit
...
+    @available(iOS 11, *)
+    private static func extractAllowedCookiesThenClear(in cookieStore: WKHTTPCookieStore) {
+        let cookieStorage = CookieStorage()
+        cookieStore.getAllCookies { cookies in
+            let cookies = cookies.filter({ $0.domain == Constants.cookieDomain })
+            for cookie in cookies {
+                cookieStorage.setCookie(cookie)
+                
             }
+            self.dataStore.removeData(ofTypes: self.allDataTypes, modifiedSince: Date.distantPast) {}
```
subsymbolic: Can we change the `self.dataStore` to `WebCacheManager.dataStore` to make it more apparent that this is a static call and we don't need to worry about captures and memory leaks?

subsymbolic (approved):
Work great @brindy. Just one last suggestion, feel free to merge.

brindy:
Thanks @subsymbolic !

PR #326, opened 2018-06-11
hotfix/7.6.1.2 - fix crash and error flash

brindy:
Task/Issue URL: https://app.asana.com/0/414235014887631/699825463344755
Tech Design URL:
CC:

**Description**:

During testing of 7.6.1.1 discovered a crash when using the fire button and observed an error "flash" (which has been there since before this but seems more obvious now)

**Steps to test this PR**:

Crash:
1. Use the fire button
1. Tap the tabs button
1. Use the fire button
1. Tap the tabs button
1. Use the fire button
1. (repeat a few times)
1. Should be no crash

Error flash
1. Use the fire button
1. Kill the app
1. Launch the app and load a non-DDG bookmark
1. Search for BBC
1. Tap top link
1. Should be no error "flash"
1. Ensure regular errors are handled correctly (e.g. invalid URL, no network, etc)

---
###### Internal references:


subsymbolic:
@brindy what caused (and subsequently fixed) the crash? I see some timing changes but it doesn't jump out at me.

Core/WebCacheManager.swift
```diff
@@ -64,7 +64,10 @@ public class WebCacheManager {
                 cookieStorage.setCookie(cookie)
                 
             }
-            WebCacheManager.dataStore.removeData(ofTypes: self.allDataTypes, modifiedSince: Date.distantPast) {}
+            
+            DispatchQueue.main.async {
```
brindy, line 68: @subsymbolic This fixes the crash.  The code below wasn't running on the UI thread but only seemed to cause the crash when you used the fire button on the tabs screen after a couple of attempts. Easy to replicate it though.

subsymbolic: My questions is why? I see a timing change, but don't yet understand why it would fix the crash. My concern is it might mask it/move the timing rather than fix the root cause. Happy to zoom to discuss.

brindy: When I debugged the crash the logs told me explicitly that that code needs to be run on the UI thread and that it wasn't.

subsymbolic: Ahh thanks @brindy, that makes sense, testing it now.

subsymbolic:
fire_from_tabs

subsymbolic (approved):

PR #331, opened 2018-06-15
home row 1 rollout

brindy:
Task/Issue URL: https://app.asana.com/0/414235014887631/709078800235903
Tech Design URL:
CC:

**Description**:

* Roll out experiment 1 of the home row to all users. Existing users will now see the CTA and a reminder after 3 days.
* Some refactoring (renaming) of classes
* Place holders for subsequent experiment

**Steps to test this PR**:

Upgrade

1. Install an old version from Test Flight and launch it - get through onboarding and then kill the app
1. Install this version - ensure the app launches and runs correctly and can see CTA
1. If possible, fast forward time on device by 3 days and relaunch
1. Perform a search, should see reminder

New Install

1. Clean install of the app
1. After onboarding can see CTA
1. If possible, fast forward time on device by 3 days and relaunch
1. Perform a search, should see reminder

---
###### Internal references:

### DIFF
--- a/Core/VariantManager.swift
+++ b/Core/VariantManager.swift
@@ -21,16 +21,14 @@ import Foundation
 
 public enum FeatureName {
     
-    case homeRowOnboarding, homeRowReminder
+    case homeRowCTADefault, homeRowCTAAlternative1, homeRowCTAAlternative2
     
 }
 
 public struct Variant {
     
-    public static let defaultVariants = [
-        Variant(name: "mb", percent: 50, features: []),
-        Variant(name: "mc", percent: 25, features: [.homeRowOnboarding]),
-        Variant(name: "md", percent: 25, features: [.homeRowOnboarding, .homeRowReminder]),
+    public static let defaultVariants: [Variant] = [
+        // no experiments just now
     ]
     
     public let name: String
--- a/DuckDuckGo/HomeViewController.swift
+++ b/DuckDuckGo/HomeViewController.swift
@@ -50,10 +50,13 @@ class HomeViewController: UIViewController {
     override func viewWillAppear(_ animated: Bool) {
         super.viewWillAppear(animated)
         
-        let feature = HomeRowOnboarding()
-        if !feature.showNow() {
+        let feature = HomeRowCTA()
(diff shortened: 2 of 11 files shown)


DuckDuckGo/HomeViewController.swift
```diff
@@ -110,6 +113,10 @@ class HomeViewController: UIViewController {
         infoViewHeight.constant = 0
     }
     
+    private func applyHomeRowCTA(type: HomeRowCTA.CTAType) {
```
CDRussell, line 116: **opinion, not a request**

maybe a personal style thing, but i wouldn't have stub methods like this in the code. for one, "experiment 2" implies a lot of hidden knowledge that isn't obvious from the code.

when experiment 2 starts, i'd add the function in but not before - because there's always the chance that "experiment 2" will never start and you'd be the only person confident to remove this function/comment.

Core/VariantManager.swift
```diff
@@ -21,16 +21,14 @@ import Foundation
...
     
 }
 
 public struct Variant {
     
-    public static let defaultVariants = [
-        Variant(name: "mb", percent: 50, features: []),
-        Variant(name: "mc", percent: 25, features: [.homeRowOnboarding]),
-        Variant(name: "md", percent: 25, features: [.homeRowOnboarding, .homeRowReminder]),
+    public static let defaultVariants: [Variant] = [
```
CDRussell, line 30: Do you need to migrate existing users' variants to a default state? e.g., users who were previously `mc`, should they now lose that variant?

DuckDuckGo/HomeViewController.swift
brindy: Generally agree, but we're definitely doing experiment 2 :)

Core/VariantManager.swift
brindy: If I understand you, that's been done in #332

CDRussell: Yeah, that's what I was asking.

CDRussell (approved):
Tested the out described scenarios and is working as expected

PR #337, opened 2018-06-27
Add feedback and broken site form

subsymbolic:
Task/Issue URL: https://app.asana.com/0/361428290920652/688515420618260
Tech Design URL: N/A

**Description**:
Add a native feedback form for general and broken sites feedback.

**Steps to test this PR**:
TEST 1: Broken site:
1. From the browser page, select the "Report Broken Site"
1. Enter text in the comment box and note that the submit button becomes active
1. Submit the feedback
1. Ensure it appears in https://app.asana.com/0/569473074191537/list
1. Check the url and device details in the Asana task are correct
1. Now delete it from Asana!

TEST 2:  Feedback Message:
1. From the settings page, select "Leave Feedback"
1. Note that the submit button is inactive
1. Enter message text 
1. Submit the feedback
1. Ensure it appears in https://app.asana.com/0/413877547933097/list
1. Check the message and device details in the Asana task are correct
1. Now delete it from Asana!

---
###### Internal references:


DuckDuckGo/FeedbackViewController.swift
```diff
@@ -0,0 +1,179 @@
...
+                                               name: NSNotification.Name.UIKeyboardDidShow,
+                                               object: nil)
+        
+        NotificationCenter.default.addObserver(self,
+                                               selector: #selector(keyboardWillHide),
+                                               name: NSNotification.Name.UIKeyboardWillHide,
+                                               object: nil)
+    }
+    
+    @objc private func keyboardDidShow(notification: NSNotification) {
```
brindy, line 84: Not sure if this is working as desired. On iPhone 8 Plus and iPhone SE simulators there's no movement of the field so it leaves a small input area and you have to manually scroll up to submit.  So it feels a bit wrong.

 e.g. 

simulator screen shot - iphone se - 2018-06-27 at 10 30 27

However, I expected it to scroll up like this:

simulator screen shot - iphone se - 2018-06-27 at 10 30 37

DuckDuckGoTests/FeedbackModelTests.swift
```diff
@@ -0,0 +1,165 @@
+//
+//  FeedbackModelTests.swift
+//  DuckDuckGo
+//
+//  Copyright © 2017 DuckDuckGo. All rights reserved.
```
brindy, line 5: Should we be putting 2018 in our copy right now?  I probably do this as well tbh :)

DuckDuckGo/FeedbackViewController.swift
brindy, line 84: Also would it be possible to tap out of the fields to hide the keyboard? 

That feels a bit more intuitive than scrolling to me.

brindy: Another example: 
simulator screen shot - iphone se - 2018-06-27 at 10 44 50

brindy (commented):
Code looks good on the whole and it works.  The UX doesn't seem quite right though, with regards to showing and hiding the keyboards. Comments/images inline.

brindy:
One more thing - when place holder text changes for the message it loses the line height (text attributes, etc)

DuckDuckGo/FeedbackViewController.swift
subsymbolic: Nice idea, done!

subsymbolic: @brindy and I discussed this and it is sadly the default behavior to move the field just enough so that typing can happen. We will run with it and if we get feedback we will revisit.

subsymbolic:
@brindy this is ready for rereview

brindy (approved):
LGTM!

PR #339, opened 2018-07-05
Tweak version formatting

subsymbolic:
Task/Issue URL: https://app.asana.com/0/inbox/361428290920650/731096826873594/733111159210424

**Description**:
• Tweak format of build number in version text to contain a dot instead of brackets
• Remove app name from feedback version

**Steps to test this PR**:
TEST ONE:
1. Open the feedback form and submit feedback
1. Check the feedback in Asana  to ensure a) the title does not contain the app name b) the build number is formatted with dots rather than brackets e.g 7.8.9.1 rather than 7.8.9 (1)
1. Delete test Asana Feedback task!

TEST TWO:
1. Open settings
1. Again, note that the build number is formatted with dots rather than brackets e.g 7.8.9.1 rather than 7.8.9 (1)

---
###### Internal references:

### DIFF
--- a/DuckDuckGo/AppVersionExtension.swift
+++ b/DuckDuckGo/AppVersionExtension.swift
@@ -22,11 +22,11 @@ import Core
 
 extension AppVersion {
     
-    public var versionNumberAndBuild: String {
+    public var versionAndBuildNumber: String {
         return "\(versionNumber).\(buildNumber)"
     }
     
     public var localized: String {
-        return "\(name) \(versionNumberAndBuild)"
+        return "\(name) \(versionAndBuildNumber)"
     }
 }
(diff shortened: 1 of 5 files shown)


DuckDuckGo/AppVersionExtension.swift
```diff
@@ -22,11 +22,11 @@ import Core
 
 extension AppVersion {
     
+    public var versionNumberAndBuild: String {
```
brindy: optional: how about `versionAndBuildNumber` instead?  Or `versionNumberAndBuildNumber` if you want to be super accurate?

brindy (approved):
LGTM, one optional item

DuckDuckGo/AppVersionExtension.swift
subsymbolic: I like that, updating now.

PR #343, opened 2018-07-13
app metrics

brindy:
Task/Issue URL: 

* Implement pixel request: https://app.asana.com/0/392891325557410/738818866019757
* Grafana parameters: https://app.asana.com/0/392891325557410/751512552599843

Tech Design URL: https://app.asana.com/0/392891325557410/726260612891763
CC:

**Description**:

* When user launches the app, executes forget all, or opens the privacy dashboard a pixel request is triggered.
* A pixel request is in the following format https://improving.duckduckgo.com/t/{pixel}_{platform}_{form_factor}?atb={atb}
  * atb: a valid atb or blank for the very first launch
  * platform: `ios`
  * form factor: `phone` or `tablet` as determined by the device being used

**Steps to test this PR**:

***Pixel requests***:
1. Open kibana and use this query: `(pixel_id:ml) OR (pixel_id:mf) OR (pixel_id:mp)`
1. Launch the app, wait a few seconds and run the query, the ml request should appear
1. Perform a forget all while browsing or from the home screen, wait a few seconds and run the query, the mf request should appear
1. Perform a forget all in the tab switcher, wait a few seconds and run the query, the mf request should appear
1. Open the privacy dashboard, wait a few seconds and run the query, the mf request should appear

---
###### Internal references:

### DIFF
--- a/Core/AppUrls.swift
+++ b/Core/AppUrls.swift
@@ -41,7 +41,7 @@ public struct AppUrls {
         static let feedback = "\(home)/feedback.js?type=app-feedback"
         static let faviconService = "\(home)/ip3/%@.ico"
         
-        static let pixel = "https://improving.duckduckgo.com/t/%@?p=ios&f=%@"
+        static let pixel = "https://improving.duckduckgo.com/t/%@_ios_%@"
     }
 
     private struct Param {
--- a/Core/AppUrls.swift
+++ b/Core/AppUrls.swift
@@ -41,7 +41,7 @@ public struct AppUrls {
         static let feedback = "\(home)/feedback.js?type=app-feedback"
         static let faviconService = "\(home)/ip3/%@.ico"
         
-        static let pixel = "https://improving.duckduckgo.com/t/%@"
+        static let pixel = "https://improving.duckduckgo.com/t/%@?p=ios&f=%@"
     }
 
     private struct Param {
@@ -175,8 +175,9 @@ public struct AppUrls {
         return true
     }
     
-    public func pixelUrl(forPixelNamed pixelName: String) -> URL {
-        var url = URL(string: Url.pixel.format(arguments: pixelName))!
+    public func pixelUrl(forPixelNamed pixelName: String, deviceType: UIUserInterfaceIdiom = UIDevice.current.userInterfaceIdiom) -> URL {
+        let formFactor = deviceType == .pad ? "tablet" : "phone"
+        var url = URL(string: Url.pixel.format(arguments: pixelName, formFactor))!
         url = url.addParam(name: Param.atb, value: statisticsStore.atbWithVariant ?? "")
         return url
     }
--- a/Core/AppUrls.swift
+++ b/Core/AppUrls.swift
@@ -177,7 +177,9 @@ public struct AppUrls {
     }
     
     public func pixelUrl(forPixelNamed pixelName: String) -> URL {
-        return applyStatsParams(for: URL(string: Url.pixel.format(arguments: pixelName))!)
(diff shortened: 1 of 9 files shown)


brindy:
@subsymbolic can you take a quick look at this, please - just want to check the general approach is ok. I'll be looking for a detailed review once I've done the opt out/survey tasks.  Thanks!

Core/Pixel.swift
```diff
@@ -0,0 +1,53 @@
...
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+import Alamofire
+
+public enum PixelDefinition: String {
```
subsymbolic: The meaning of `PixelDefinition` isn't clear to me. Do you think `PixelCode` would make sense?

brindy: I think that's just as ambiguous.  How about `PixelName`?

Core/Pixel.swift
```diff
@@ -0,0 +1,53 @@
...
+    
+}
+
+public class Pixel: PixelFiring {
+
+    private static let pixel = Pixel()
+    
+    private let appUrls = AppUrls()
+    
+    public static func shared() -> Pixel {
```
subsymbolic: Do you need this? Why not expose the property directly?

brindy: Was future proofing as I didn't know if I'd have to do batching. That's not happening now - these functions could just be static even?

subsymbolic: I'm not clear on how that helps future proof. If you're keeping the shared property, I would like to understand. But it sounds like you're removing it, in which case, it doesn't matter.

brindy: I guess using `shared()` is old-school now. It's a standard pattern and hides implementation details so that you can defer architectural changes until later, is all. I am just pondering what the implications of using the static var directly are if any - guess you can keep the same API and still change the implementation using a dynamic getter.  Anyway yeah, I think I'm going to ditch it.

subsymbolic: That's exactly where my thinking was too - because we're in Swift land we don't need to wrap things in methods to protect against future changes.

brindy:
@subsymbolic - This is now ready to review, thanks!

Core/AppUrls.swift
```diff
@@ -40,6 +40,8 @@ public struct AppUrls {
         static let exti = "\(home)/exti/"
         static let feedback = "\(home)/feedback.js?type=app-feedback"
         static let faviconService = "\(home)/ip3/%@.ico"
+        
+        static let pixel = "https://improving.duckduckgo.com/t/%@_ios_%@"
```
subsymbolic: Can we replace `duckduckgo.com` with `\(base)`?

Core/AppUrls.swift
```diff
@@ -172,5 +174,12 @@ public struct AppUrls {
         }
         return true
     }
+    
+    public func pixelUrl(forPixelNamed pixelName: String, deviceType: UIUserInterfaceIdiom = UIDevice.current.userInterfaceIdiom) -> URL {
+        let formFactor = deviceType == .pad ? "tablet" : "phone"
```
subsymbolic: Can we pull the strings into constants?

Optional: Can we move `let formFactor = deviceType == .pad ? "tablet" : "phone"` into `Pixel.swift`. It feels like it should provide the string it requires for a particular form factor.

subsymbolic (approved):
Looks good. A couple of small suggestions, merge when you're ready.

PR #347, opened 2018-07-26
add fatal error if tearDown called twice

brindy:
Task/Issue URL: https://app.asana.com/0/414235014887631/757819156125520
Tech Design URL:
CC:

**Description**:

Attempts to fix a crash caused when the WKWebView process crashes where it appears the Notification Center is unable to remove observers.

**Steps to test this PR**:

On a physical device:
1. Visit https://www.falkirkrpg.org.uk/noxwall/js.html
1. Press the button until the WebView crashes 
1. Confirm page reloads and app continues without error
1. Confirm no memory leaks
1. Browse, open new tabs, forget all a few times to ensure no regression

---
###### Internal references:

### DIFF
--- a/Core/WebViewController.swift
+++ b/Core/WebViewController.swift
@@ -56,6 +56,8 @@ open class WebViewController: UIViewController {
     private lazy var httpsUpgrade = HTTPSUpgrade()
     private lazy var tld = TLD()
 
+    private var tearDownCount = 0
+    
     public var name: String? {
         return webView.title
     }
@@ -252,10 +254,10 @@ open class WebViewController: UIViewController {
     }
     
     public func tearDown() {
-        if #available(iOS 11, *) {
-            // prevents a crash if the observers were not there for some reason (e.g. WKWebView process crashed)
-            addObservers()
+        guard tearDownCount == 0 else {
+            fatalError("tearDown has already happened")
         }
+        tearDownCount += 1
         removeObservers()
         webView.removeFromSuperview()
         webEventsDelegate?.detached(webView: webView)
--- a/Core/WebViewController.swift
+++ b/Core/WebViewController.swift
@@ -252,7 +252,10 @@ open class WebViewController: UIViewController {
     }
     
     public func tearDown() {
-        addObservers() // prevents a crash if the observers were not there for some reason (e.g. WKWebView process crashed)
+        if #available(iOS 11, *) {
+            // prevents a crash if the observers were not there for some reason (e.g. WKWebView process crashed)
+            addObservers()
+        }
         removeObservers()
         webView.removeFromSuperview()
         webEventsDelegate?.detached(webView: webView)
--- a/Core/WebViewController.swift
+++ b/Core/WebViewController.swift
@@ -93,17 +93,21 @@ open class WebViewController: UIViewController {
         shouldReloadOnError = true
     }
 
+    fileprivate func addObservers() {
+        webView.addObserver(self, forKeyPath: #keyPath(WKWebView.estimatedProgress), options: .new, context: nil)
+        webView.addObserver(self, forKeyPath: #keyPath(WKWebView.hasOnlySecureContent), options: .new, context: nil)
+        webView.addObserver(self, forKeyPath: #keyPath(WKWebView.url), options: .new, context: nil)
+        webView.addObserver(self, forKeyPath: #keyPath(WKWebView.canGoBack), options: .new, context: nil)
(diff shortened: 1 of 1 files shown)


brindy:
@CDRussell if you don't have a physical device, then feel free to just to regression test browsing.  Thanks in advance!

Core/WebViewController.swift
```diff
@@ -239,12 +243,17 @@ open class WebViewController: UIViewController {
...
+    fileprivate func removeObservers() {
         webView.removeObserver(self, forKeyPath: #keyPath(WKWebView.estimatedProgress))
         webView.removeObserver(self, forKeyPath: #keyPath(WKWebView.hasOnlySecureContent))
         webView.removeObserver(self, forKeyPath: #keyPath(WKWebView.url))
         webView.removeObserver(self, forKeyPath: #keyPath(WKWebView.canGoForward))
         webView.removeObserver(self, forKeyPath: #keyPath(WKWebView.canGoBack))
+    }
+    
+    public func tearDown() {
+        addObservers() // prevents a crash if the observers were not there for some reason (e.g. WKWebView process crashed)
```
subsymbolic: You're adding two sets of observers and only removing one, this will crash every time the webview is closed - e.g when the user closes a tab.

subsymbolic (changes requested):
@brindy sadly the solution causes a crash whenever a tab is closed. Could you reproduce the original crash? I didn't manage it but my _guess_ is that it happens if:

1. the webview is destroyed before observers are set
2. teardown is called twice

If we're not sure which scenario is causing the crash we can add a couple of assertions checking against the two fatal conditions - the code was going to crash anyway and this way we can get informative stacktraces to help us identify and resolve the issue.

Core/WebViewController.swift
brindy: Did you actually observe that? I've been running this on my device since I made the change and haven't seen that. I stepped over and memory profiled that code as well.

brindy:
> the webview is destroyed before observers are set

How would that be possible? The web view is created, assigned and observers set all in the attachWebView function.

> teardown is called twice

This is what this patch is protecting against. Rather than having logic to say "is tore down" (remember the webView used to get nilled for this reason and we removed that) then instead give the web view some observers to actually remove.

Core/WebViewController.swift
subsymbolic: Oh interesting, I'm only seeing the crash on iOS 10, Apple must have fixed it for iOS 11. Good move Apple, I always found it annoying, especially since there's no way to "check" if a KVO observer is set.

brindy: I was just downloading a 10.x simulator to test it on... 🙄

brindy: OK, just pushed a commit that wraps this `addObservers` in a `if #available`

subsymbolic:
> How would that be possible? 

It may not be, I was just listing theories. If you've looked into it and isn't, I'm happy to close the book on that one.

> This is what this patch is protecting against. Rather than having logic to say "is tore down" (remember the webView used to get nilled for this reason and we removed that) then instead give the web view some observers to actually remove.

Putting the crash aside, I would still prefer logic which checks if tearDown has happened, as that is the root cause and is more readable. One more question, do we know why tearDown is getting called twice? Is that normal or is that the bug we need to squash?

brindy:
Sorry, no, I'm confusing things here.  My theory is that it's not being called twice (I see no evidence of it in stack traces).  What I think is happening is the wk process is crashing, recovers, but without observers, so when tearDown is actually called as part of the `process did terminate`, because there are no observers, it crashes.

brindy:
screen shot 2018-07-26 at 13 17 15

Core/WebViewController.swift
subsymbolic: Let's grab a zoom call and dig into this more on Monday.

brindy:
Since I've been unable to replicate it I was trying to put something in that wouldn't do any harm, but might also fix it - and it'll be obvious if it's worked because those crashes will reduce to zero after the next release.

Core/WebViewController.swift
brindy: tbh, this was supposed to be a quick harmless speculative fix - if it's anything more than that (and it already is) then I've failed to achieve the objective.  Better to can it and try to replicate, but that'll take time so probably needs spinning up as a project.

brindy:
Thanks for the Zoom call earlier @subsymbolic  - I've updated this to guard for multiple tear downs and generate a fatal error instead so hopefully we'll see that in crash reports if it's the cause.

subsymbolic (approved):
Nice work @brindy.

PR #352, opened 2018-08-06
handle custom url schemes

brindy:
Task/Issue URL: https://app.asana.com/0/193817002363848/705214455305866
Tech Design URL:
CC:

**Description**:

Prompts the user when a custom url scheme is detected

**Steps to test this PR**:
1. With Chrome and/or Spotify installed visit falkirkrpg.org.uk/noxwall/custom.html
1. Tap on one of the links. Should be prompted about leaving the app.
1. Tap yes.  System dialog opens to prompt you again. Tap yes.  Opens in other app
1. Repeat with other link.  Tap no.  Nothing happens.
1. Browse sites to perform a regression test

---
###### Internal references:

### DIFF
--- a/Core/SupportedExternalURLScheme.swift
+++ b/Core/SupportedExternalURLScheme.swift
@@ -21,7 +21,7 @@ import Foundation
 
 public struct SupportedExternalURLScheme {
 
-    static let schemes = [
+    static let defaultSchemes = [
         "tel",
         "mailto",
         "maps",
@@ -32,10 +32,20 @@ public struct SupportedExternalURLScheme {
         "itms-appss",
         "itunes"
     ]
+    
+    static var supportedSchemes = defaultSchemes
 
+    public static func addScheme(_ scheme: String) {
+        supportedSchemes.append(scheme)
+    }
+    
     public static func isSupported(url: URL) -> Bool {
         guard let scheme = url.scheme else { return false }
-        return schemes.contains(scheme)
+        return supportedSchemes.contains(scheme)
+    }
+    
+    public static func reset() {
+        supportedSchemes = defaultSchemes
     }
 
 }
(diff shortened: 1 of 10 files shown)


Core/SupportedExternalURLScheme.swift
```diff
@@ -32,10 +32,15 @@ public struct SupportedExternalURLScheme {
...
     public static func isSupported(url: URL) -> Bool {
         guard let scheme = url.scheme else { return false }
-        return schemes.contains(scheme)
+        return defaultSchemes.contains(scheme)
     }
-
+    
+    public static func isProhibited(url: URL) -> Bool {
+        guard let scheme = url.scheme else { return true }
+        return scheme == "about"
```
subsymbolic: Can we follow a similar pattern to `defaultSchemes` here?

subsymbolic (commented):
What's the expected behavior if the url cannot be opened externally e.g because there is no app to handle it? Do we show an error?

Core/SupportedExternalURLScheme.swift
brindy: done

brindy:
@subsymbolic good call - added a toast when the open externally fails

DuckDuckGo/TabViewController.swift
```diff
@@ -333,7 +333,9 @@ class TabViewController: WebViewController {
     }
     
     private func openExternally(url: URL) {
-        UIApplication.shared.open(url, options: [:])
+        UIApplication.shared.open(url, options: [:]) { completion in
+            self.view.showBottomToast(UserText.failedToOpenExternally)
```
subsymbolic: This displays even when an app can handle the scheme. There's some confusion here. The parameter named `completion` is a bool indicating success or failure. Maybe rename it to `success` and then only display the toast if `!success`?

subsymbolic (approved):
Looks good!

PR #354, opened 2018-08-22
new background tab

brindy:
Task/Issue URL: https://app.asana.com/0/392891325557410/770186107328906
Tech Design URL:
CC:

**Description**:

Open tab in new background.

Note, waiting on final animation, but that should just be a change to the animation JSON.

**Steps to test this PR**:

1. Long press a link and select "open in background tab"

  * tab switcher button animates
  * number of tabs increments
  * tab switcher button shows correct number of tabs and unread tab dot

2. Open tab switcher

  * current tab has blue border
  * other tabs are slightly transparent
  * unread tabs have a blue dot
  * tab title might be the domain name with `www.` prefix stripped if the tab hasn't had chance to load yet

3. Select tab

 * Tab opens
 * Number of tabs remains the same
 * If there are no unread tabs the dot is removed
 * Open tab switcher -> selected tab now has blue border and others are slightly transparent

4. Confirm that open in background tab pixel has been fired (`mlp_b`)

---
###### Internal references:


Core/UIColorExtension.swift
```diff
@@ -64,6 +68,10 @@ extension UIColor {
     public static var cornflowerBlue: UIColor {
         return UIColor(red: 103.0 / 255.0, green: 143.0 / 255.0, blue: 255.0 / 255.0, alpha: 1.0)
     }
+    
+    public static var skyBlue: UIColor {
```
subsymbolic, line 72: This is rendering too dark

subsymbolic (commented):
@brindy don't forget that colors in code don't exactly match zeplin sRGB colors. You'll need to convert them. Ping me if you wat to talk it through

Core/UIColorExtension.swift
brindy: I use was using `cornflowerBlue` by mistake.

brindy:
According to this the default is sRGB but I've updated the colour of skyBlue as discussed (by choosing sRGB profile in the colour picker, entering the values and then selecting the Generic RGB).  Both look the same to me though :) 

https://developer.apple.com/design/human-interface-guidelines/ios/visual-design/color/

DuckDuckGo/TabIconMaker.swift
```diff
@@ -1,65 +0,0 @@
...
-        static let fontSize: CGFloat = 10
-        static let fontWeight: CGFloat = 5
-        static let xTextOffset: CGFloat = -2
-        static let yTextOffset: CGFloat = 7.5
-        static let maxTextTabs = 100
-    }
-
-    func icon(forTabs count: Int) -> UIImage {
-        let image = #imageLiteral(resourceName: "Tabs")
-        let text = count < Constants.maxTextTabs ? "\(count)" : "🦆"
```
subsymbolic, line 36: 😭

DuckDuckGo/TabSwitcherButton.swift
```diff
@@ -0,0 +1,139 @@
...
+    let label = UILabel()
+    
+    var tabCount: Int = 0 {
+        didSet {
+            if tabCount == 0 {
+                label.text = nil
+                return
+            }
+            
+            if tabCount >= Constants.maxTextTabs {
```
subsymbolic: This isn't rendering quite right. As it is no longer an image, can we render it in the attirbuted string below.
max_text_tabs

subsymbolic (commented):
Looks great, I love being able to open tabs in the background and the blue dot to show which one was opened makes such a usability difference. Just one last UI glitch.

subsymbolic (approved):
Looks great!

PR #355, opened 2018-08-22
Add large scale https updates with bloom filter

subsymbolic:
Task/Issue URL: https://app.asana.com/0/72649045549333/721626625236431
Tech Design URL: https://app.asana.com/0/72649045549333/727813825116580

**Description**:
Add large scale https support:
- Remove legacy https solution
- Add updated https solution
- Add new api requests 
- Update core data model and persitence layer to support new data

**Steps to test this PR**:

TEST ONE:
1. Visit http://example.com
1. Check the logs and ensure they report that this site is "upgradable"
1. Note that the scheme is updated to https

TEST TWO:
1. Visit http://aaaaaaaaaaaaa.co.uk
1. Check the logs and ensure they report that this site is not "upgradable"
1. Note that the scheme remains as "http"
1. The site itself doesn't exist (thus definitely not in our upgrade list), so won't load

TEST THREE:
1. Visit any site from https://staticcdn.duckduckgo.com/https/https-mobile-whitelist.json
1. Note the logs stats that it is whitelisted and cannot be autoupgraded
1. Note that the scheme remains as "http" (beware some sites will upgrade themselves!)

TEST FOUR:

1. Install the app from develop and run it
2. Then install this version over the top
3. Ensure that nothing in the db goes boom!

---
###### Internal references:


subsymbolic:
I am seeing an atb test failure on CI (but not locally) which I am investigating. Other than that, this is ready for review @brindy.

subsymbolic:
@brindy and I are seeing the ATB Test failures on other branches too so have ruled out it being introduced by this PR. We will investigate it separately.

Core/HTTPSTransientBloomFilterSpecification.swift
```diff
@@ -0,0 +1,34 @@
...
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+
+public struct HTTPSTransientBloomFilterSpecification {
+    let totalEntries: Int
```
brindy: This could be `Decodable` which would make the JSON parsing a little easier.  Can use optionals if you're worried about data missing in future versions of the file.  Data in the JSON that isn't handled is ignored by default.

Core/HTTPSUpgradeParser.swift
```diff
@@ -0,0 +1,39 @@
...
+
+    static func whitelist(fromJSONData data: Data) -> [String]? {
+        guard let json = try? JSONSerialization.jsonObject(with: data, options: []) else { return nil }
+        guard let jsonObject = json as? [String: Any] else { return nil }
+        guard let domains = jsonObject["data"] as? [String] else { return nil }
+        return domains
+    }
+    
+    static func bloomFilterSpecification(fromJSONData data: Data) -> HTTPSTransientBloomFilterSpecification? {
+        guard let json = try? JSONSerialization.jsonObject(with: data, options: []) else { return nil }
```
brindy: If `HTTPSTransientBloomFilterSpecification` was `Decodable` you could just use the `JSONDecoder`.  Likewise for this whitelist.

We've got a couple of examples in the code already, though we also have examples of `JSONSeralization`, but I'd like us to move to the `Codable` pattern as our standard.  I'll add a LHF task to update them.

Core/HTTPSUpgrade.xcdatamodeld/HTTPSUpgrade.xcdatamodel/contents
```diff
@@ -1,19 +1,18 @@
...
         <attribute name="domain" optional="YES" attributeType="String" syncable="YES"/>
         <fetchIndex name="domainIndex">
             <fetchIndexElement property="domain" type="Binary" order="ascending"/>
         </fetchIndex>
     </entity>
     <elements>
-        <element name="HTTPSUpgradeSimpleDomain" positionX="-63" positionY="-18" width="128" height="60"/>
-        <element name="HTTPSUpgradeWildcardDomain" positionX="-54" positionY="-9" width="128" height="60"/>
+        <element name="HTTPSBloomFilterSpecification" positionX="-1269" positionY="-171" width="128" height="90"/>
+        <element name="HTTPSWhitlistedDomain" positionX="-1098" positionY="-157" width="128" height="58"/>
```
brindy: Spelling mistake `HTTPSWhitlisted` -> `HTTPSWhitelisted`

Core/HTTPSTransientBloomFilterSpecification.swift
```diff
@@ -0,0 +1,37 @@
...
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+
+public struct HTTPSTransientBloomFilterSpecification: Equatable {
```
brindy: This is starting to seem overly complex with too many layers. 

Why not just dump Core Data for this and store the files on disk?  Using Codable pattern and optionals vars will help with migration issues.

If the whitelist is not likely to grow too big, drop Core Data for that too.

brindy (commented):
A few comments.  

In summary, I think CoreData is unnecessary here and adding complexity that isn't needed.

Otherwise, it's looking good and HTTPS functionality seems to work sweetly.

Core/HTTPSTransientBloomFilterSpecification.swift
subsymbolic: We discussed this and decided that it made sense to stick with CoreData over files on disk for structured data for the migration benefits.

Core/HTTPSUpgradeParser.swift
subsymbolic: Good suggestion. I had considered using Codable but was refactoring existing code and didn't know we had it in the project. I feel better using the modern pattern here, thanks!

brindy (approved):
Looks good! Nice work!

subsymbolic:
Thanks for the review @brindy!

PR #359, opened 2018-08-29
Fix in landscape layout for iPad devices in Settings->About

ArtCC:
Fix in landscape layout for iPad devices in Settings->About


ArtCC:
Hi!

Sorry for not use 'feature/'. My apology.

Regards.

PR #363, opened 2018-09-20
update privacy grade

brindy:
Task/Issue URL:  https://app.asana.com/0/392891325557410/709741655065319
Tech Design URL: https://app.asana.com/0/392891325557410/813385326952863
CC:

**Description**:

* Implement new Grade calculation algorithm
* Remove major networks data and replace with prevalence data
* Implement TOSDR scoring algorithm
* Use entity list for entity lookups (instead of Disconnect)
* Update the UI to support new B+ and C+ grades

**Steps to test this PR**:

1. Test upgrade from previous version as string cache is cleared
1. Thorough regression test around the privacy grade UI
1. Hacker News (news.ycombinator.com) is a good source of links that should generate the broad range of grades (A, B+, B, C+, C, D)

---
###### Internal references:


Core/ContentBlockerLoader.swift
```diff
@@ -52,39 +53,61 @@ public class ContentBlockerLoader {
         }
         easylistStore.removeLegacyLists()
     }
-
+    
     private func startRequests(with semaphore: DispatchSemaphore) -> Int {
-
         let contentBlockerRequest = ContentBlockerRequest()
-
+        requestEntityList(contentBlockerRequest, semaphore)
```
brindy, line 59: method was getting too big (lint) so refactored to smaller methods.

Core/ContentBlockerLoader.swift
```diff
@@ -52,39 +53,61 @@ public class ContentBlockerLoader {
...
+        requestDisconnectMe(contentBlockerRequest, semaphore)
+        requestTrackerWhitelist(contentBlockerRequest, semaphore)
+        requestHttpsUpgrade(contentBlockerRequest, semaphore)
+        requestHttpsWhitelist(contentBlockerRequest, semaphore)
+        requestSurrogates(contentBlockerRequest, semaphore)
+        return contentBlockerRequest.requestCount
+    }
+    
+    fileprivate func requestEntityList(_ contentBlockerRequest: ContentBlockerRequest, _ semaphore: DispatchSemaphore) {
```
brindy, line 68: new entity list downloader

Core/DisconnectMeStore.swift
```diff
@@ -30,26 +30,26 @@ public class DisconnectMeStore {
 
     public init() {
         stringCache = ContentBlockerStringCache()
+        loadTrackers()
```
brindy: Load the trackers in to memory on initialisation so that we can keep a reference to this.   The `networkNameAndCategory` function was moved in to this class as it made more sense and it uses the `trackers` property so that could get expensive when showing the trackers list in the UI.

Core/GradeCache.swift
```diff
@@ -0,0 +1,65 @@
...
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+
+public class GradeCache {
```
brindy, line 22: Moved in to its own file - it was previously in an extension which has now been deleted.

Core/TermsOfService.swift
```diff
@@ -20,125 +20,84 @@
...
 
 public struct TermsOfService: Decodable {
-
-    public init(from decoder: Decoder) throws {
-        let container = try decoder.container(keyedBy: CodingKeys.self)
-        self.classification = try? container.decode(Classification.self, forKey: .classification)
-        self.score = try container.decode(Int.self, forKey: .score)
-        let matchContainer = try container.nestedContainer(keyedBy: CodingKeys.Match.self, forKey: .match)
-        self.goodReasons = try matchContainer.decodeIfPresent([String].self, forKey: .good) ?? []
-        self.badReasons = try matchContainer.decodeIfPresent([String].self, forKey: .bad) ?? []
```
brindy: This seemed nasty to cope with the class being a string or false. I changed the source JSON to make false null instead and made the struct match the json format more directly.

DuckDuckGo/TabViewController.swift
```diff
@@ -32,6 +32,7 @@ class TabViewController: WebViewController {
 
     private lazy var appUrls: AppUrls = AppUrls()
     private lazy var appRatingPrompt: AppRatingPrompt = AppRatingPrompt()
+    private lazy var disconnectMeStore = DisconnectMeStore()
```
brindy, line 35: No longer used by the site rating, but is needed for showing the trackers UI

DuckDuckGoTests/MockFiles/tosdr.json
```diff
@@ -26,7 +26,7 @@
                      "you can request access and deletion of personal data"
                      ]
         },
-        "class": false
+        "class": null
```
brindy: makes the json parsing cleaner - and we're likely to dump this file soon anyway

Core/TermsOfService.swift
```diff
@@ -20,125 +20,84 @@
...
-            let container = try decoder.singleValueContainer()
-            let lowercaseString = try container.decode(String.self).lowercased()
-            guard let classification = Classification(rawValue: lowercaseString) else {
-                let context = DecodingError.Context(codingPath: [], debugDescription: "Classification string didn't match")
-                throw DecodingError.typeMismatch(Classification.self, context)
-            }
-            self = classification
+    
+    // see https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/e42533/shared/js/background/privacy-practices.es6.js#L20
+    var derivedScore: Int {
```
brindy, line 93: @andrey-p this is where the new score for the terms of service is calculated - should look pretty similar

Core/TermsOfService.swift
```diff
@@ -20,125 +20,84 @@
...
+        if hasGoodReasons && hasBadReasons {
             return .mixed
-        } else {
-            practices = Lookups.derivedScoreAsPractices[normalizeScore()]!
         }
-
-        if let practices = practices {
-            return practices
+        
+        if score < 0 {
```
brindy, line 81: @andrey-p we use an enum rather than returning the message directly so we can localise and do other stuff in the UI, but otherwise should look pretty familiar

DuckDuckGo/Grade.swift
```diff
@@ -0,0 +1,188 @@
...
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+
+public class Grade {
```
brindy, line 22: @andrey-p again this should look mostly familiar - but I've Swiftified some of the code

Core/PrivacyPractices.swift
```diff
@@ -0,0 +1,85 @@
...
+    private let siteScores: [String: Int]
+    private let termsOfServiceStore: TermsOfServiceStore
+    private let entityMapping: EntityMapping
+    
+    public init(termsOfServiceStore: TermsOfServiceStore = EmbeddedTermsOfServiceStore(), entityMapping: EntityMapping = EntityMapping()) {
+        let tld = TLD()
+        var entityScores = [String: Int]()
+        var siteScores = [String: Int]()
+        
+        termsOfServiceStore.terms.forEach {
```
brindy, line 53: @andrey-p I've made TOSDR the underlying implementation for the privacy practices.  This constructor floats the score up to the parent, but the score is actually derived as a function on the `TermsOfService` struct.

DuckDuckGoTests/GradeTests.swift
```diff
@@ -0,0 +1,158 @@
...
+        let grade = Grade()
+        XCTAssertEqual(.cPlus, grade.scores.site.grade)
+        XCTAssertEqual(10, grade.scores.site.httpsScore)
+        XCTAssertEqual(2, grade.scores.site.privacyScore)
+        XCTAssertEqual(12, grade.scores.site.score)
+        XCTAssertEqual(0, grade.scores.site.trackerScore)
+        
+    }
+    
+    func testPredefinedTestCases() throws {
```
brindy, line 108: @andrey-p this is where we run against the test cases in the submodule

DuckDuckGoTests/MockFiles/tosdr.json
subsymbolic: Why change it at all? We're going to dump it soon anyway but in the meantime this change makes any updates brittle as we can no longer dump the file i, we need to update it.

Core/ContentBlockerLoader.swift
subsymbolic: Nice change, much more readable.

Core/DisconnectMeStore.swift
```diff
@@ -30,21 +30,24 @@ public class DisconnectMeStore {
...
-            return [String: DisconnectMeTracker]()
+            self.trackers = [:]
         }
     }
+    
+    var hasData: Bool {
+        return !trackers.isEmpty
+    }
+
+    public var trackers: [String: DisconnectMeTracker] = [:]
```
subsymbolic, line 40: Public properties go at the top of the file and functions later. What do we want the order to be for the computed properties, they kind of sit between the two concepts. Perhaps they should go at the top after the standard properties?

Core/PrivacyPractices.swift
```diff
@@ -0,0 +1,85 @@
...
+            guard let url = URL(string: "http://\($0.key)") else { return }
```
subsymbolic: This seems convoluted. You have a key which perfectly matched the keys in entityMapping but the key gets turned into a url and the url gets stripped during the lookup. If you need the url option in findEntity, perhaps add an extra method?

subsymbolic (commented):
Great work, just a few comments

Core/TermsOfService.swift
subsymbolic: I don't think the json should be changed, see my comment on the DuckDuckGoTests/MockFiles/tosdr.json file

brindy: Reverted the tosdr files and reverted the parsing logic

DuckDuckGoTests/MockFiles/tosdr.json
brindy: Reverted the tosdr files and reverted the parsing logic

Core/DisconnectMeStore.swift
brindy: agreed

Core/PrivacyPractices.swift
brindy: Good shout.  Have removed the need to pass URL in a bunch of places where really just the host name is needed.

brindy:
Ready for re-review when you are @subsymbolic

subsymbolic (approved):

PR #365, opened 2018-09-28
Add https upgrades summary pixel

subsymbolic:
Task/Issue URL: https://app.asana.com/0/361428290920652/839627367764302
Tech Design URL: N/A however spec was discussed in https://app.asana.com/0/361428290920652/832602107339150

**Description**:
- Add a summary `ehs` pixel to determine https failures versus successes

**Steps to test this PR**:
1. Go to a broken https upgrade site ensure that the new ehd pixel is fired.
It may be easier to got to `HTTPSUpgrade.swift` and return true from the ` isInUpgradeList` method. Then visit https://badssl.com and tap on the "expired" link.
1. Navigate to a few sites successfully
1. Stop and relaunch the app ensure that the `ehs` summary pixel is fired on sync showing the count of failed versus total https sites in our upgrade list
1. Ensure that other app pixels continue to fire as normal
1. Ensure the app suncs as normal

---
###### Internal references:

### DIFF
--- a/Core/Pixel.swift
+++ b/Core/Pixel.swift
@@ -41,6 +41,7 @@ public enum PixelName: String {
     case longPressMenuShareItem = "mlp_s"
 
     case httpsUpgradeSiteError = "ehd"
+    case httpsUpgradeSiteSummary = "ehs"
 }
 
 public class Pixel {
@@ -48,8 +49,10 @@ public class Pixel {
     private static let appUrls = AppUrls()
     
     public struct Parameters {
-        static let url = "url"
-        static let errorCode = "error_code"
+        public static let url = "url"
+        public static let errorCode = "error_code"
+        public static let totalCount = "total"
+        public static let failureCount = "failures"
     }
     
     private struct Constants {
@@ -63,17 +66,17 @@ public class Pixel {
     public static func fire(pixel: PixelName,
                             forDeviceType deviceType: UIUserInterfaceIdiom = UIDevice.current.userInterfaceIdiom,
                             withAdditionalParameters params: [String: String?] = [:],
-                            withHeaders headers: HTTPHeaders = APIHeaders().defaultHeaders) {
+                            withHeaders headers: HTTPHeaders = APIHeaders().defaultHeaders,
+                            onComplete: @escaping (Error?) -> Void = {_ in }) {
         
         let formFactor = deviceType == .pad ? Constants.tablet : Constants.phone
         let url = appUrls
             .pixelUrl(forPixelNamed: pixel.rawValue, formFactor: formFactor)
             .addParams(params)
         
-        Alamofire.request(url, headers: headers).response { data in
-            Logger.log(items: "Fire pixel \(pixel.rawValue) \(data)")
+        Alamofire.request(url, headers: headers).validate(statusCode: 200..<300).response { response in
+            Logger.log(items: "Pixel fired \(pixel.rawValue) \(response)")
+            onComplete(response.error)
         }
-    
     }
-    
 }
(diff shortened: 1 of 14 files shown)


DuckDuckGo/AppConfigurationFetch.swift
```diff
@@ -0,0 +1,72 @@
+//
+//  AppBackgroundFetch.swift
```
brindy: wrong name

DuckDuckGo/AppConfigurationFetch.swift
```diff
@@ -0,0 +1,72 @@
...
+        DispatchQueue.global(qos: .background).async {
+
+            var newData = false
+            let semaphore = DispatchSemaphore(value: 0)
+            
+            self.sendHttpsSummaryPixel { newHttpsData in
+                newData = newData || newHttpsData
+                semaphore.signal()
+            }
+            semaphore.wait()
```
brindy: If these can be run in parallel then consider putting both `wait` calls at the end (either twice or use a for-loop). If you want them to run in serial then this is fine.

DuckDuckGo/AppConfigurationFetch.swift
```diff
@@ -0,0 +1,72 @@
...
+            
+            completion?(newData)
+        }
+    }
+    
+    private func sendHttpsSummaryPixel(completion: @escaping (Bool) -> Void) {
+
```
brindy, line 53: Day you need the boolean in the completion handler? The `newData` is to let iOS know if there was new data from the remote source or not so it can schedule the background fetch efficiently and I don't think this impacts that.

Later we should have a rethink around the `newData` thing for downloading the configuration files as we probably want iOS to schedule this background fetch as often as possible really. (Really we should use a server push!)

Core/Pixel.swift
```diff
@@ -41,15 +41,18 @@ public enum PixelName: String {
...
 public class Pixel {
 
     private static let appUrls = AppUrls()
     
     public struct Parameters {
-        static let url = "url"
-        static let errorCode = "error_code"
+        public static let url = "url"
+        public static let errorCode = "error_code"
+        public static let totalCount = "total"
```
brindy, line 57: can we make it clear these params are specific to `ehs` and retrospectively `ehd` ?

Core/WebViewController.swift
```diff
@@ -352,6 +357,7 @@ extension WebViewController: WKNavigationDelegate {
         let error = error as NSError
         if let url = webView.url, isHttpsUpgradeSite(url: url) {
             reportHttpsUpgradeSiteError(url: url, error: "\(error.domain)_\(error.code)")
+            statisticsStore.httpsUpgradesFailures += 1
```
brindy: This needs to happen in both places (or just inside `reportHttpsUpgradeSiteError`).

brindy (commented):
1 bug, 1 super minor change and a couple of queries

DuckDuckGo/AppConfigurationFetch.swift
subsymbolic: In this case we're sending data rather than receiving it but it still feels valid to tell apple that this sync "needed to happen" versus not. I understand that we want to kill this entirely one day, for now, I feel it's in keeping with the existing pattern in the app.

Core/WebViewController.swift
subsymbolic: Good find! Thanks

brindy (approved):
LGTM!

subsymbolic:
This was merged as a hotfix.

PR #368, opened 2018-10-15
Remove path from https reports

subsymbolic:
Task/Issue URL: https://app.asana.com/0/361428290920652/859498148695195

**Description**:
- Remove path from https reports
- Remove simpleUrl extension method based on previous team conversations that it is not needed

**Steps to test this PR**:
1. Update the `HTTPSUpgrade.isInUpgradeList` method to always return true.
1. Launch the app and watch traffic. Charles proxy can be a useful tool for this.
1. Visit https://expired.badssl.com/abcd
1. Note the EHD pixel fires with a url of https://expired.badssl.com (now omitting the "abcd" path section)

---
###### Internal references:

### DIFF
--- a/Core/WebViewController.swift
+++ b/Core/WebViewController.swift
@@ -460,9 +460,10 @@ extension WebViewController: WKNavigationDelegate {
     }
     
     private func reportHttpsUpgradeSiteError(url: URL, error: String) {
+        guard let host = url.host else { return }
         let params = [
             Pixel.EhdParameters.errorCode: error,
-            Pixel.EhdParameters.url: "https://\(url.host ?? "")"
+            Pixel.EhdParameters.url: "https://\(host)"
         ]
         Pixel.fire(pixel: .httpsUpgradeSiteError, withAdditionalParameters: params)
         statisticsStore.httpsUpgradesFailures += 1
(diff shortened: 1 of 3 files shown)


Core/WebViewController.swift
```diff
@@ -462,7 +462,7 @@ extension WebViewController: WKNavigationDelegate {
     private func reportHttpsUpgradeSiteError(url: URL, error: String) {
         let params = [
             Pixel.EhdParameters.errorCode: error,
-            Pixel.EhdParameters.url: url.simpleUrl
+            Pixel.EhdParameters.url: "https://\(url.host ?? "")"
```
nilnilnil: How I'm reading this, the default case is `https://` if `url.host` isn't defined?

nilnilnil (changes requested):

Core/WebViewController.swift
subsymbolic: In reality it will be defined (we have it as blank in the current logic too) but the host is optional so I need to provide a fallback. What would you prefer the default to be?

nilnilnil: I'd just prefer the whole operation not happen without the required info.

subsymbolic: @nilnilnil I added a guard, let me know if you think it needs to be tweaked more

nilnilnil (approved):

nilnilnil:
Looks good to me.

brindy (approved):
LGTM

bwaresiak (approved):
LGTM

nilnilnil:
Is this ready for merge?

subsymbolic:
Now merged! Thanks

PR #369, opened 2018-10-16
Improve scrolling behaviour

bwaresiak:
Task/Issue URL: https://app.asana.com/0/72649045549333/756199134358913/f
Tech Design URL: https://app.asana.com/0/481882893211075/862426684723572/f

**Description**:
- Change scrolling behaviour to match closely how it works in Safari.
- Fix performance issue around navigation bar hide/reveal animations.

**Steps to test this PR**:
1. Open any website, try scrolling.
2. Open DDG lite, bars should not hide on scroll.

This should be tested on both "classic" ;) and notch devices.

---
###### Internal references:

### DIFF
--- a/DuckDuckGo/BrowserChromeManager.swift
+++ b/DuckDuckGo/BrowserChromeManager.swift
@@ -30,25 +30,10 @@ protocol BrowserChromeDelegate: class {
 
 }
 
-private extension UIScrollView {
-    
-    /// Calculate Y-axis content offset corresponding to very bottom of the scroll area
-    var contentOffsetYAtBottom: CGFloat {
-        let yOffset = contentSize.height - bounds.height - contentInset.top + contentInset.bottom
-        if #available(iOS 11.0, *) {
-            return yOffset - safeAreaInsets.top + safeAreaInsets.bottom
-        } else {
-            return yOffset
-        }
-    }
-}
-
 class BrowserChromeManager: NSObject, UIScrollViewDelegate {
 
     struct Constants {
-
-        static let threshold: CGFloat = 60
-
+        static let threshold: CGFloat = 30
     }
 
     weak var delegate: BrowserChromeDelegate?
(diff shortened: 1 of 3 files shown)


DuckDuckGo/BrowserChromeManager.swift
```diff
@@ -33,68 +33,94 @@ protocol BrowserChromeDelegate: class {
...
+        let isInBottomBounceArea = scrollView.contentOffset.y > scrollView.contentOffsetYAtBottom
+        let startedFromVeryBottom = abs(draggingStartPosY - scrollView.contentOffsetYAtBottom) < 1
+        
+        if isInBottomBounceArea && startedFromVeryBottom {
+            updateBars(shouldHide: false)
+            // Fix for iPhone X issue: reaching bottom of the web page and then revealing
+            // bars by executing "bottom bounce" gesture, caused web view to re layout and
+            // cover bottom of the web page.
+            DispatchQueue.main.async { [weak scrollView] in
+                guard let scrollView = scrollView else { return }
```
brindy, line 74: just use `scrollView?.setContentOffset...` and you don't need this guard

brindy (commented):
Looks good, just one comment.

Also, can you test the app after commenting out this line:

https://github.com/duckduckgo/iOS/blob/develop/DuckDuckGo/MainViewController.swift#L383

This will force the in-app notification to show (it usually takes 3 days).  I think it's OK, but just see what you think.

DuckDuckGo/BrowserChromeManager.swift
bwaresiak: Initially I did so :) but then realised I'd have to add some silly default to `CGPoint(x: 0, y: scrollView.contentOffsetYAtBottom)` and went with guard as it seems a bit cleaner to me.

bwaresiak:
Good to know that, thanks for the tip @brindy :)

I think it looks good, I don't have any glitches around that in-app notification while testing on my phone.

DuckDuckGo/SiteRatingView.swift
```diff
@@ -51,7 +51,6 @@ public class SiteRatingView: UIView {
 
     public override func layoutSubviews() {
```
subsymbolic: This method is no longer needed, can we delete it?

DuckDuckGo/BrowserChromeManager.swift
brindy: You could force unwrap the scrollview at that point as it must be safe to use because the optional wouldn't let it get that far, but you're right a guard is probably cleaner than that.

bwaresiak: Yup, I got the same impression! Thanks :)

DuckDuckGo/SiteRatingView.swift
bwaresiak: Thanks for pointing that out, done 👍

brindy (approved):
LGTM - merge when you're ready

PR #373, opened 2018-10-22
fix tracker blocking bug

brindy:
Task/Issue URL: https://app.asana.com/0/392891325557410/872025706767190
Tech Design URL:
CC:

**Description**:

* remove protection id - compare current tab url with site rating url and only report trackers if they match
* inject scripts when web frame is created (otherwise they don't take effect quickly enough)
* fix bug where disconnect was injecting network name rather than parent url causing a javascript error which meant some trackers were missed
* add more useful javascript logging (will propagate to app, but not the javascript console and will not be logged in release builds)
* bump build number

**Steps to test this PR**:

1. Visit a range of URLs and ensure that trackers are being blocked as expected and that any JavaScript errors are not used by our scripts.
1. Ensure that the correct grade is applied to sites (for instance, navigating to and from DuckDuckGo.com which should always be an A)

---
###### Internal references:

### DIFF
--- a/Core/DisconnectMeStore.swift
+++ b/Core/DisconnectMeStore.swift
@@ -89,7 +89,7 @@ public class DisconnectMeStore {
     }
 
     private func convertToInjectableJson(_ trackers: [String: DisconnectMeTracker]) throws -> String {
-        let simplifiedTrackers = trackers.mapValues({ $0.networkName })
+        let simplifiedTrackers = trackers.mapValues { $0.parentUrl?.host }
         let json = try JSONSerialization.data(withJSONObject: simplifiedTrackers, options: .prettyPrinted)
         if let jsonString = String(data: json, encoding: .utf8) {
             return jsonString
--- a/Core/detection.js
+++ b/Core/detection.js
@@ -33,8 +33,12 @@
             type = event.target.nodeName
         }
 
+        duckduckgoMessaging.log("checking " + event.url + " (" + type + ")");
         duckduckgoContentBlocking.shouldBlock(event.url, type, function(url, block) {
-            if (!block) { return }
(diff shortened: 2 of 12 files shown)


Core/detection.js
```diff
@@ -33,8 +33,12 @@
             type = event.target.nodeName
         }
 
+        duckduckgoMessaging.log("checking " + event.url + " (" + type + ")");
         duckduckgoContentBlocking.shouldBlock(event.url, type, function(url, block) {
-            if (!block) { return }
+            if (!block) { 
+                duckduckgoMessaging.log("don't block " + url);
+                return 
+            }
```
bwaresiak, line 41: I admit I may be too careful here (as I don't know the exact cost) but wouldn't all these new Log calls add extra overhead? I'm all for logging these in debug (development) but I'm unsure about pushing too many events from JS to WebKit on production (especially, if they are going to be silenced anyway by Logger's `#if DEBUG`).

Core/DisconnectMeStore.swift
```diff
@@ -89,7 +89,7 @@ public class DisconnectMeStore {
     }
 
     private func convertToInjectableJson(_ trackers: [String: DisconnectMeTracker]) throws -> String {
-        let simplifiedTrackers = trackers.mapValues({ $0.networkName })
+        let simplifiedTrackers = trackers.mapValues { $0.parentUrl?.host }
```
bwaresiak, line 92: Awesome 👍

bwaresiak (approved):
I've got in one comment (optional), but overall this looks good :)

subsymbolic (approved):
Nothing additional from me. Looks good 👍

PR #374, opened 2018-10-23
Add bookmarks and quick action today extensions

subsymbolic:
Task/Issue URL: https://app.asana.com/0/361428290920652/866517273193956
Tech Design URL: N/A

**Description**:
Add a quick actions and bookmarks widget to the app

**Steps to test this PR**:

Before beginning setup network sniffing (e.g with Charles proxy) so you can watch pixels fire.

**TEST QUICK ACTIONS**
1. Run the app
1. Add the `quick actions` today extensions
1. Tap the `search bar` and note that the app launches into a new search. Ensure that the `mqe_s` pixel fires
1. Return to the extension.
1. Tap the `Bookmarks` button and note that the app launches into the bookmarks screen. Ensure the `mqe_b` pixel fires.
1. Return to the extension.
1. Tap the `Clear Data` button and note that the app launches into the fire screen. Ensure the `mqe_f` pixel fires.

**TEST BOOKMARKS EXTENSION**
1. Run the app and ensure there are no bookmarks
1. Add the `bookmarks` today extension
1. Note the `No bookmarks yet` text
1. Tap `New private search` and note that the app launches into a new search. Ensure that the `mbe_s` pixel fires
1. Add some bookmarks into the app
1. Return to the widget and note that these now show. Tap a bookmark and ensure the correct one opens and the `mbe_b` pixel fires.

---
###### Internal references:


BookmarksTodayExtension/TodayViewController.swift
```diff
@@ -22,7 +22,12 @@ import Core
 import NotificationCenter
 
 class TodayViewController: UIViewController, NCWidgetProviding, UITableViewDelegate, UITableViewDataSource {
-
+    
+    struct Index {
+        static let search = 0
+        static let startBookmarks = 1
+    }
```
bwaresiak, line 28: I'd try using something like 
```
    enum StaticRows: CaseIterable {
        static let search = 0
    }
```

and then just call `StaticRows.allCases.count` to get number of static entires.

BookmarksTodayExtension/TodayViewController.swift
```diff
@@ -44,12 +49,10 @@ class TodayViewController: UIViewController, NCWidgetProviding, UITableViewDeleg
...
-            extensionContext?.widgetLargestAvailableDisplayMode = mode
-        }
+        let mode = bookmarks.count > 2 ? NCWidgetDisplayMode.expanded : NCWidgetDisplayMode.compact
+        extensionContext?.widgetLargestAvailableDisplayMode = mode
 
-        if #available(iOSApplicationExtension 10.0, *), extensionContext?.widgetActiveDisplayMode == NCWidgetDisplayMode.compact {
+        if extensionContext?.widgetActiveDisplayMode == NCWidgetDisplayMode.compact {
             updatePreferredContentHeight(height: defaultHeight)
         } else {
             updatePreferredContentHeight(height: preferredHeight)
```
bwaresiak: Looking at `preferredHeight` method, it uses `let headerHeight = CGFloat(54.0)` const to adjust view height. Problem is, that with '54', expanded widget is a bit too big :)

I also wonder about defaultHeight method with 110 constant - we should probably be using [widgetMaximumSize(for:)](https://developer.apple.com/documentation/foundation/nsextensioncontext/1649135-widgetmaximumsize) there. One way to see problem around it is to change preferred font size in accessibility settings.

BookmarksTodayExtension/Info.plist
```diff
@@ -5,7 +5,7 @@
 	<key>CFBundleDevelopmentRegion</key>
 	<string>en</string>
 	<key>CFBundleDisplayName</key>
-	<string>DuckDuckGo</string>
+	<string>DuckDuckGo - Bookmarks</string>
```
bwaresiak: Maybe we could have a shorter name? With show more/show less button visible it doesn't look that good. :)

bwaresiak (changes requested):

BookmarksTodayExtension/Info.plist
subsymbolic: This came from the designs. I'll propose it to design and add it as a potential future tweak.

BookmarksTodayExtension/TodayViewController.swift
subsymbolic: I like that, much neater.

BookmarksTodayExtension/TodayViewController.swift
subsymbolic: Good point - those hardcoded values aren't nice (and some of them come from the initial hack day project). I've updated the code to remove them.

subsymbolic:
Thanks for the review and great feedback @bwaresiak. This is now ready for re-review.

bwaresiak (approved):
@subsymbolic This looks great, well done 👍

subsymbolic:
Thanks for the review @bwaresiak!

PR #376, opened 2018-11-02
Support for Light Theme

bwaresiak:
Task/Issue URL: https://app.asana.com/0/856498667320406/732857702748674
Tech Design URL: https://app.asana.com/0/856498667320406/872957960932374
CC:

**Description**:
Support for Light Theme in the application, like Android app does.

Notable features:
- Experimentation (see test steps).
- Pixels for entering settings and switching themes.
- New launch screen with fixes to logo alignment.
- New "page not found" image. 
- Dark keyboard appearance used when Dark Theme is set (in all cases, not only for experiments).

**Steps to test this PR**:

_Update scenario_
1. Install old version.
2. Update App.

Expectation: No theme toggle present in settings, Dark Theme is set.

_Fresh install: control group_
1. Set "mg" variant in code.
2. Run the app.

Expectation: No theme toggle present in settings, Dark Theme is set.

_Fresh install: experiment: light theme_
1. Set "mh" variant in code.
2. Run the app.

Expectation: Theme toggle present in settings, light theme by default.

_Fresh install: experiment: light theme_
1. Set "mj" variant in code.
2. Run the app.

Expectation: Theme toggle present in settings, dark theme by default.

_Toggle theme setting

1. Open Settings screen.
2. Change theme. 

Expectation: all screens should have appearance matching chosen theme.

_Persisting Theme setting

1. Open Settings screen.
2. Change theme. 
3. Restart the app.

Expectation: all screens should have appearance matching chosen theme.

_Testing about screen text_

1. Open About screen on iPhone.
2. Open About screen on iPad 

Expectation: text should be larger on iPad.

_Testing pixels_

For following actions:

- Open settings from actions menu.
- Open settings from tab switcher.
- Light mode enabled in settings.
- Light mode disabled in settings.

Pixels should be sent.

---
###### Internal references:


DuckDuckGo/EditBookmarkAlert.swift
```diff
@@ -31,8 +31,23 @@ class EditBookmarkAlert {
                            cancelCompletion: @escaping CancelCompletion) -> UIAlertController {
 
         let editBox = UIAlertController(title: title, message: "", preferredStyle: .alert)
-        editBox.addTextField { (textField) in textField.text = bookmark.title }
-        editBox.addTextField { (textField) in textField.text = bookmark.url.absoluteString }
+        
+        let keyboardAppearance: UIKeyboardAppearance
+        switch ThemeManager.shared.currentTheme.currentImageSet {
+        case .light:
+            keyboardAppearance = .light
```
brindy: Can we make the keyboard appearance a property of the theme as well? Then we won't have to repeat this code (seems to happen 4 times in total).

DuckDuckGo/ThemeManager.swift
```diff
@@ -0,0 +1,92 @@
...
+        }
+        
+        appSettings = settings
+        if settings.lightTheme {
+            currentTheme = LightTheme()
+        } else {
+            currentTheme = DarkTheme()
+        }
+        // Uncomment to enable automatic theme cycling for testing
+//        cycle(seconds: 4.0)
```
brindy: Delete

DuckDuckGo/ThemeManager.swift
```diff
@@ -0,0 +1,92 @@
...
+        
+        if lightThemeEnabled {
+            currentTheme = LightTheme()
+        } else {
+            currentTheme = DarkTheme()
+        }
+    }
+}
+
+// MARK: - Runtime Testing
```
brindy: Delete this extension - now that main dev is finished we can just test manually by changing in settings, like a user

DuckDuckGo/AppDependencyProvider.swift
```diff
@@ -0,0 +1,36 @@
...
+import Core
+
+protocol DependencyProvider {
+    var appSettings: AppSettings { get }
+    var variantManager: VariantManager { get }
+}
+
+/// Provides dependencies for objects that are not directly instantiated
+/// through `init` call (e.g. ViewControllers created from Storyboards).
+class AppDependencyProvider: DependencyProvider {
```
brindy, line 30: Optional: This is cool, but I wonder if it should be named more specifically to do with ViewControllers otherwise it feels as though it could be used throughout the entire app?  `ViewControllerDependencyProvider` perhaps?  But I'm not that bothered, just have a think about it.  

> Naming things is hard™️

:)

DuckDuckGo/AppUserDefaults.swift
```diff
@@ -46,4 +47,24 @@ public class AppUserDefaults: AppSettings {
         }
 
     }
+    
+    var lightTheme: Bool {
```
brindy: Optional: Thinking ahead slightly - I know we only have light and dark now, but perhaps a safer approach here would be to use the theme name.  If we do ever add more themes then we'll end up having to write migration code.  Then again YAGNI.  Also if we ever let the user define their own theme then this entire approach won't work anyway.  Your call! :)

brindy (commented):
Great job! Few comments for you to look at/ponder.

DuckDuckGo/ThemeManager.swift
```diff
@@ -0,0 +1,92 @@
...
+    enum ImageSet {
+        case light
+        case dark
+    }
+    
+    public static let shared = ThemeManager()
+
+    private var appSettings: AppSettings
+    
+    var rootController: UIViewController?
```
subsymbolic: I'm not comfortable keeping a strong reference to the rootViewController if we swap out the instance we'll leak a lot of memory. Does grabbing it as required from the UIAplication in the `curentTheme.didSet` work? e.g `UIApplication.shared.keyWindow?.rootViewController?`

You may need to wrap it in a provider for your unit tests.

DuckDuckGo/AppUserDefaults.swift
subsymbolic: I second this. Since we are expecting to add more themes and the complexity of the two approaches is similar (save string versus bool user defaults) I think we should choose the string approach.

DuckDuckGoTests/SettingsViewControllerTests.swift
```diff
@@ -0,0 +1,111 @@
...
+    override func setUp() {
+        mockDependencyProvider = MockDependencyProvider()
+        AppDependencyProvider.shared = mockDependencyProvider
+    }
+    
+    override func tearDown() {
+        AppDependencyProvider.shared = AppDependencyProvider()
+    }
+
+    func testLightThemeToggleInitialState() {
```
subsymbolic: We try to use a `when<condition>Then<expected outcome>` naming convention for tests (although there are still some old tests that don't do this)

subsymbolic (commented):
Good work @bwaresiak, the light theme looks great. I found a few issues, listed below.

**Tabs Crash**
When all tabs are deleted the app crashes. To reproduce, add a couple of tabs, go to the tab switcher and delete them all, the app will crash.
crash_when_all_tabs_deleted

**Autocomplete line**
There is a new line underneath autocomplete. This looks good on the light theme but too bright in dark mode.
autocomplete_line

**Tab icon on iOS 10**
When the theme is changed the tab icon is not displayed correctly
tab_icon_lost_on_ios_10

**Settings spacing on iOS 10**
The spacing in the SettingsViewController is incorrect, see `Application Lock` option compared to the option above it in attached image
settings_spacing_on_ios_10

DuckDuckGo/EditBookmarkAlert.swift
bwaresiak: That's a really good idea. :)

DuckDuckGo/AppDependencyProvider.swift
bwaresiak: Yes, naming things :)

`ViewControllersDependencyProvider` is good, tho:
`class AppViewControllersDependencyProvider: ViewControllersDependencyProvider` 
is a bit long :)

Alternative 1:
`class AppDependencyProvider: ViewControllersDependencyProvider`
... to keep it shorter, or...

Alternative 2:
`class AppMVCDependencyProvider: MVCDependencyProvider`
... to make it more related to MVC, so in theory it will be obsolete once we replace MVC with something else?

DuckDuckGo/AppUserDefaults.swift
bwaresiak: Good point :)

DuckDuckGo/AppDependencyProvider.swift
brindy: I'm starting to think this is a distraction, sorry. What is there works - we'll just have to watch out in reviews to make sure that if we use the `AppDependencyProvider` we're doing it for the right reason.

I look forward to our future architecture discussion!  :)

DuckDuckGo/ThemeManager.swift
bwaresiak: I've added a provider :)

bwaresiak:
@subsymbolic @brindy Thank you for the feedback, I've made tweaks to address your comments and re-tested app on iOS 10.  It should be ready for re-review :)

subsymbolic (approved):
Great work!

PR #382, opened 2018-11-17
Update log statement to display milliseconds

subsymbolic:
Task/Issue URL: https://app.asana.com/0/414709148257752/914681888297007
Tech Design URL: N/A, this is a tiny PR

**Description**:
A small tweak to fix an incorrect log statement that was displaying seconds rather than milliseconds. I was doing some investigation into bloom timing and noticed it so fixed it but this is not urgent, it can be reviewed whenever time allows.

**Steps to test this PR**:
1. Check that the code change makes sense
1. Visit example.com and make sure everything still runs fine and the "Site example.com can be upgraded. Lookup took ..ms" log statement appears in the console

---
###### Internal references:

### DIFF
--- a/Core/HTTPSUpgrade.swift
+++ b/Core/HTTPSUpgrade.swift
@@ -21,6 +21,10 @@ import Foundation
 
 public class HTTPSUpgrade {
     
+    private struct Constants {
+        static let millisecondsPerSecond = 1000.0
+    }
+    
     public static let shared = HTTPSUpgrade()
     
     private let dataReloadLock = NSLock()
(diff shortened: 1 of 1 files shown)


brindy (approved):
LGTM

One thing, if you wanted to avoid multiple date instances you could just use the start date and then use `timeIntervalSinceNow`. The result will be negative but that's easy to fix.  However, it's fine as is, up to you.

subsymbolic:
Thanks for the review @brindy, I've updated the log to use only one date instance

PR #390, opened 2019-01-04
Attempt to fix ETag crashes 

bwaresiak:
Task/Issue URL: https://app.asana.com/0/414709148257752/890456148190619
Tech Design URL:
CC:

**Description**:
ETag crashes (despite being really rare) are the most frequent ones. See asana task for details about investigation.

**Steps to test this PR**:
1. Install and open the app.
2. Check whether initial requests are processed properly and content blocking info is available.

---
###### Internal references:

### DIFF
--- a/DuckDuckGoTests/APIRequestTests.swift
+++ b/DuckDuckGoTests/APIRequestTests.swift
@@ -117,6 +117,30 @@ class APIRequestTests: XCTestCase {
         }
         waitForExpectations(timeout: 1.0, handler: nil)
     }
+    
+    func testWhenMultipleRequestsAreFiredThenCallbacksAreProcessedOnSerialQueue() {
+        
+        stub(condition: isHost(host)) { _ in
+            return fixture(filePath: self.validJson(), status: 200, headers: nil)
+        }
+        
+        let expectFirst = expectation(description: "first request has been processed")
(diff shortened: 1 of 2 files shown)


DuckDuckGoTests/APIRequestTests.swift
```diff
@@ -117,6 +117,30 @@ class APIRequestTests: XCTestCase {
...
+        
+        stub(condition: isHost(host)) { _ in
+            return fixture(filePath: self.validJson(), status: 200, headers: nil)
+        }
+        
+        let expectFirst = expectation(description: "first request has been processed")
+        let expectLast = expectation(description: "second request has been processed")
+        
+        APIRequest.request(url: url) { (_, error) in
+            // Give second request a chance to execute the callback
```
brindy, line 131: I think this comment would be more useful if it also said why. Would it be correct if we added "because it shouldn't since the requests are run serially." ?

brindy: Actually, the test name does imply that so maybe that's unnecessary. Your call.

bwaresiak: I'd keep it as it is now. I think - like you mentioned - that the test name clarifies this.

brindy (approved):
LGTM!

PR #395, opened 2019-01-31
Matches weak retain of controller present in other renderers

RobertPieta:
Task/Issue URL:
Tech Design URL:
CC:

**Description**:
The strong reference to the `HomeViewController` from a `HomeViewSectionRenderers` causes tabs to be retained much longer than expected, if not perpetually. Other similar instances already use a `weak` reference, and this PR matches the pattern used elsewhere in the codebase including the implicit unwrap.

**Steps to test this PR**:
1. Checkout the current branch without this PR
2. Build, Run, and Profile the app using the Allocations tool in Instruments
3. Make 5 tabs going to any website, wipe them with the flames 
4. End profile
5. Filter by objects with 'DuckDuckGo' in them, notice at least 5 retained tabs 
6. Checkout this PR, repeat steps 2-4
7. Now, only a single tab should be retained as is expected after using the flames

---
###### Internal references:

### DIFF
--- a/DuckDuckGo/HomeViewSectionRenderers.swift
+++ b/DuckDuckGo/HomeViewSectionRenderers.swift
@@ -144,7 +144,7 @@ class HomeViewSectionRenderers: NSObject, UICollectionViewDataSource, UICollecti
         
     }
     
-    private let controller: HomeViewController
+    private weak var controller: HomeViewController!
     private var theme: Theme
     private var renderers = [HomeViewSectionRenderer]()
     


RobertPieta:
All `AtbIntegrationTests` within the `AtbIntegrationTests` scheme pass locally on an iPhone 6s Plus running 11.0.1 and 12.1 using Xcode 10.1. Bitrise logs show tests ran using Xcode 10.0 on an iPhone 6s Plus using the 'latest' OS. 

One option may have been `weak` caused the renderer to release early and prevent the UI from appearing. However,  `HomeCollectionView` is retained by the view hierarchy and strongly references the `HomeViewSectionRenderers` so this case is unlikely (especially since observing local tests shows the UI as expected).

Will continue investigating, as I think some form of commit is needed to trigger a re-run of the tests.

bwaresiak:
This looks good, thank you for your contribution!

With regard to CI failure, please don't worry about it - that's on us. We are aware of the problem and already pushed a potential fix. Sorry for that!

PR #396, opened 2019-01-31
Flame asset optimization

RobertPieta:
Task/Issue URL:
Tech Design URL:
CC:

**Description**:
Memory usage for graphics on iOS is based on the UIImage size of the stored image. The flame image was 2000x878, with 20 images in the animation. On some devices, like iPad, the full image is needed to display a high resolution animation.

In the case of a portrait iPhone 7s, or even a portrait iPhone Xs, such a large resolution is not possible to display at once. This PR better leverages the asset catalogs features for device and width specific assets to reduce memory load of the flame animation.

Specifically, assets are split into iPhone Compact Width, iPhone Any Width, iPad Compact Width, and iPad Any Width variants with associated @2x and @3x resolutions.

The reduction in run-time memory usage is significant for portrait iOS devices, especially iPhone devices loading @2x compact width assets.

Finally, app bundle size should not be impacted (I was unable to archive to test this). On some devices, due to app thinning, the bundle size might even slightly decrease.

**Steps to test this PR**:
1. Checkout PR
2. Build and run the app
3. Try the flame animation in portrait
4. Try the flame animation in landscape
5. Repeat 3-4 for supported devices
6. Archive the app using this PR and verify the estimated bundle size has not changed significantly

---
###### Internal references:


bwaresiak:
Thank you for bringing that issue to light. 

While we agree that memory spike caused by the flame animation should be fixed, we are concerned about side effects of the approach proposed by you, mainly:
- Due to UIImage cache mechanism, for certain devices under certain conditions (clearing data in portrait mode, then in the landscape) memory spike can be even bigger due to two sets of images being loaded into memory (for both 'compact' and 'any' width variants).
- While this may seem counter-intuitive, thinning size report generated by the Xcode shows that size of the packages has increased: 10-15% for compressed, ~40% for uncompressed ones. Universal binary size jumped by 60% (due to Assets.car size growing from 12MB to 28MB).
- If possible, we would prefer to generate lower-resolution images ourselves, based on original assets made by our designers.

Additionally, there is a chance that the whole flame animation will be changed in the future. 

As a result, while we really appreciate the effort you put into creating this PR, we are going to pursue asset optimization in-house, as part of the regular app development cycle.

Thanks again.

PR #398, opened 2019-02-03
Auto clear data/tabs

bwaresiak:
Task/Issue URL: https://app.asana.com/0/856498667320406/765761023328863
Tech Design URL: https://app.asana.com/0/856498667320406/905625717624365

**Description**:
This PR adds option to set up automatic tabs/data clearing upon leaving the app. 

Notable features:
- Clearing tabs or tabs and data (e.g. cookies) when trigger conditions are met.
- Can be triggered by killing the app and (optionally) after certain amount of inactivity (when app is put to the background).
- Introduced new settings screen with auto clear options.
- Introduced app preview blocking (using blank screen) when clear tabs/data feature is enabled.

**Steps to test this PR**:
There are multiple scenarios to test:

**Clearing tabs**
1. Open some tabs.
2. Log in to a service (email, or whatever). 
3. Set up auto clear for 'tabs'.
4. Kill the app.
Expectation: upon restart, there should be no tabs open. When accessing mentioned in #2 service again, user should be logged in.

**Clearing data and tabs**
1. Open some tabs.
2. Log in to a service (email, or whatever). 
3. Set up auto clear for 'tabs and data'.
4. Kill the app.
Expectation: upon restart, there should be no tabs open. When accessing mentioned in #2 service again, user should **not** be logged in.

**Clearing action triggered by termination**
A:
1. Set up clearing with 'app exit' trigger.
2. Validate that killing the app clears the tabs.
B:
1. Set up clearing for 'app exit, inactivity X mins' .
2. Validate that killing the app clears the tabs.

**Clearing action triggered after delay**
1. Set up clearing with 'app exit, inactivity X mins' trigger.
2. Put app to the background.
Expectation: coming back to the app within chosen time period should not clear the tabs/data. Coming back once time has elapsed, should result in clear action being performed.

**Snapshot blocking**
1. Turn on tabs/data clearing.
2. Put app to the background.
Expectation: app preview in task manager should be replaced by "blank snapshot" of the app.
Note: This snapshot has two variants, depending on enabled features (home row&centered search vs other).

**Validating screen lock**
1. Turn on tabs/data clearing.
2. Turn on application lock.
3. Put app to the background.
Expectation: app screen locking (requiring touch ID/face ID) works as expected when coming back to the app or upon restart.

**Pixels**
A:
1. Open auto clear settings. 
Expectation: Pixel should be fired.
B:
1. Open auto clear settings. 
2. Tweak settings.
3. Exit auto clear settings screen.
Expectation: Pixels for tweaked options (action, trigger, or both) should be fired after #3.
C:
1. Open auto clear settings. 
2. Tweak settings, but then reset all to the initial values.
3. Exit auto clear settings screen.
Expectation: Pixels should not be fired after #3.

**Settings**
1. Open DDG settings screen.
Expectation: auto clear on/off status label should be properly displayed.

---
###### Internal references:


DuckDuckGo/AutoClearLogic.swift
```diff
@@ -0,0 +1,93 @@
...
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+import UIKit
+
+/// Represents worker that can peform clearing operation.
```
brindy: This comment is redundant imho.

DuckDuckGo/AutoClearLogic.swift
```diff
@@ -0,0 +1,93 @@
...
+protocol AutoClearWorker {
+    
+    func forgetData()
+    func forgetTabs()
+}
+
+class AutoClearLogic {
```
brindy: Can we think of a better name here?  `Logic` is fairly implicit by the virtue of being code so could maybe just drop that.

DuckDuckGo/AutoClearSettingsModel.swift
```diff
@@ -0,0 +1,59 @@
...
+    init?(settings: AppSettings) {
+        action = Action(rawValue: settings.autoClearAction)
+        guard action.isEmpty == false else { return nil }
+        
+        timing = Timing(rawValue: settings.autoClearTiming) ?? .termination
+    }
+    
+    /// Create settings model with default values.
+    init() {
+        action = [.clearTabs]
```
brindy: The current behaviour of the fire button is to clear both - I was a little surprised to find that the default action was just tabs. I think users turning this on will expect the default to match.

DuckDuckGo/AutoClearSettingsViewController.swift
```diff
@@ -0,0 +1,199 @@
...
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import UIKit
+import MessageUI
+import Core
+import Device
+
+// Back button
```
brindy: What is this comment for?

DuckDuckGo/AutoClearSettingsViewController.swift
```diff
@@ -0,0 +1,199 @@
...
+        case timing
+    }
+    
+    @IBOutlet weak var clearDataToggle: UISwitch!
+    @IBOutlet var labels: [UILabel]!
+    
+    private lazy var appSettings = AppDependencyProvider.shared.appSettings
+    private var clearDataSettings: AutoClearSettingsModel?
+    
+    static func loadFromStoryboard() -> UIViewController {
```
brindy: This is only for tests right? If it's not used for production can it be moved to a private extension in the tests where it's needed (or create a utility function the test can use in the test), please?

DuckDuckGo/MainViewController.swift
```diff
@@ -24,7 +24,7 @@ import Lottie
 
 // swiftlint:disable type_body_length
 // swiftlint:disable file_length
-class MainViewController: UIViewController {
+class MainViewController: UIViewController, AutoClearWorker {
```
brindy: Can you put this as an extension, please? Then it will match what we've done with the rest of the class.  We could do with tidying this class up at some point anyway because it's massive.

DuckDuckGoTests/AutoClearLogicTests.swift
```diff
@@ -0,0 +1,148 @@
...
+
+import XCTest
+@testable import DuckDuckGo
+@testable import Core
+
+class AutoClearLogicTests: XCTestCase {
+    
+    class MockWorker: AutoClearWorker {
+        
+        var forgetDataExpectation: XCTestExpectation?
```
brindy: I'm not a huge fan of this.  I understand the motivation, but I don't think it's particularly clear, especially if a test fails. 

For example, if you change this code to make the test fail and run it you get the test failed markers in Xcode but you don't get the inline reason for the error, you have to go hunting for it in the issue navigator.  Also it's adding a delay to the tests when they do fail.

Personally, I think asserts are clearer (for the reader) and the tests will fail faster. Can you change this to match how we've done mocks elsewhere, please?

This is another one to add to our future "architecture" discussion.  :)

DuckDuckGo/AppUserDefaults.swift
```diff
@@ -64,6 +67,30 @@ public class AppUserDefaults: AppSettings {
         
     }
     
+    var autoClearAction: Int {
```
brindy: Why not just have this return the enum or an optional? 

Likewise for the `autoClearTiming` value

brindy (commented):
I'm still testing it but it's looking good so far. A few comments for you which I'm happy to discuss.

DuckDuckGo/AutoClearSettingsViewController.swift
bwaresiak: Good catch, I forgot to remove it. :)

DuckDuckGo/AutoClearSettingsModel.swift
bwaresiak: As discussed, I made tabs&data the default and swapped the cells around.

DuckDuckGoTests/AutoClearLogicTests.swift
bwaresiak: Replaced with simple invocation counter. :)

bwaresiak:
@brindy Thanks for the review, I made some fixes based on your comments - could you have another look?

DuckDuckGoTests/AutoClearSettingsScreenTests.swift
```diff
@@ -0,0 +1,89 @@
...
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import XCTest
+@testable import Core
+@testable import DuckDuckGo
+
+extension AutoClearSettingsViewController {
```
brindy: minor nit: can you put this at the bottom and make it fileprivate, please?

brindy (approved):
1 minor nit but otherwise looks good, feel free to merge when you address that. :)

PR #402, opened 2019-02-07
emoji and international url support

brindy:
Task/Issue URL: 
* https://app.asana.com/0/414235014887631/873275390533187
* https://app.asana.com/0/392891325557410/1108230911455307
Tech Design URL:
CC:

**Description**:

Adds support emoji and international URLs while applying style to de-emphasise the everything beyond the domain name.

**Steps to test this PR**:

***General***
1. Ensure general functionality not broken

***Emoji domains***
1. Visit 💩.la domain - ensure the website is loaded
1. Check site rating and privacy dashboard
1. Search for 💩 - ensure a search is performed and the Emoji lookup kicks in
1. Visit https://emojipedia.org/pile-of-poo/ - scroll down and click on Poopla! link - it should load and show emoji in address bar
1. Add to bookmarks - make sure it appears and can be selected, edited and deleted
1. Add to whitelist - ensure it is on the list and that private protection is paused
1. Remove it from the whitelist - ensure it is removed and privacy protection is enabled

***International domains***
1. Visit 82.мвд.рф domain - ensure website loads and is auto-upgraded to https (maybe use a breakpoint)
1. Check site rating and privacy dashboard
1. Navigate the site - note that paths with international characters still show correctly
1. Search for мвд - search page should show related results and address bar should show мвд
1. Add to bookmarks - make sure it appears and can be selected, edited and deleted
1. Add to whitelist - ensure it is on the list and that private protection is paused
1. Remove it from the whitelist - ensure it is removed and privacy protection is enabled

***Deemphasis of path/query***

1. While browsing ensure the path and query string of the address is deemphasised 
1. Check both themes
1. Ensure theme is applied when changed 

---
###### Internal references:


AtbIntegrationTests/AtbIntegrationTests.swift
```diff
@@ -18,7 +18,6 @@
 //
 
 import XCTest
-import Swifter
```
brindy: This dependency seemed to be unused so I removed it.

Cartfile
```diff
@@ -19,3 +19,5 @@ github "duckduckgo/swifter" == 1.4.3
 
 github "duckduckgo/lottie-ios" "tags/2.5.0"
 
+github "duckduckgo/PunycodeSwift" "brindy/app_extension_api_flag"
```
brindy: This prevents a warning about linking to libraries that don't have the "Allow safe API extension" flag set. I've raised this with the project owner (and will probably create PR for it myself).  We can update this to a tag once that's done.

Core/SiteRating.swift
```diff
@@ -26,7 +26,7 @@ public class SiteRating {
     }
 
     public var domain: String? {
-        return url.host
+        return url.host?.punycodeDecodedHostname
```
brindy: This var is used mainly for display to the user so should show the punycode decoded version.

Core/SiteRating.swift
```diff
@@ -158,7 +158,7 @@ public class SiteRating {
     }
     
     public func isFor(_ url: URL?) -> Bool {
-        return domain == url?.host
+        return self.url.host == url?.host
```
brindy, line 161: This probably shouldn't have been using the domain var in the first place. This ensures both hosts are compared on equal terms.

Core/URLExtension.swift
```diff
@@ -141,7 +141,7 @@ extension URL {
         }
 
         // from https://stackoverflow.com/a/25717506/73479
-        let hostNameRegex = "^(((?!-)[A-Za-z0-9-]{1,63}(?<!-)\\.)+[A-Za-z]{2,63})$"
+        let hostNameRegex = "^(((?!-)[A-Za-z0-9-]{1,63}(?<!-)\\.)+[A-Za-z0-9-]{2,63})$"
```
brindy, line 144: The change allows hyphens in the domain name which is what happens when characters are punycoded. In actual fact this regex is slightly wrong because it doesn't allow local host names but that suits has as we treat them like a search.

DuckDuckGo/OmniBar.swift
```diff
@@ -169,7 +169,13 @@ class OmniBar: UIView {
             return
         }
         resignFirstResponder()
-        omniDelegate?.onOmniQuerySubmitted(query)
+        
+        if let url = query.punycodedUrl {
```
brindy, line 194: If the query can be coverted to a punycoded url it most likely means the user entered something like 💩.la or some other international domain name.  In that case we should use the punycoded version for processing, otherwise just use whatever text was entered and this will get treated like a query.

brindy:
@bwaresiak as discussed, I've removed the puny-decoding

bwaresiak (approved):
Great, LGTM @brindy :)

PR #405, opened 2019-02-12
Fix video auth bypass

subsymbolic:
Task/Issue URL: https://app.asana.com/0/361428290920652/944767190704272

**Description**:
Fix issue where video can show aboce authentications screen

**Steps to test this PR**:
1. Launch app
1. Go to settings and turn on "Application Lock"
1. Visit Vimeo https://vimeo.com/watch
1. Play a viceo and make it full screen
1. Background the app for 5 seconds
1. Launch the app again

Previous behavior: Video was visible above lock screen but under dialog
New behavior: Video is now completely hidden behind lock screen and dialog

Note: the video will continue to play when the app is backgrounded. Whether this is desirable behavior or not is another question and goes beyond the scope of this task.

---
###### Internal references:

### DIFF
--- a/DuckDuckGo/AppDelegate.swift
+++ b/DuckDuckGo/AppDelegate.swift
@@ -33,6 +33,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     var window: UIWindow?
 
     private lazy var bookmarkStore: BookmarkStore = BookmarkUserDefaults()
+    private lazy var privacyStore = PrivacyUserDefaults()
     private var autoClear: AutoClear?
 
     // MARK: lifecycle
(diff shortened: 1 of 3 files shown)


DuckDuckGo/AppDelegate.swift
```diff
@@ -57,22 +58,33 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+        autoClear?.applicationDidLaunch()
+        AppConfigurationFetch().start(completion: nil)
+        initialiseBackgroundFetch(application)
     }
 
     func applicationWillEnterForeground(_ application: UIApplication) {
-        beginAuthentication()
+        if privacyStore.authenticationEnabled {
+            beginAuthentication()
+        }
```
bwaresiak, line 90: `beginAuthentication()` has a side-effect: it always calls `removeOverlay()`. :)

As a result, this scenario fails:
1. Enable auto clear feature. Make sure App Lock is disabled.
2. Background the app.
3. Move back to the app - blank screen overlay is not removed.

Just calling `removeOverlay()` in `else` should fix that (unless you have some other small scale refactoring in mind that makes more sense to you :) ).

DuckDuckGo/AppDelegate.swift
```diff
@@ -57,22 +58,33 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
         if appIsLaunching {
             appIsLaunching = false
-            autoClear = AutoClear(worker: mainViewController!)
-            autoClear?.applicationDidLaunch()
-            AppConfigurationFetch().start(completion: nil)
+            onApplicationLaunch(application)
+        }
+    }
+    
+    private func onApplicationLaunch(_ application: UIApplication) {
```
bwaresiak, line 72: Great, I'm glad you extracted that. 😄

bwaresiak (commented):
Nice! Full screen videos seem to work as expected now. 👍 

I've found one small issue tho (in the comment), but seems pretty simple to fix. :-)

> Note: the video will continue to play when the app is backgrounded. Whether this is desirable behavior or not is another question and goes beyond the scope of this task.

I wonder, do we have task for that in the backlog?

DuckDuckGo/AppDelegate.swift
subsymbolic: Oh nice catch, that's fixed.

subsymbolic:
> I wonder, do we have task for that in the backlog

Not that I'm aware. I prefer the current behavior as I like to continue listening to music and spanish learning videos when my screen locks. That said I can imagine users hating it too, hence my comment. Shall we wait and see if there's feedback?

bwaresiak:
Makes sense! 👍

DuckDuckGo/AppDelegate.swift
bwaresiak: @subsymbolic I'm afraid you've added that `else` in the wrong place: it should be here (in `applicationWillEnterForeground`) and it is not needed in `onApplicationLaunch` as there is no overlay on app startup.

subsymbolic: Oh dear, I should try not to code when I'm trying to do maths over chat. Thanks @bwaresiak, will fix.

bwaresiak: Now it works as expected. 👍

bwaresiak (approved):

subsymbolic:
Many thanks for the review @bwaresiak!

PR #408, opened 2019-02-15
simplify approach to deemphasising the url hopefully to reduce crashes

brindy:
Task/Issue URL: https://app.asana.com/0/392891325557410/1109882026372264
Tech Design URL:
CC:

**Description**:

Some users get a crash after updating because (we think) they have a URL that was crashing out our existing deempshasis URL logic.  This PR hardens it and simplifies the logic at the same time.

**Steps to test this PR**:
1. Enter some searches and URLs
1. Browse the web

---
###### Internal references:

### DIFF
--- a/Core/StringExtension.swift
+++ b/Core/StringExtension.swift
@@ -76,10 +76,14 @@ extension String {
         }
         
         let urlAndQuery = s.split(separator: "?")
+        guard urlAndQuery.count > 0 else {
+            return nil
+        }
+        
         let query = urlAndQuery.count > 1 ? "?" + urlAndQuery[1] : ""
         let componentsWithoutQuery = [String](urlAndQuery[0].split(separator: "/").map { String($0) }.dropFirst())
         guard componentsWithoutQuery.count > 0 else {
-            return URL(string: self)
+            return nil
         }
         
         let host = componentsWithoutQuery[0].punycodeEncodedHostname
--- a/Core/StringExtension.swift
+++ b/Core/StringExtension.swift
@@ -53,6 +53,10 @@ extension String {
     /// URL and URLComponents can't cope with emojis and international characters so this routine does some manual processing while trying to
     ///  retain the input as much as possible.
     public var punycodedUrl: URL? {
+        if let url = URL(string: self) {
+            return url
+        }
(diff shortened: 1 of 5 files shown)


Core/StringExtension.swift
```diff
@@ -53,6 +53,10 @@ extension String {
     /// URL and URLComponents can't cope with emojis and international characters so this routine does some manual processing while trying to
     ///  retain the input as much as possible.
     public var punycodedUrl: URL? {
+        if let url = URL(string: self) {
```
brindy, line 56: If this string can be a URL, it might as well be - nothing needs to be done

DuckDuckGo/OmniBar.swift
```diff
@@ -156,26 +156,32 @@ class OmniBar: UIView {
...
-        let s = url.absoluteString
-        let parts = s.split(separator: "/")
-        let offset = (parts[0] + "//" + parts[1]).count
+    public class func demphasisePath(forUrl url: URL) -> NSAttributedString? {
         
+        let s = url.absoluteString
         let attributedString = NSMutableAttributedString(string: s)
-        let theme = ThemeManager.shared.currentTheme
-        
-        let idx = s.unicodeScalars.index(s.unicodeScalars.startIndex, offsetBy: offset)
```
brindy, line 171: This is the line causing the crash.  When we removed puny decoding this should have been changed then anyway as it's redundant if the string is always ascii. However, the URL components approach is better anyway.

DuckDuckGoTests/OmniBarTests.swift
```diff
@@ -0,0 +1,64 @@
...
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+
+import XCTest
+@testable import DuckDuckGo
+
+class OmniBarTests: XCTestCase {
```
brindy, line 25: The `demphasisePath` function doesn't return an optional so we just exercise a bunch of URLs to check for crashes.

bwaresiak (approved):
Looks good to me 👍

brindy:
Thanks @bwaresiak !

Issue #416, opened 2019-02-22
bookmark title is clipping few letters


pastmaster007:
**Issue:**
bookmark title is clipping few letters refer below image

https://i.imgur.com/uDZuZeN.png

bwaresiak:
Closing this, as explained in https://github.com/duckduckgo/iOS/pull/417

> This problem is caused by system font - internally we use Proxima Nova fonts that are proprietary, and with these, problem does not exist (as can be seen in the APpStore version of the app).

PR #417, opened 2019-02-22
fix for bookmark title clipping

pastmaster007:
Task/Issue URL:
#416 
**Description**:
fixed bookmark title clipping by increasing image view height constraints inside stackview from 16 to 24..

**Steps to test this PR**:
Issue image:
https://i.imgur.com/uDZuZeN.png

Fixed image:
https://i.imgur.com/uIxK4X7.png

---
###### Internal references:


bwaresiak:
Thank you for finding time to do this and sorry it took me a while to get back to you.

The problem you noticed is caused by system font - internally we use Proxima Nova fonts that are proprietary, and with these, problem does not exist (as can be seen in the APpStore version of the app). I consulted with design team about your change to the image view size, but the conclusion is that we don't want to tweak it that way right now. 

Since this problem does not affect AppStore version (as it is build internally, with Proxima Nova fonts) we are not going to proceed with this fix.

Once again, thank you for bringing that up.

PR #418, opened 2019-02-25
Improve app statistics

subsymbolic:
Task/Issue URL: https://app.asana.com/0/414235014887631/1109730748569886
Tech Design URL: https://app.asana.com/0/481882893211075/1109730748569882

**Description:**
Update the app to maintain two versions of the latest atb (this is the set_atb parameter sent in atb.js) to provide better statistics. One for search and one for app usage.

**Steps to test this PR**

**Test 1: Migration:**
1. Install and run a develop version of the app with an old atb and yesterday's retentionAtb (these values can be hardcoded in StatisticsUserDefaults)
1. Open Charles to watch requests
1. Now install this branch (revert any hardcoding code from step 1) over the top
1. Note that when the app starts an atb.js request is made with an additional parameter of `at` and the original atb for both `atb` and `set_atb`
https://duckduckgo.com/atb.js?test=1&at=app_use&atb=v157-0mn&set_atb=v157-0
1. Background the app and open again note that the atb request is fired again only this time set_atb contains the atb from the previous response e.g if the latest atb is v157-6
https://duckduckgo.com/atb.js?test=1&at=app_use&atb=v157-0mn&set_atb=v157-6
1. Perform a search, this will not contain the `at` parameter and will use the previously known searchRetentionAtb, which will be yesterday's value e.g v157-5
https://duckduckgo.com/atb.js?test=1&atb=v157-0mn&set_atb=v157-5
1. Perform another search and note that the search atb is up-to-date now too
https://duckduckgo.com/atb.js?test=1&atb=v157-0mn&set_atb=v157-6
1. Note that no exti or initial atb request (the ones with no `atb` or `set_atb`  params) are sent after step 1
`https://duckduckgo.com/exti/?test=1&atb=v157-6mn`
`https://duckduckgo.com/atb.js?test=1`

**Test 2: New installation:**
1. Install a fresh version of the app and watch requests in Charles
1. Note that the initial atb request (with no `atb` or `set_atb`  params) appears only once
`https://duckduckgo.com/atb.js?test=1`
1. Note that the `exti` request only appears once
`https://duckduckgo.com/exti/?test=1&atb=v157-6mn`

---
###### Internal references:


subsymbolic:
I'm leaving this PR as a draft until backend work is done as things may change. Note also that atb tests are often failing on CI. This may have been an issue before this PR, either way, l will investigate.

AtbIntegrationTests/AtbIntegrationTests.swift
```diff
@@ -24,25 +24,26 @@ class AtbIntegrationTests: XCTestCase {
 
     struct Constants {
         static let initialAtb = "v100-1"
-        static let retentionAtb = "v102-7"
+        static let searchRetentionAtb = "v102-7"
+        static let appRetentionAtb = "v102-6"
         static let devmode = "test"
         static let atbParam = "atb"
         static let setAtbParam = "set_atb"
+        static let activityType = "at"
```
subsymbolic: This param is likely to change

DuckDuckGoTests/VariantManagerTests.swift
```diff
@@ -32,7 +32,6 @@ class VariantManagerTests: XCTestCase {
 
         let mockStore = MockStatisticsStore()
         mockStore.atb = "atb"
-        mockStore.retentionAtb = "ratb"
```
subsymbolic, line 67: This was unused

subsymbolic:
The last 4 builds of atb tests on CI  have been stable (as have been all local runs), I'll continue to monitor them.

subsymbolic:
@brindy the backend is done and this is now ready for review

AtbIntegrationTests/AtbIntegrationTests.swift
```diff
@@ -202,12 +205,21 @@ class AtbIntegrationTests: XCTestCase {
         }
         
         server["/atb.js"] = {
-            self.atbRequests.append($0)
+            var atb = ""
```
brindy: This logic makes me nervous.  Rather than putting in to an appropriate bucket by type why not assert requests are the type that are expected in the order that is expected?

AtbIntegrationTests/AtbIntegrationTests.swift
```diff
@@ -202,12 +205,21 @@ class AtbIntegrationTests: XCTestCase {
...
+            if $0.queryParam(Constants.activityType)  != nil {
+                atb = Constants.appRetentionAtb
+                self.appRetentionAtbRequests.append($0)
+            } else if $0.queryParam(Constants.atbParam)  != nil {
+                atb = Constants.searchRetentionAtb
+                self.searchRetentionAtbRequests.append($0)
+            } else {
+                atb = Constants.initialAtb
+                self.installAtbRequests.append($0)
```
brindy: Aren't install requests sen to `/exti` ?

DuckDuckGo/AppDelegate.swift
```diff
@@ -66,8 +66,10 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     func applicationDidBecomeActive(_ application: UIApplication) {
         guard !testing else { return }
         
-        Pixel.fire(pixel: .appLaunch)
-        StatisticsLoader.shared.load()
+        StatisticsLoader.shared.load {
+            StatisticsLoader.shared.refreshAppRetentionAtb()
+            Pixel.fire(pixel: .appLaunch)
```
brindy, line 71: There were questions asked about whether this was the right place to fire this pixel (ie is our app launch logic correct?).  Does that apply here? No one really cares about the pixels which is why we never prioritised fixing it,  but it could be more of an issue for the stats which we'll end up depending on more.

brindy (commented):
Looks good. Some comments which might be quicker to discuss over zoom if you want.

DuckDuckGo/AppDelegate.swift
subsymbolic: This is just a small fix I put in place as I noticed that there is a race condition where the pixel may fire before the atb request making a blank atb more likely, the pixel logic is otherwise unchanged by this work. There is however another task to look at the launch pixel timing across both apps.

AtbIntegrationTests/AtbIntegrationTests.swift
subsymbolic: Install requests involve both a call to get the initial atb (with no params) followed by a call to exti.

DuckDuckGo/AppDelegate.swift
brindy: Fair enough, but doesn't the refreshAppRetentionAtb happen async and could that mean the pixel still has a blank atb?

Did you mean to put it in the completion handler for the refresh call?

AtbIntegrationTests/AtbIntegrationTests.swift
subsymbolic: @brindy that's a good point, testing the overall request count and order would be preferable. I've refactored the code to do that.

DuckDuckGo/AppDelegate.swift
subsymbolic: The pixel only needs the initial atb to be loaded, it doesn't care about the retention atb so the code is as intended.

brindy (approved):
Solid improvement! 👍

subsymbolic:
Thanks for the review @brindy 😄

PR #421, opened 2019-03-14
release enhanced homescreen

brindy:
Task/Issue URL: https://app.asana.com/0/392891325557410/1111571693570069
Tech Design URL:
CC:

**Description**:

* Makes the enhanced home page available to users via a setting.
* Makes the cancel button version of the omni bar the default
* Cleans up variants
* Adds pixels to measure engagement 
* Moves refresh from menu to address bar
* Other minor UI updates as per design reviews

**Steps to test this PR**:

Clean Install:
1. Test each mode of the home page
1. After adding favorites return to a previous mode of the home page - the favorites should be migrated to bookmarks
1. The bookmarks view will show a favorites section when in favorites mode - the user can move bookmarks in  and between each section

Upgrade:
1. Install previous version with variant selected (may need to change code to make weightings in favour of favorites variant)
1. Upgrade to new version - should still show favorites
1. Repeat for centre search only

Ideally check on following device types:
1. iPad
1. iPhone
1. Small phone

---
###### Internal references:


DuckDuckGo/OmniBar.swift
```diff
@@ -228,14 +243,18 @@ class OmniBar: UIView {
...
+        omniDelegate?.onRefreshPressed()
+    }
 }
 
 extension OmniBar: UITextFieldDelegate {
 
     func textFieldDidBeginEditing(_ textField: UITextField) {
-        omniDelegate?.onTextFieldDidBeginEditing(self)
-        refreshState(state.onEditingStartedState)
         DispatchQueue.main.async {
```
brindy, line 255: Moved these lines in here to prevent strange animation caused by the keyboard

DuckDuckGo/MainViewController.swift
```diff
@@ -392,8 +399,8 @@ class MainViewController: UIViewController {
 
     fileprivate func displayAutocompleteSuggestions(forQuery query: String) {
         if autocompleteController == nil && appSettings.autocomplete {
-            allowContentUnderflow = false
             let controller = AutocompleteViewController.loadFromStoryboard()
+            controller.shouldOffsetY = allowContentUnderflow
```
brindy, line 408: Originally I had it set `allowContentUnderflow = false` when the autocomplete was shown, but there was an easy to replicate bug:

1. With centered search on the home page, enter a search term that shows autocomplete results
1. Tap the screen to exit autocomplete / close the keyboard
1. The fake search field would then be visible

I had a hack which set `allowContentUnderflow = true` but the keyboard animation made it look really obvious and I couldn't work out how to hide it without it being see by the user.

This approach tells the autocomplete view controller to use a vertical offset or not depending on the current state.

Happy to hear other suggestions for this!

DuckDuckGo/SettingsViewController.swift
```diff
@@ -57,11 +61,27 @@ class SettingsViewController: UITableViewController {
         super.viewWillAppear(animated)
         
         configureAutoClearCellAccessory()
+        configureHomePageCellAccessory()
+        migrateFavoritesIfNeeded()
+    }
+    
+    private func migrateFavoritesIfNeeded() {
```
brindy, line 68: This ensures the favorites are not lost if the user changes the setting.  However, there's no going back to favourites, so the user will have to set them up again (using the edit functionality in bookmarks).

adhoc-build.sh
```diff
@@ -0,0 +1,3 @@
+
```
brindy, line 1: Handy script for adhoc builds:

> ./adhoc-build ~/Desktop

Will deposit a timestamped IPA on your desktop so long as you have the private key installed and your Xcode setup for managing profiles, etc.

DuckDuckGo/UserText.swift
```diff
@@ -181,4 +181,11 @@ public struct UserText {
     public static let bookmarkAllTabsSaved = NSLocalizedString("bookmarkAll.tabs.saved", comment: "All open tabs are bookmarked")
     public static let bookmarkAllTabsFailedToSave = NSLocalizedString("bookmarkAll.tabs.failed", comment: "Failed to bookmark %lu out of %lu tabs")
 
+    public static let autoClearAccessoryOn = NSLocalizedString("autoclear.on", comment: "On")
+    public static let autoClearAccessoryOff = NSLocalizedString("autoclear.off", comment: "Off")
+
+    public static let homePageSimple = NSLocalizedString("homepage.simple", comment: "Simple")
+    public static let homePageCenterSearch = NSLocalizedString("homepage.centerSearch", comment: "Center Search")
+    public static let homePageCenterSearchAndFavorites = NSLocalizedString("homepage.centerSearchAndFavorites", comment: "Center Search and Favorites")
```
bwaresiak, line 192: I noticed there is a slight inconsistency with copy that is on main Settings screen and the one in Home Screen mode picker - is that intended?

DuckDuckGo/HomePageSettingsViewController.swift
```diff
@@ -0,0 +1,88 @@
...
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import UIKit
+import Core
+
+protocol HomePageSettingsDelegate: NSObjectProtocol {
+    
+    func homePageChanged(toConfigName config: HomePageConfiguration.ConfigName)
```
bwaresiak: Probably we can make it just homePageChanged(to:) since 'ConfigName' is part of the passed type, what do you think?

DuckDuckGo/AppSettings.swift
```diff
@@ -23,7 +23,7 @@ protocol AppSettings {
     
     var autoClearAction: AutoClearSettingsModel.Action { get set }
     var autoClearTiming: AutoClearSettingsModel.Timing { get set }
+
+    var homePage: HomePageConfiguration.ConfigName { get set }
     
-    // Required for Light Theme experiment
-    func setInitialThemeNameIfNeeded(name: ThemeName)
```
bwaresiak, line 28: Oh, right 🤦‍♂️  👍

DuckDuckGo/TabSwitcherViewController.swift
```diff
@@ -63,6 +64,15 @@ class TabSwitcherViewController: UIViewController {
         scrollToInitialTab()
     }
 
+    override func prepare(for segue: UIStoryboardSegue, sender: Any?) {
+        
+        if let nav = segue.destination as? UINavigationController,
+            let controller = nav.topViewController as? SettingsViewController {
+            controller.homePageSettingsDelegate = self
```
bwaresiak: What are your thoughts about setting it directly to 'homePageSettingsDelegate' instead of conforming to HomePageSettingsDelegate here?

bwaresiak (commented):
Awesome to see that being released @brindy I'll definitely use the favorites :)

In general it works really well, great job. I've found just few edge cases:

**AutoClear blank snapshot's omni bar configuration**
When auto clear is turned on, we make a snapshot for iOS task manager. In   'BlankSnapshotViewController' 'configureOmniBar(variantManager:)' method should take into account centered search/favorites setting and hide top bar.

**In call bar related issues**
Rotation issue:
1. Toggle in call bar.
2. Navigate to home screen.
3. Rotate.

Result: With 'simple' home screen UI glitches. With centered searc/favourites it crashes.

Centered search misalignment:
1. Toggle in call bar.
2. Press fire button.
3. Turn off in call bar.
4. Tap search field -> top bar glitches.

**Querying Favorites icons**
1. Add favourite for which we can not obtain favicon (e.g. 'www.wp.pl').
2. Observe requests being sent on rotation or during scrolling (if cell is reused).
Shouldn't we have some caching mechanism for that?

**Centered search - iPad glitches on rotation**
1. Enter home screen in centered search/favorites mode in portrait orientation.
2. Change to Landscape orientation.
3. Change to Face up orientation -> observer glitch.
4. Change to Portrait orientation.
3. Change to Face up orientation -> observer glitch.

bwaresiak:
One more question: shouldn't we clear homeScreen/centeredSearch from FeatureName in VariantManager.swift?

DuckDuckGo/MainViewController.swift
bwaresiak: No reasonable alternative has came to my mind so far. :)

bwaresiak:
There is one more thing I noticed after using it over the weekend on the iPhone SE: in centered search mode when I'm on the home screen, it is often very difficult to navigate to settings icon (some taps in proximity of settings button seem to be ignored).

This problem does not exist for simple home screen layout or when omni bar is present.

brindy:
Thanks @bwaresiak ! I'll get on these and know when it needs reviewed again.

brindy:
Apart from the 404 caching this should be ready to re-review.  We're using that Kingfisher library there so might have to do something a bit low level to get caching working.

brindy:
OK, am caching image not found / loading errors in memory for an hour using a simple dictionary which should reduce the spam on the favicon service and I've integrated in other parts of the app as well.

bwaresiak (approved):
Great!

As discussed, I think this is good to go, I've noticed only few small things:

1. Onboarding contextual tips/ home row CTA can be triggered by changing Home Screen mode.
2. iPad with centered search, when in landscape mode on home screen, start searching so the autocomplete is being shown, then rotate -> glitch (autocomplete is still being shown).
3. iPad with centered search (but not with favorites), when in landscape mode on home screen, enter omni bar, then rotate -> small glitch (logo is moved a bit towards the top).

PR #427, opened 2019-03-28
onboarding experiment 2

brindy:
Task/Issue URL: https://app.asana.com/0/inbox/392891325557408/1112862352963248/1116310383481169
Tech Design URL:
CC:

**Description**:

This experiment is to see the effect of an extra information button thus the control group is also the summary screen and we have released the contextual tips from experiment 1.

Check on following devices:

* SE
* iPhone 8
* iPhone X 
* iPad

You'll also need to update your local fonts to include extra bold.  Check Team folder in Resilio.

**Steps to test this PR**:

Variant `sc` control:
1. No second button on boarding
1. Confirm contextual tips shown after dismissing onboarding

Variant `mr`:
1. Shows explore privacy button
1. Tap it to launch a web container which shows an app info page (if not merged yet will show a search result for `app_info`
1. Dismiss to return to onboarding
1. Confirm contextual tips shown after dismissing onboarding

Variant `mq`:
1. Shows customize settings button
1. Tap it to launch settings
1. Dismiss to return to onboarding
1. Check contextual tips shown appropriately even if initial layout is changed to a centered search

Pixels:

* `mo` - fired when onboarding is shown
* `mo_s` - fired first time only when customize settings button pressed
* `mo_e` - fired first time only when explore privacy button pressed

Home Row CTA

* This has now been deferred to a day that is not the current installation day. Change date on device and relaunch app to show it.

---
###### Internal references:


DuckDuckGo/Onboarding.swift
```diff
@@ -72,19 +79,51 @@ extension MainViewController {
 
 extension MainViewController: OnboardingDelegate {
     
+    static var onboardingSettingsPixelFired = false
+    static var onboardingExplorePrivacyPixelFire = false
```
bwaresiak: Initially I haven't seen that this is all included in Onboarding.swift - hence some thoughts about it:

I don't have strong opinion about it, but I'd consider moving all this "presentation/fire pixel" logic to OnboardingVC (or some Onboarding-related logic class) directly, as this seems to me like a logic that could be encapsulated within such class. This way MainVC will have less Onboarding logic. :)

If you think there is any merit to that, then few noes from me:
1. I know Settings screen needs to be bound to MainVC, but we probably can just do this through delegate (e.g. `onboarding(_ VC, willDisplaySettings: SettingsVC)` 
2. Since Onboarding is living only at the start, we probably can replace static vars with lazy ones (being a poor man's version of `dispatch_once`):
```
    private var fireSettingOpenedOnce: ()-> Void = {
        Pixel.fire(pixel: "mo_s")
        return {}
    }()
```
And just invoke `fireSettingOpenedOnce()`

DuckDuckGo/TabSwitcherViewController.swift
```diff
@@ -52,10 +55,12 @@ class TabSwitcherViewController: UIViewController {
         currentSelection = tabsModel.currentIndex
         applyTheme(ThemeManager.shared.currentTheme)
         becomeFirstResponder()
+        tipsEnabled = tipsStorage.isEnabled
     }
 
     override func viewWillAppear(_ animated: Bool) {
         super.viewWillAppear(animated)
+        tipsStorage.isEnabled = false
```
bwaresiak: Only concern about this, is that it is persisted and in case user kills the app (or leaves it, so it is then killed by the system) while being on Tabs screen, then no pending tips will be shown.

bwaresiak (commented):
Great, everything seems to work as expected. :)

I've left two comments below, nothing major just some thoughts.

DuckDuckGo/Onboarding.swift
brindy: Great feedback.   The lazy function call there is 👌  but I've gone for a couple of instance vars in the VC instead as I think it's a bit clearer?

DuckDuckGo/TabSwitcherViewController.swift
brindy: The ideal situation would be if the tab switcher caused "viewDidDisappear" to happen on the parent controller but it doesn't because we need the underlying context due to it being semi transparent.  Any other ideas (short of redesigning the entire tab switcher :) )?

brindy: Feels convoluted, but I'm going to use the TabSwitcher delegate to call back up to the MainVC then down in to the TabVC to install/remove tips.

bwaresiak (approved):
Great! LGTM :)

PR #429, opened 2019-04-01
accessibility improvements

brindy:
Task/Issue URL: https://app.asana.com/0/392891325557410/1114078510432707
Tech Design URL:
CC:

**Description**:

Improve accessibility for the app.  I will reach out to an interested user who will evaluate the app for us and give us some feedback directly.

General concepts:

* Anything that's an action should have the "button" trait. During voice over Siri will say "button" after the item description. 

* Images that don't add any extra info should be ignored

* Don't worry about what's on web pages - that's up to the web pages (though feel free to report accessibility problems on the SERP directly on Asana)

**Steps to test this PR**:
1. Use Siri to turn on Voice Over (note, use Siri to turn it off again, it's the easiest way)
1. Check that all buttons on the home screen are accessible (even if disabled)
1. Check all visible buttons in the omnibar are accessible
1. Change the home screen type and ensure that the center search is accessible
1. Browse to a page and check that buttons on a page are accessible
1. Open the tab switcher and check that it's possible to select tab and close a tab
1. Open bookmarks and check that it's possible to select a bookmark
1. Open settings
1. Open feedback from the settings

---
###### Internal references:


DuckDuckGo/EditBookmarkAlert.swift
```diff
@@ -49,12 +49,15 @@ private class ValidatingAlert: UIAlertController {
...
         addTextField { textField in
+            textField.accessibilityLabel = "Bookmark Title" // UserText.bookmarkTitleAccessibility
             textField.text = link?.title
             textField.placeholder = UserText.bookmarkTitlePlaceholder
             textField.keyboardAppearance = keyboardAppearance
             self.titleField = textField
         }
         addTextField { textField in
+            textField.accessibilityLabel = "Bookmark Address"
+//            textField.acc
```
bwaresiak: Please remove :)

DuckDuckGo/TabViewCell.swift
```diff
@@ -65,7 +67,11 @@ class TabViewCell: UICollectionViewCell {
         background.layer.borderColor = UIColor.cornflowerBlue.cgColor
         background.alpha = isCurrent ? Constants.selectedAlpha : Constants.unselectedAlpha
 
-        title.text = tab.link?.displayTitle
+        if let link = tab.link {
+            accessibilityLabel = UserText.openTab(withTitle: link.displayTitle ?? "", atAddress: link.url.host ?? "")
+            title.text = tab.link?.displayTitle
+        }
+
```
bwaresiak, line 75: Compared to Safari, our tabs are more difficult to navigate as VoiceOver describe all UI elements within each Tab ell. Safari only mentions tab name and existence of 'Close' button for each tab, but we describe every element (even image view). I'd consider keeping it much more brief, as listing all tabs using VoiceOver seems really inefficient right now.

DuckDuckGo/MainViewController.swift
```diff
@@ -432,9 +437,11 @@ class MainViewController: UIViewController {
     }
 
     private func forgetAllAction() -> UIAlertAction {
-        return UIAlertAction(title: UserText.actionForgetAll, style: .destructive) { [weak self] _ in
+        let action = UIAlertAction(title: UserText.actionForgetAll, style: .destructive) { [weak self] _ in
             self?.forgetAll {}
         }
+        action.accessibilityLabel = UserText.confirm
```
bwaresiak, line 443: Maybe we should be a bit more explicit here, like "Confirm clearing all tabs"? I know people are *supposed* to know the context since they navigated to this action sheet, but still... :)

bwaresiak (commented):
Nice! Definitely very useful changes! Ive added some comments/thoughts inline.

DuckDuckGo/TabViewCell.swift
brindy: Good shout - will update.

DuckDuckGo/MainViewController.swift
brindy: I checked that with my contact and they said Confirm was sufficient, presumably to your previous point about being concise.

DuckDuckGo/TabViewCell.swift
brindy: Only the title field and close button are accessible now.  The message is says "Open/Close <website title> at <domain name> - button".  We'll get feedback from my contact anyway so don't worry too much if it's not perfect.

bwaresiak: Great!

bwaresiak (approved):
LGTM!

One thing I've noticed and I'm curious about your contact's opinion about it: When site is open and we are listing elements top-down, we start with Omni Bar elements. Safari seems to ignore browser controls and just focuses on the website (unless I've missed something obvious here). I wonder which is more accessible for daily usage :)

brindy:
Thanks!

PR #435, opened 2019-04-12
onboarding experiment 3

brindy:
Task/Issue URL: https://app.asana.com/0/392891325557410/1117195387371337
Tech Design URL: 
CC:

**Description**:

Tests a new theme selection screen as part of onboarding.

Also adds new "load" pixel to allow for timed pixels, as well as adds a test parameter which is enabled in debug mode so they can be excluded.  Also uses new top level `m` pixel.

**Steps to test this PR**:
Two variants to test:

* `sc`: Summary screen only

1. Click start browsing.  App launches in to home page

* `mr`: Theme selection + Summary screen

1. Continue button disabled until one of the themes selected
1. Skip button moves to Summary screen
1. Selecting a theme has haptic feedback and enabled continue button
1. If a theme was selected (ie test with light) then it is applied after summary screen

Pixels:

*    m_o - on boarding shown
*    m_o_t - fired when themes screen is dismissed regardless of user choices, includes timing parameter
*    m_o_t_d - fired if dark theme selected
*    m_o_t_l - fired if light theme selected
*    m_o_t_s - fired if themes are skipped
*    m_o_s - fired when summary screen is dismissed, includes timing parameter

Devices:

* SE

1. All items should be visible on both screens but spacings adjusted to fit

* Other phones

1. All items shown, spacing better used

* iPad

1. Should behave like a larger phone but centred on screen with outer-padding.  The animation when moving from one screen to another is deliberately clipped.

---
###### Internal references:


Core/Pixel.swift
```diff
@@ -151,14 +156,31 @@ public class Pixel {
...
                             onComplete: @escaping (Error?) -> Void = {_ in }) {
+        
+        var newParams = params
+        if isDebugBuild {
+            newParams["test"] = "1"
+        }
+        
+        if let dateLoaded = loaded[pixel] {
+            newParams["dur"] = String(Date().timeIntervalSince(dateLoaded))
+            loaded[pixel] = nil
```
bwaresiak: This change actually caused this API to no longer be thread safe - are we fine with it? If we are sure that all Pixels are caused by user activity and originate from Main Thread this should be OK, but some kind of warning in documentation would be good. Otherwise, we may need to make it thread safe.

There's also an alternative - please check the comment below.

Core/Pixel.swift
```diff
@@ -151,14 +156,31 @@ public class Pixel {
...
         Alamofire.request(url, headers: headers).validate(statusCode: 200..<300).response { response in
             Logger.log(items: "Pixel fired \(pixel.rawValue)")
             onComplete(response.error)
         }
     }
+    
+    /// Don't load a pixel unless you intend to fire it!
+    public static func load(pixel: PixelName) {
+        loaded[pixel] = Date()
+    }
```
bwaresiak, line 189: Alternative approach would be to return object (`TimedPixel` ?) that encapsulates PixelName and start date, then `fire()` it when needed. If not used (or forgotten) no harm is done. This approach should also keep Pixel API thread safe (but still does not guarantee that this `TimedPixel` is thread safe, but this should be fine if used in a context of a single thread).

Core/VariantManager.swift
```diff
@@ -20,9 +20,13 @@
...
 
 public enum FeatureName: String {
+    
     case onboardingSummary
-    case onboardingCustomizeSettings
-    case onboardingExplorePrivacy
+    case onboardingThemes
+    
+    // Used for unit tests
+    case dummy
```
bwaresiak, line 28: I assume this is so we don't have to tweak tests every time we change the Feature set?

DuckDuckGo/OnboardingThemesViewController.swift
```diff
@@ -0,0 +1,71 @@
...
+        delegate?.setContinueEnabled(true)
+        darkThemeRadio.accessibilityTraits.formUnion(.selected)
+        lightThemeRadio.accessibilityTraits.remove(.selected)
+        feedback.selectionChanged()
+        selectedTheme = .dark
+        exitPixel = .onboardingThemesDarkThemeSelected
+    }
+    
+    override func finished() {
+        AppUserDefaults().currentThemeName = selectedTheme
```
bwaresiak: ThemeManager is supposed to take care of updating `currentThemeName` and reload all views. Hence all you should need here is to call `ThemeManager.shared.enableTheme(with: selectedTheme)` and get rid of all `reloadTheme()` logic from `MainViewController`

DuckDuckGo/OnboardingThemesViewController.swift
```diff
@@ -0,0 +1,71 @@
...
+        super.viewDidLoad()
+        feedback.prepare()
+        Pixel.load(pixel: .onboardingThemesFinished)
+        canContinue = false
+    }
+    
+    override func viewDidDisappear(_ animated: Bool) {
+        super.viewDidDisappear(animated)
+        Pixel.fire(pixel: .onboardingThemesFinished)
+        Pixel.fire(pixel: exitPixel)
```
bwaresiak, line 44: What about following scenario:
1. Open app.
2. Select Light theme.
3. Instead of 'Done' press 'Skip'.

Pixels are fired for the Light Theme selection (`m_o_t_l`), but since 'Skip' button has been pressed, `finished()` is not being called - thus theme is not set. I'd expect this scenario to fire `m_o_t_s` pixel.

DuckDuckGo/OnboardingViewController.swift
```diff
@@ -20,174 +20,145 @@
...
-        shrinkImages(withRatio: ratio)
-    }
 
-    private func transitionBackgroundColor(withRatio ratio: CGFloat) {
-        guard let nextColor = transitioningToPage?.preferredBackgroundColor else { return }
-        guard let currentColor = currentPage.preferredBackgroundColor else { return }
-        view.backgroundColor = currentColor.combine(withColor: nextColor, ratio: ratio)
+        if sender == continueButton {
+            contentController?.finished()
+        }
```
bwaresiak, line 93: Considering Light Theme/Skip `exitPixel` issue mentioned above, I'd consider making it a bit more explicit when `finished()` is being called, or maybe introducing something like `skipped()`?

bwaresiak (commented):
Nice! Good to see another onboarding experiment!

I've left some comments inline.

UI-wise (iPhone SE, XS Max, iPad Pro/Mini) everything looks good 👍 One thing that caught my eye is that on iPhone SE, the 'Skip' button is pretty close to the bottom, but I understand that it is hard to take care of that as the screen is pretty small.

Core/Pixel.swift
brindy: Great idea. Have updated, see what you think. I think what I've done is thread safe as we're not manipulating any shared data structures.

Core/VariantManager.swift
brindy: Yep

DuckDuckGo/OnboardingThemesViewController.swift
brindy: Gah, done!

DuckDuckGo/OnboardingViewController.swift
brindy: Yep, updated this to be more explicit.

DuckDuckGo/OnboardingThemesViewController.swift
brindy: Nice one - fixed.

brindy:
Yep, there's not much more I can do with SE I'm afraid. It's been approved by design with a view to tweaking if the experiment is successful.

Core/Pixel.swift
```diff
@@ -151,14 +154,43 @@ public class Pixel {
...
         Alamofire.request(url, headers: headers).validate(statusCode: 200..<300).response { response in
             Logger.log(items: "Pixel fired \(pixel.rawValue)")
             onComplete(response.error)
         }
     }
+    
+    /// Don't load a pixel unless you intend to fire it!
+    public static func load(pixel: PixelName) -> TimedPixel {
+        return TimedPixel(pixel)
+    }
```
bwaresiak: Looks good @brindy! 

I'd only consider either removing this method (and use `TimedPixel` directly) or tweaking the name to something like `makeTimedPixel` (in case of the latter, I assume we can't make `TimedPixel`'s `init` method fileprivate since it is used in the tests).

bwaresiak (approved):

brindy:
Thanks @bwaresiak !

PR #449, opened 2019-06-19
favourites by default

brindy:
Task/Issue URL: https://app.asana.com/0/392891325557410/1127121491900615
Tech Design URL:
CC:

**Description**:

An experiment to test favorites and center search by default.

When adding a favorite from the home screen it prompts for a URL only, then takes the host name with `www.` stripped as the title.

Variants:

* `ma`: current production
* `mb`: favorites and center search by default

Pixels:

    case homeScreenShown = "mh"
    case homeScreenSearchTapped = "mh_st"
    case homeScreenFavouriteLaunched = "mh_fl"
    case homeScreenAddFavorite = "mh_af"
    case homeScreenAddFavoriteOK = "mh_af_o"
    case homeScreenAddFavoriteCancel = "mh_af_c"
    case homeScreenEditFavorite = "mh_ef"
    case homeScreenDeleteFavorite = "mh_df"

Note: currently in design review

**Steps to test this PR**:
1.  Fresh install, ensure that both variants manifest.
1. Ensure pixels are fired at appropriate times.
1. Test migration
1. General testing of favorites functionality.

---
###### Internal references:


Core/Pixel.swift
```diff
@@ -90,7 +90,16 @@ public enum PixelName: String {
...
-
+    
+    case homeScreenShown = "mh"
+    case homeScreenSearchTapped = "mh_st"
+    case homeScreenFavouriteLaunched = "mh_fl"
+    case homeScreenAddFavorite = "mh_af"
+    case homeScreenAddFavoriteOK = "mh_af_o"
+    case homeScreenAddFavoriteCancel = "mh_af_c"
+    case homeScreenEditFavorite = "mh_ef"
+    case homeScreenDeleteFavorite = "mh_df"
```
bwaresiak, line 101: Do we want to differentiate between launching favorites from Home Screen and launching them from the Overlay?

DuckDuckGo/EditBookmarkAlert.swift
```diff
@@ -36,24 +36,30 @@ class EditBookmarkAlert {
...
         let keyboardAppearance = ThemeManager.shared.currentTheme.keyboardAppearance
-        addTextField { textField in
-            textField.accessibilityLabel = "Bookmark Title" // UserText.bookmarkTitleAccessibility
-            textField.text = link?.title
-            textField.placeholder = UserText.bookmarkTitlePlaceholder
-            textField.keyboardAppearance = keyboardAppearance
-            self.titleField = textField
+        
+        if nil != link {
+            // only show this when editing
```
bwaresiak: I have serious doubts about usability after this change. Now, in order to customize text while adding a new favorite entry, I need to do twice as much work: 
1. Add entry.
2. Long press fav icon to enter edit mode.
3. Edit title, save.

I don't see the benefit here. If we want to make it easier for the user, then I'd not optimize one use case (for people who are happy with domain in name) and instead optimize for everyone. I suggest:
1. Keep both fields in add mode, but swap them (URL first).
2. Once url is filled, it is automatically copied to fill the 'Title'.
3. If user is happy, he just confirms, otherwise he has a chance to change it.

Any thoughts? I'd be very happy to talk about this if needed. :)

Core/Pixel.swift
```diff
@@ -90,7 +90,16 @@ public enum PixelName: String {
     case onboardingThemesLightThemeSelected = "m_o_t_l"
     case onboardingThemesSkipped = "m_o_t_s"
     case onboardingSummaryFinished = "m_o_s"
-
+    
+    case homeScreenShown = "mh"
+    case homeScreenSearchTapped = "mh_st"
+    case homeScreenFavouriteLaunched = "mh_fl"
+    case homeScreenAddFavorite = "mh_af"
```
bwaresiak, line 97: Is this used anywhere?

DuckDuckGo/EditBookmarkAlert.swift
```diff
@@ -89,15 +96,22 @@ private class ValidatingAlert: UIAlertController {
         }
     }
     
+    private func urlBasedTitle(_ urlString: String?) -> String? {
+        guard let urlString = urlString else { return nil }
+        return URL.webUrl(fromText: urlString)?.host?.replacingOccurrences(of: "www.", with: "")
+    }
```
bwaresiak: Totally minor: pasting 'https://i❤.ws' does not populate the title. :)

DuckDuckGo/CenteredSearchHomeCell.swift
```diff
@@ -74,6 +75,7 @@ class CenteredSearchHomeCell: UICollectionViewCell {
     }
 
     @objc func onTap() {
+        Pixel.fire(pixel: .homeScreenSearchTapped)
```
bwaresiak, line 78: This fires for both 'centered search' and 'centered search w/ favorites' modes - is that intended (asking since this project is about making favorites the default, so I worry about possibility if misinterpreting the data afterwards)?

bwaresiak (commented):
Nice, seems like an exciting experiment! 👍 I've left some comments inline.

DuckDuckGo/EditBookmarkAlert.swift
brindy: Chris seems to have come to the same conclusion, so I'm reverting that.

Core/Pixel.swift
brindy: I don't think so - what we're interested in is if people are using favourites or not, so doesn't matter where they're doing that from.

Core/Pixel.swift
brindy: Good spot, thanks.

DuckDuckGo/EditBookmarkAlert.swift
brindy: :) no problem since that change is being reverted anyway

DuckDuckGo/CenteredSearchHomeCell.swift
brindy: I think that's ok as we'll see if people change the settings and then use the centered search.  Also I want to keep the separation of concerns, this thing shouldn't know about other renderers that are enabled.

bwaresiak (approved):
LGTM 👍

PR #459, opened 2019-06-28
Etag OOS fix

bwaresiak:
Task/Issue URL: 
- https://app.asana.com/0/856498667320406/1128900793643895
- https://app.asana.com/0/856498667320406/1109742356053775

**Description**:
- Fixes a problem of missing content blocker list despite existing etag. 
- Handling weak etags, by ignoring the 'W/' prefix.
- Do not store etags for http bloom filter requests (as they use spec instead for update processing).
- Introduce statistic for configuration fetches.

**Steps to test this PR**:

Regular scenario
1. Comment out `try disconnectMeStore.persist(data: data)` from `StorageCache`.
2. Run the app (fresh install) - list should be downloaded but not stored. Content blocker should not be available.
3. Restart app and check etag storage logs - no etag for disconnect me list should be returned.

Migration scenario
1. Check out develop.
2. Comment out `try? disconnectMeStore.persist(data: data)` from `StorageCache`.
3. Run the app (fresh install) - list should be downloaded but not stored. Content blocker should not be available.
4. Restart app and check etag storage logs for disconnectMe list query - entry should be present (this is bug).
5. Checkout this branch (and clear changes).
6. Run the app - disconnectMe should be fetched and stored despite matching etag. 
7. Ensure content blocking is working.

Weak etags
1. Run the app (fresh install), see that `httpsWhitelist` is downloaded (with Weak etag) and stored.
2. Restart the app - see in logs that `httpsWhitelist` etag entry in etag storage has no weak prefix.
3. Observe that `httpsWhitelist` was not updated after restart.

Config Fetch info
- Run the app -> on startup info Pixel should be fired.
- In case of background download, pixel is not fired, but info is incremented.

---
###### Internal references:


Core/StorageCache.swift
```diff
@@ -54,45 +65,56 @@ public class StorageCache {
...
-            easylistStore.persistEasylistWhitelist(data: data)
+            guard let data = data as? Data else { return false }
+            return easylistStore.persistEasylistWhitelist(data: data)
             
         case .disconnectMe:
-            guard let data = data as? Data else { return }
-            try? disconnectMeStore.persist(data: data)
+            guard let data = data as? Data else { return false }
+            do {
+                try disconnectMeStore.persist(data: data)
```
brindy, line 79: Use `try?` instead of a `do { ... } catch { } `

DuckDuckGoTests/ContentBlockerLoaderTests.swift
```diff
@@ -0,0 +1,181 @@
...
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import XCTest
+@testable import Core
+
+class ContentBlockerLoaderTests: XCTestCase {
```
brindy, line 23: 😻

brindy (commented):
This is great work! Thanks for the brainstorming session earlier.  Just one code change for now, I want to run it on my device and through Charles etc first before approving.

Core/StorageCache.swift
bwaresiak: Actually I need this to return true/false. But thanks for pointing that out, as I could move that return statement from the bottom to 'catch', and all code paths are now encapsulated within switch.

brindy (approved):
LGTM! Check with @subsymbolic before merging as I think she had some things she wanted to try.

bwaresiak:
I will, thanks @brindy!

subsymbolic (approved):
@bwaresiak I have tested this with Charles, killing and allowing requests through to make sure they are updated, that ETags work and that UI is correct. It looks good. Great work.

bwaresiak:
Thanks @subsymbolic :)

Issue #467, opened 2019-07-30
HTTP Basic Authentication isn't supported


salty-horse:
Similar bug for the Android app, where this was already implemented: https://github.com/duckduckgo/Android/issues/383

Using 7.24.0.

When visiting a website that requires Basic Authentication, you immediately get a 401 Unauthorized error.

bwaresiak:
This is now tracked internally at https://app.asana.com/0/856498667320406/1133521518040402

Thank you for your report!

brindy:
Basic support is now in the app!  Thanks for raising this @salty-horse.

PR #472, opened 2019-08-19
Support for basic authentication

bwaresiak:
Task/Issue URL: https://app.asana.com/0/856498667320406/1133521518040402

**Description**:
Support for basic authentication as requested in #467 

**Steps to test this PR**:
1. Open website requesting basic authentication.
2. Check if authentication is possible.

Note: there are two variants of alert message: one for HTTP and another one for HTTPS connection.

---
###### Internal references:


brindy:
1) This isn't working quite the way I would expect. It seems to be fine for single pages, but if you navigate to a page in the same domain, you have to authenticate again.  For instance try this page in Safari or some other browser:

https://falkirkrpg.org.uk/noxwall/secure/

username: test
password: test

When you navigate to the other pages it remembers you are authenticated and I think this is what we should do as well.  

Basically, if the user enters a valid authentication we should remember it while that tab is open and send it whenever they access that domain, presumably in response to a challenge on each navigation.

Ideally we'd also have some kind of UI/UX which lets the user tell us to forget the credentials without closing the tab, but I think a tab close is good enough for now.

2) Also, a convention for Basic auth is to allow address entry in the following format:

https://username:password@domain/path

e.g. https://test:test@falkirkrpg.org.uk/noxwall/secure/index.html

However, that always seemed like a poor idea to me so I'm ok for us to ignore that.

bwaresiak:
Thanks @brindy 

About navigation: I see I've messed up option for credential persistence - instead of 'none' I should've used 'forSession'. This is now fixed.

brindy (approved):
This LGTM.  I have a slight concern about the session thing as that means credentials could hang around for a while and the only way to clear them is the fire button.  Which might be ok, but we should monitor feedback about it.

brindy:
Merging since we have the same implementation as Android.

PR #475, opened 2019-08-24
Add https lookup service

subsymbolic:
Task/Issue URL: https://app.asana.com/0/inbox/361428290920650/1132869367787864/1136983228981334
Tech Design URL: https://app.asana.com/0/inbox/361428290920650/890441675382858/1136983228981334

**Description**:
Add a https lookup service

**Steps to test this PR**:

**Test item in local list is upgraded**
1. Visit http://facebook.com
1. Check logs "facebook.com is locally upgradable"
1. Check that https local lookup pixel "m_https_lu" is sent
1. Note this request subsequently loads https and redirects so will also trigger some "m_https_nl" pixels will also fire

**Test when item not in local list but in service then is upgraded and correct pixels are sent****
1. Visit http://channel4.com
1. Check logs "channel4.com is not locally upgradable" and "channel4.com is service upgradable"
1. Check that https service request pixel "m_https_sru" is sent
1. Note this request subsequently loads https and redirects so will also trigger some "m_https_nl" pixels will also fire

**Test when item not in local list but in service cache, it is upgraded and correct pixels are sent**
1. Visit http://channel4.com again
1. Check logs "channel4.com is not locally upgradable" and "channel4.com is service upgradable"
1. Check that this time the https service request pixel "m_https_scu" is sent
1. Note this request subsequently loads https and redirects so will also trigger some "m_https_nl" pixels will also fire

**Test when item not in local list or in service list, it is not upgraded and correct pixels are sent**
1. Visit http://example.com
1. Check logs "example.com is not locally upgradable" and "example.com is not service upgradable"
1. Check that the https service not upgraded request pixel "m_https_srn" is sent

**Test when item not in local list and not in service cache, it is not upgraded and correct pixels are sent**
1. Visit http://example.com again
1. Check logs "example.com is not locally upgradable" and "example.com is not service upgradable"
1. Check that this time the https service not upgraded cache pixel "m_https_scn" is sent

**Test when item in whitelist then is not upgraded and correct pixels are sent****
1. Visit http://www.bjtax.gov.cn
1. Check that page is not upgraded
1. Check that https no lookup pixel "m_https_nl" pixel is sent

---
###### Internal references:


Core/HashExtension.swift
```diff
@@ -0,0 +1,52 @@
...
+    }
+    
+    private func hash(algorithm: Algorithm, length: Int32) -> String {
+        var hash = [UInt8](repeating: 0, count: Int(length))
+        let dataBytes = [UInt8](self)
+        _ = algorithm(dataBytes, CC_LONG(self.count), &hash)
+        let output = NSMutableString(capacity: Int(length))
+        hash.forEach { byte in output.appendFormat("%02x", byte) }
+        return output as String
+    }
```
brindy, line 40: Optional: If you wanted an immutable approach to this, you could do something like:

`return hash.map { String(format: "%02x", $0) }.joined()`

DuckDuckGo/TabViewController.swift
```diff
@@ -973,6 +984,14 @@ extension TabViewController: UIGestureRecognizerDelegate {
...
         }
     }
+    
+}
 
+extension URL {
+    func toHttps() -> URL? {
+        let urlString = absoluteString
+        return URL(string: urlString.replacingOccurrences(of: "http", with: "https", options: .caseInsensitive, range: urlString.range(of: "http")))
+    }
```
brindy: This can can create a malformed url if the scheme is already https.  I realise it's only going to be used when upgrading URLs, but that extension has global scope so could be dangerous. 

Using `URLComponents` would be safer:

```
        var components = URLComponents(url: self, resolvingAgainstBaseURL: false)
        components?.scheme = "https"
        return try? components?.asURL()
```

And since it's globally available, could have a unit test :)

brindy (commented):
Super duper! Couple of small comments.... one optional.

DuckDuckGo/TabViewController.swift
subsymbolic: Oh good point, I just moved it, assumed it was solid without reading it. Will update and test.

subsymbolic:
@brindy over to you for rereview

brindy (approved):
LGTM!

subsymbolic:
Thanks for the review. I'll merge and released this when I get the go ahead from the DRI.

PR #486, opened 2019-09-19
iOS 13 SDK Support 

bwaresiak:
Task/Issue URL: 
- https://app.asana.com/0/856498667320406/1128551253387167
- https://app.asana.com/0/856498667320406/1140757945084788
- https://app.asana.com/0/856498667320406/1140857123487538

**Description**:
Support for iOS 13 and Xcode 11, fixes to cookie handling and long press action sheet.

**Steps to test this PR**:

iOS 13 & UI
- Status bar style should contrast with screens like tabs list, home row instructions, settings etc.
- No more glitches on Privacy dashboard in dark/system dark themes.
- Edit bookmarks cell accessory color should always be visible (issue present when system light style was combined with app dark theme).
- Our Widgets should look good for both system themes.
- Cells padding for bookmarks.
- Actions sheets/alerts should be matching any selected in-app theme.
- Long press should work in Web View, query string parameters should be stripped for better clarity.

DDG Cookies preservation
- Set cookies through DDG settings (e.g. theme, languate etc).
- After fire button is pressed, these should be restored on first navigation.
- Check also new tab from empty state and app restoration after restart scenarios.

Supporting System theme
- Check settings screen functionality on both iOS 12 and 13.
- App UI should automatically match system theme and update if needed.
- Floating tip that points to the settings should be generic.

Onboarding
- Theme should not be covered during onboarding anymore. 

Pixels
- Entering Theme settings screen.
- Changing theme and leaving settings screen (also by swiping down).
- Changing theme, but then reverting the change, then leaving the screen.

Tests
- Unit & ATB tests should pass.

---
###### Internal references:


Core/Pixel.swift
```diff
@@ -55,9 +55,12 @@ public enum PixelName: String {
...
     
     case settingsOpened = "ms"
-    case settingsThemeToggledLight = "ms_tl"
-    case settingsThemeToggledDark = "ms_td"
     case settingsHomeRowInstructionsRequested = "ms_hr"
+    
+    case settingsThemeShown = "ms_tp"
+    case settingsThemeChangedSystemDefault = "ms_ts"
+    case settingsThemeChangedLight = "ms_tl"
+    case settingsThemeChangedDark = "ms_td"
```
subsymbolic, line 63: Are these captured in Grafana somewhere?

Core/WebCacheManager.swift
```diff
@@ -33,14 +33,36 @@ public class WebCacheManager {
...
+                        
         for cookie in cookieStorage.cookies {
-            WebCacheManager.dataStore.httpCookieStore.setCookie(cookie)
+            WebCacheManager.dataStore.httpCookieStore.setCookie(cookie) {
+                semaphore.signal()
+            }
+        }
+        
+        DispatchQueue.global(qos: .userInitiated).async {
+            for _ in 0 ..< cookieStorage.cookies.count {
```
subsymbolic: You're depending on the `cookieStorage.cookies.count` still being the same here however that's not guaranteed by the API. Take a copy of the array or count before the previous for loop.

DuckDuckGo/TabViewController.swift
```diff
@@ -216,16 +217,19 @@ class TabViewController: UIViewController {
     
     private func consumeCookiesThenLoadUrl(_ url: URL?) {
         webView.configuration.websiteDataStore.fetchDataRecords(ofTypes: WKWebsiteDataStore.allWebsiteDataTypes()) { _ in
-            WebCacheManager.consumeCookies()
-            if let url = url {
-                self.load(url: url)
+            WebCacheManager.consumeCookies { [weak self] in
```
subsymbolic, line 220: Good change!

AtbIntegrationTests/AtbIntegrationTests.swift
```diff
@@ -276,6 +275,11 @@ class Springboard {
             let springboardFrame = springboard.frame
             icon.press(forDuration: 1.3)
             
+            let rearrangeButton = springboard.buttons["Rearrange Apps"]
```
subsymbolic, line 278: Strange this new "Rearrange Apps" button from Apple!

subsymbolic:
Looks great. Just beware the app is defaulting to the Dark theme, it looks like AppUserDefaults.urrentThemeName is returning dark by default.

Core/WebCacheManager.swift
bwaresiak: Oh good catch, thanks!

Core/Pixel.swift
bwaresiak: I don't think so. I'd like just to use this as engagement factor to figure out whether we should optimize in the future the app <-> serp theme sync issue.

AtbIntegrationTests/AtbIntegrationTests.swift
bwaresiak: Yes! Seems they struggled with force touch and just decided to change it.. but it's weird. :)

subsymbolic:
One wee bug:
iOS_13_system_theme_dark

bwaresiak:
Thanks for the review! Besides addressing inline feedback I've changed the default theme on iOS 13 and fixed the glitch.

subsymbolic (approved):
Looks great! Feel free to merge once CI is passing.

Core/WebCacheManager.swift
```diff
@@ -33,14 +33,38 @@ public class WebCacheManager {
...
+        
+        DispatchQueue.global(qos: .userInitiated).async {
+            for _ in 0 ..< cookies.count {
+                semaphore.wait()
+            }
+            
+            DispatchQueue.main.async {
+                cookieStorage.clear()
+                completion()
+            }
```
roop, line 66: Minor nitpick / suggestion:

When waiting for multiple items, I find that using DispatchGroup makes code a little more readable, compared to a semaphore. Something like:

``` Swift
        let group = DispatchGroup()

        for cookie in cookies {
            group.enter()
            WebCacheManager.dataStore.httpCookieStore.setCookie(cookie) {
                group.leave()
            }
        }

        group.notify(queue: DispatchQueue.main) {
            cookieStorage.clear()
            completion()
        }
```

bwaresiak: Yes, this is very good suggestion. Since this is already merged I'm not going to touch that right now, but I'll track it internally and address it next time I'm doing some cleaning.

PR #490, opened 2019-10-04
Decrease debounce time on iOS 13

bwaresiak:
Task/Issue URL: https://app.asana.com/0/856498667320406/1142493726420761
Tech Design URL:
CC:

**Description**:
Decrease Debounce delay on iOS 13 to improve responsiveness.

**Steps to test this PR**:
1. Make sure autocomplete works.

---
###### Internal references:

### DIFF
--- a/DuckDuckGo/AutocompleteViewController.swift
+++ b/DuckDuckGo/AutocompleteViewController.swift
@@ -23,7 +23,12 @@ import Core
 class AutocompleteViewController: UIViewController {
     
     struct Constants {
-        static let debounceDelay = 0.3
+        static let debounceDelay: TimeInterval = {
+            if #available(iOS 13.0, *) {
+                return 0.1
+            }
+            return 0.3
+        }()
     }
 
     weak var delegate: AutocompleteViewControllerDelegate?


DuckDuckGo/AutocompleteViewController.swift
```diff
@@ -23,7 +23,12 @@ import Core
 class AutocompleteViewController: UIViewController {
     
     struct Constants {
-        static let debounceDelay = 0.3
+        static let debounceDelay: TimeInterval = {
+            if #available(iOS 13.0, *) {
```
subsymbolic, line 27: Do we need to split the logic here? Will 100ms work for lower versions too?

bwaresiak: I'd prefer to split it, as on iOS 12 this is workaround for a crash within Apple SDK. It is supposed to be fixed on iOS 13. If this change does not result in crashes we can consider removing this from iOS 13+ completely.

subsymbolic: I may need some more context here. The previous workaround was the debounce right? Did the 300ms delay eliminate or reduce the issue?

bwaresiak: Yes, debounce is a workaround. 300ms greatly reduced amount of crashes (only ~10 for 7.28.0 thus far). 

About timing: I've managed to reproduce crash for 100ms within <1h of testing. 300ms seemed to resolved it.

subsymbolic: Thanks for the clarification, if we still see these then it makes sense to limit the debounce reduction to iOS 13 only 👍

subsymbolic (approved):

DuckDuckGo/AutocompleteViewController.swift
bwaresiak: Thanks for the review!

PR #494, opened 2019-10-14
handle link previews in iOS 13 (mainly to fix save images)

brindy:
Task/Issue URL: https://app.asana.com/0/392891325557410/1144355399715969
Tech Design URL:
CC:

**Description**:

In iOS 13 our long press gesture appears to not work for images, as such users can't save them.  To enable this functionality again we have to handle link previews more generically.

This change shows a preview and our usual list of long press items for links.

For images iOS handles showing the preview and list of clickable items.

Note that "Open" has been dropped because tapping the preview opens the link.  We've also dropped Add To Reading List as it's not used very often and pushes people to Safari.

**Steps to test this PR**:
1. Search for something and then long press a link.
1. Check each menu item.
1. Check iPhone, iPad, Portrait, Landscape.
1. Check previewing images
1. Retest on iOS 12 or earlier - previous behaviour should still remain

---
###### Internal references:

### DIFF
--- a/DuckDuckGo/TabManager.swift
+++ b/DuckDuckGo/TabManager.swift
@@ -49,7 +49,6 @@ class TabManager {
         controller.attachWebView(configuration: configuration, andLoadUrl: url, consumeCookies: model.isEmpty)
         controller.delegate = delegate
         controller.loadViewIfNeeded()
-        decorate(tabController: controller, with: ThemeManager.shared.currentTheme)
         return controller
     }
 
@@ -171,27 +170,9 @@ class TabManager {
 extension TabManager: Themable {
     
     func decorate(with theme: Theme) {
-
         for tabController in tabControllerCache {
-            decorate(tabController: tabController, with: theme)
+            tabController.decorate(with: theme)
         }
     }
     
-    private func decorate(tabController controller: TabViewController,
-                          with theme: Theme) {
-        // This is done in a crude way: ideally tabController should implement
-        // 'Themable' protocol, but that would require moving Themable to 'Core'
-        // module.
-        controller.view.backgroundColor = theme.backgroundColor
-        controller.error?.backgroundColor = theme.backgroundColor
-        controller.errorHeader.textColor = theme.barTintColor
-        controller.errorMessage.textColor = theme.barTintColor
-        
-        switch theme.currentImageSet {
-        case .light:
-            controller.errorInfoImage?.image = UIImage(named: "ErrorInfoLight")
-        case .dark:
-            controller.errorInfoImage?.image = UIImage(named: "ErrorInfoDark")
-        }
-    }
 }
--- a/DuckDuckGo/TabViewController.swift
+++ b/DuckDuckGo/TabViewController.swift
@@ -142,6 +142,7 @@ class TabViewController: UIViewController {
     
     override func viewDidLoad() {
         super.viewDidLoad()
+        applyTheme(ThemeManager.shared.currentTheme)
         addContentBlockerConfigurationObserver()
         addStorageCacheProviderObserver()
     }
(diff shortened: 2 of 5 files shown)


DuckDuckGo/TabViewController.swift
```diff
@@ -173,8 +174,13 @@ class TabViewController: UIViewController {
         instrumentation.willPrepareWebView()
         webView = WKWebView(frame: view.bounds, configuration: configuration)
         webView.autoresizingMask = [.flexibleWidth, .flexibleHeight]
-        webView.allowsLinkPreview = false
-        attachLongPressHandler(webView: webView)
+        
+        if #available(iOS 13, *) {
+            attachLongPressHandler(webView: webView)
```
bwaresiak: Shouldn't this be attached for iOS 12 instead?

DuckDuckGo/TabManager.swift
```diff
@@ -171,27 +170,9 @@ class TabManager {
...
     
     func decorate(with theme: Theme) {
-
         for tabController in tabControllerCache {
-            decorate(tabController: tabController, with: theme)
+            tabController.decorate(with: theme)
         }
     }
     
-    private func decorate(tabController controller: TabViewController,
```
bwaresiak, line 180: Aww awesome :)

DuckDuckGo/TabViewControllerLongPressMenuExtension.swift
```diff
@@ -22,6 +22,30 @@ import Core
...
+            self?.onBackgroundTabAction(url: url)
+        })
+        items.append(UIAction(title: UserText.actionReadingList, image: UIImage(systemName: "eyeglasses")) { [weak self] _ in
+            self?.onReadingAction(forUrl: url)
+        })
+        items.append(UIAction(title: UserText.actionCopy, image: UIImage(systemName: "doc.on.doc")) { [weak self] _ in
+            self?.onCopyAction(forUrl: url)
+        })
+        items.append(UIAction(title: UserText.actionShare, image: UIImage(systemName: "square.and.arrow.up")) { [weak self] _ in
+             self?.onShareAction(forUrl: url, atPoint: nil)
```
bwaresiak: I'm afraid that passing nil here broke the 'share' action on iPads with iOS 13

bwaresiak (changes requested):
Great to see that addressed 👍 I've left few comments inline.

DuckDuckGo/TabViewController.swift
brindy: Oops - was rushing to push it before heading to dinner 🤦‍♂

DuckDuckGo/TabViewControllerLongPressMenuExtension.swift
```diff
@@ -87,9 +102,58 @@ extension TabViewController {
...
     }
 }
+
+@available(iOS 13.0, *)
+extension TabViewController {
+
+    static let excludedLongPressItems = [
+        UIImage(systemName: "safari"),
+        UIImage(systemName: "eyeglasses"),
+        nil // hide/show link previews
```
bwaresiak: I think this is non-nil on iPad running iOS 13.0 (at least on simulator):
prev

bwaresiak (approved):
One more comment inline and there is also one confusing thing, but I think we should be safe to ignore this: long-pressing 'Download in App Store' banner (e.g. from https://www.speedtest.net/apps/ios ) doesn't seem to invoke the delegate methods. Seems like an iOS bug (unless I'm missing something).

Other than that looks like a great improvement!

brindy:
Thanks - I've updated to look for the icon version if you can give the a test.

I think the App Store link is ok - we just say we can't handle it so it should be obvious that means in the preview window anyway.

bwaresiak (approved):
LGTM!

PR #499, opened 2019-10-16
filter out networks with less than one percent from the leaderboard

brindy:
Task/Issue URL: https://app.asana.com/0/414709148257752/1139123263078640
Tech Design URL:
CC:

**Description**:

Only show networks seen on >= 1% of pages visited.

**Steps to test this PR**:
1. Visit a bunch of sites
1. View network leaderboard
1. Ensure none show "0%"

---
###### Internal references:

### DIFF
--- a/DuckDuckGo/NetworkLeaderboard.swift
+++ b/DuckDuckGo/NetworkLeaderboard.swift
@@ -101,14 +101,13 @@ class NetworkLeaderboard {
         let request: NSFetchRequest<PPTrackerNetwork> = PPTrackerNetwork.fetchRequest()
         request.sortDescriptors = [ NSSortDescriptor(key: "detectedOnCount", ascending: false) ]
         guard let results = try? container.managedObjectContext.fetch(request) else { return [] }
-        return results
+        let pagesVisitedCount = Float(pagesVisited())
+        return results.filter { Float($0.detectedOnCount) / pagesVisitedCount >= 0.01 }
     }
 
-    func shouldShow(minimumDetectionLevel: Double) -> Bool {
-        let pagesVisitedCount = Double(pagesVisited())
-        let detected = networksDetected().filter { Double($0.detectedOnCount) / pagesVisitedCount >= minimumDetectionLevel }
+    func shouldShow() -> Bool {
         let pagesVisitedThreshold = isDebugBuild ? 3 : 30
-        return pagesVisited() > pagesVisitedThreshold && detected.count >= 3
+        return pagesVisited() > pagesVisitedThreshold && networksDetected().count >= 3
     }
     
     func incrementDetectionCount(forNetworkNamed networkName: String) {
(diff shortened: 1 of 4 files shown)


DuckDuckGo/NetworkLeaderboard.swift
```diff
@@ -101,7 +101,8 @@ class NetworkLeaderboard {
         let request: NSFetchRequest<PPTrackerNetwork> = PPTrackerNetwork.fetchRequest()
         request.sortDescriptors = [ NSSortDescriptor(key: "detectedOnCount", ascending: false) ]
         guard let results = try? container.managedObjectContext.fetch(request) else { return [] }
-        return results
+        let pagesVisitedCount = Float(pagesVisited())
+        return results.filter { Float($0.detectedOnCount) / pagesVisitedCount >= 0.01 }
```
bwaresiak, line 105: I haven't got a chance to test it yet, just a question: this looks like this method should return all networks detected, since name indicates so, and this class seems to me like a store interface. I'd say that ignoring occurrences below certain threshold is more of a logic for the Screen that is presenting the data. What do you think?

brindy: That's a fair point.  "shouldShow" logic in here is arguably just for the UI, so I'll shift this detection level logic there.

DuckDuckGo/PrivacyProtectionOverviewController.swift
```diff
@@ -69,8 +69,9 @@ class PrivacyProtectionOverviewController: UITableViewController {
     }
     
     override func shouldPerformSegue(withIdentifier identifier: String, sender: Any?) -> Bool {
-        
-        if identifier == "Leaderboard" && !NetworkLeaderboard.shared.shouldShow() {
+
+        let level = TrackerNetworkLeaderboardView.Constants.minimumDetectionLevel
+        if identifier == "Leaderboard" && !NetworkLeaderboard.shared.shouldShow(minimumDetectionLevel: level) {
```
bwaresiak: This doesn't work for me: identifier is "" when navigating to Leaderboard.

DuckDuckGo/PrivacyProtectionFooterController.swift
```diff
@@ -61,6 +61,10 @@ class PrivacyProtectionFooterController: UIViewController {
 
 class TrackerNetworkLeaderboardView: UIView {
 
+    struct Constants {
+        static let minimumDetectionLevel = 0.01
```
bwaresiak: When I specify more, e.g. 10 or 50 I don't see anything being out-filtered, still seeing entries like 4% Or do I misunderstand this functionality?

DuckDuckGo/PrivacyProtectionOverviewController.swift
brindy: That isn't new logic though.

DuckDuckGo/NetworkLeaderboard.swift
brindy: I've reverted my last commit and gone back to this one.  While this is doing some stuff with persistence it hasn't been strictly abstracted in to a store and was designed to provide data for the UI really. As such 'detected' could also be interpreted as 'detected above a certain level' which is what this is doing. 

This is just quick fix anyway as it's very likely this functionality is going to get ripped out and/or changed "soon". :)

bwaresiak (approved):
LGTM!

brindy:
Thanks!

Issue #497, opened 2019-10-15
Include app version number in iOS system settings


bwaresiak:
Internal reference: https://app.asana.com/0/856498667320406/1145008528714429

**Description**

Goal is to include app version number in iOS system settings. Example from Test Flight:

![settings](https://user-images.githubusercontent.com/2865434/66864971-99718f80-ef96-11e9-913a-5f760cad84ba.png)

nehayward:
I should be able to manage this! Is it okay if I start on a PR for this.

bwaresiak:
@nehayward Thanks for volunteering, but we already have someone working on that in #500

PR #500, opened 2019-10-16
app version number in system settings

Recouse:
Task/Issue URL: https://github.com/duckduckgo/iOS/issues/497

**Description**:
Added app version number to system settings.

**Steps to test this PR**:
1. Go to Settings -> DuckDuckGo

---
###### Internal references:


fsalata (approved):

bwaresiak:
Hello,

Unfortunately looking at the CI, code does not compile:

`❌  /Users/vagrant/git/DuckDuckGo/SystemSettingsManager.swift:18:35: use of unresolved identifier 'AppVersion':`
`UserDefaults.standart.set(AppVersion.shared.majorVersionNumber, forKey: Keys.version)`

Once this is fixed and the CI checks pass I'll be picking this up for the review.

Thanks!

Recouse:
@bwaresiak pushed fixes

Imgur

DuckDuckGo/SystemSettingsManager.swift
```diff
@@ -0,0 +1,21 @@
+//
+//  SystemSettingsManager.swift
+//  DuckDuckGo
+//
+//  Created by Firdavs Khaydarov on 16/10/19.
+//  Copyright © 2019 DuckDuckGo. All rights reserved.
```
bwaresiak: Please make sure Copyright matches what we have in other source files.

bwaresiak (commented):
Thanks for picking that up. 

Besides one inline comment (which by the way may be no longer relevant if we pick different solution), I have one concern about this: app has to be run at least once in order to populate version number properly. 

While this does not seem to be a big issue, I'd still prefer this to be always consistent so we can be more confident when relying on that or debugging startup problems. Rather than doing this from the code, could you alter the project build steps so settings plist is updated when building?

Recouse:
Moved version updating from AppDelegate to build phase. Now, the default value of version setting updates on build phase.

bwaresiak:
@Recouse sorry for late answer, a lot of things happening lately!

We are having some internal discussion about how to integrate this well in our release process. I will try to come back to you within the next 2 days.  

Also, there are two problems I've found:
1. When version number is changed, app has to be build twice in order to see anything change in the settings (seem like bundle packaging takes place before script is being run).
2. On Xcode 11, when I change version number manually through UI (Project -> General, select 'DuckDuckGo' -> change Version) , it becomes `$(MARKETING_VERSION)` in plist file, and is as such put into the Settings bundle:

mark_v

bwaresiak:
Hello again,

In order to resolve that build timing issue and incorporate it into our release process at the same time, let's do this as a manually run script that we will use during the release process:

Usage: `./set_version.sh 7.30.0`

Script should:
- Update projects (main & extensions) plist files with new version number.
- Update settings bundle.

We will commit this afterwards manually along with other changes we make before the release.

Recouse:
And you'll always run the script manually before every release?

bwaresiak:
@Recouse yes. I know this seems sub-optimal, but it is still a bit better than our current release process, where we change version number manually: with script that would be much simpler and tying that with Settings bundle tweaks is a bonus. :)

This still gives us ability to automate that better in the future (for example as part of git flow release process) and if we decide to tweak anything, part of the work would be already done thanks to this script.

Baby steps 😄

Recouse:
Okay, I'll try tonight.

bwaresiak:
I'm closing this. Please reopen if you decide to work on this again.

Issue #496, opened 2019-10-15
Current tab highlighting is broken


bwaresiak:
Internal reference: https://app.asana.com/0/276630244458377/1128277820182171

**Description**

First scenario:

1. Open a bunch of tabs.
2. Navigate to tabs switcher.
3. Select a tab I'm the middle.
4. Close a tab before it.

Expected: the selected tab still has the highlight.
Actual: the highlight moves to the next tab down and it becomes the newly selected tab

Second scenario:

1. Open a bunch of tabs.
2. Navigate to tabs switcher.
3. Select the last tab.
4. Close a tab before it.

Expected: the last tab still has the highlight.
Actual: the highlight drops off.

PR #501, opened 2019-10-17
Fixed the tab highlight in the Tab Switcher

fsalata:
#496 Fixed the tab highlight behavior.


bwaresiak (approved):
Thanks for fixing this 👍

PR #507, opened 2019-10-26
Improve Core Data stack

bwaresiak:
Task/Issue URL: https://app.asana.com/0/414235014887631/1143203875833530
Tech Design URL:
CC:

**Description**:
Make Core Data stack a shared instance, migrate to NSPersistentContainer

**Steps to test this PR**:

"Multiple NSEntityDescriptions claim the NSManagedObject subclass" error should no longer be printed in console.

Migration (iOS 10 - iOS 13): 
- Ensure everything is being migrated properly.
- After migration old DBs should be removed.
- Re-launching should not do any work related to migration if old DBs are cleared.
- Failed migration (interrupted) should resume.

Pixels:
- Replace app container with malformed DB - pixel should be fired. 

---
###### Internal references:

### DIFF
--- a/Core/Database.swift
+++ b/Core/Database.swift
@@ -28,6 +28,7 @@ public class Database {
     
     public static let shared = Database()
 
+    private let lock = NSLock()
     private let container: NSPersistentContainer
     
     init() {
@@ -38,19 +39,21 @@ public class Database {
         
         container = DDGPersistentContainer(name: "Database", managedObjectModel: managedObjectModel)
         
+        loadStore()
     }
     
-    public func loadStore() {
-        let semaphore = DispatchSemaphore(value: 0)
-        container.loadPersistentStores { _, _ in semaphore.signal() }
-        semaphore.wait()
+    private func loadStore() {
+        lock.lock()
+        container.loadPersistentStores { _, _ in self.lock.unlock() }
     }
     
     public func makeContext(concurrencyType: NSManagedObjectContextConcurrencyType, name: String? = nil) -> NSManagedObjectContext {
-        
+        lock.lock()
         let context = NSManagedObjectContext(concurrencyType: concurrencyType)
         context.persistentStoreCoordinator = container.persistentStoreCoordinator
         context.name = name
+        lock.unlock()
+        
         return context
     }
 }
--- a/Core/Database.swift
+++ b/Core/Database.swift
@@ -50,15 +50,22 @@ public class Database {
         container = DDGPersistentContainer(name: name, managedObjectModel: model)
     }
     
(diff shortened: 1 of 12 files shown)


Core/Database.swift
```diff
@@ -0,0 +1,112 @@
...
+    init(name: String, model: NSManagedObjectModel) {
+        container = DDGPersistentContainer(name: name, managedObjectModel: model)
+    }
+    
+    public func loadStore(andMigrate handler: @escaping (NSManagedObjectContext) -> Void = { _ in }) {
+        container.loadPersistentStores { _, error in
+            if let error = error {
+                Pixel.fire(pixel: .dbInitializationError, error: error)
+                // Give Pixel a chance to be sent, but not too long
+                Thread.sleep(forTimeInterval: 1)
```
brindy, line 58: `Pixel.fire` has a completion handler you could use instead - would that be better?

brindy (approved):
This is great!  Couple small proposed changes, but feel free to merge after that.

DuckDuckGo/DatabaseMigration.swift
```diff
@@ -0,0 +1,208 @@
...
+            } catch {
+                Pixel.fire(pixel: .dbMigrationError, error: error)
+                completion(false)
+                return
+            }
+        }
+
+        source.deleteAll(entities: existingEntities)
+        do {
+            try source.save()
```
brindy: could use `try?` here instead?

Core/Database.swift
bwaresiak: I considered that, but:
- I don't want to block this code for too long - just crash if needed.
- I'd need to block the thread here and wait for completion to execute.
- Pixel.fire's completion block is queued on main thread - so if there is already something on main thread waiting for context (and thus - blocking main thread till DB is ready) it could result in a deadlock. I'd have to change Pixel.fire API to return on private queue to make sure that works.

DuckDuckGo/DatabaseMigration.swift
bwaresiak: Yes 👍

bwaresiak:
@brindy I also did not put a limit on repetitive failures that result in sending too many Pixels, figuring out that we will see if that is needed shortly (and it's <1d of work). Do you think that's fine or should I spend time now on that?

brindy:
> @brindy I also did not put a limit on repetitive failures that result in sending too many Pixels, figuring out that we will see if that is needed shortly (and it's <1d of work). Do you think that's fine or should I spend time now on that?

That's fine - if we need to do that later we can.

Core/Database.swift
brindy: Makes sense

bwaresiak:
Thanks @brindy

PR #532, opened 2019-11-26
[Xcode] adds script for updating Settings bundle

KevinVitale:
duckduckgo-settings
<hr/>
Task/Issue URL: #497 

**Description**:
This adds:
 1. a config file that defines `MARKETING_VERSION`–`Version.xcconfig`;
 2. a settings bundle–`Settings.bundle`;
 3. a script file–`set_version.sh`.

The script file pulls `MARKETING_VERSION` from the config file, and writes it to the settings bundle.

**Steps to test this PR**:
1. Run `chmod +x set_version.sh` to ensure it is executable;
2. Run `set_version.sh 7.33.1`

**Known Issues**:
This script should be run prior to build the app for a release, and the resulting changes committed.

---
###### Internal references:


KevinVitale:
Run the script, passing in the version; e.g.,: `./set_version 7.33.1`.

set_version.sh
```diff
@@ -0,0 +1,102 @@
...
+  let xcconfigContent = "MARKETING_VERSION = \(version)"
+  try xcconfigContent.write(to: .xcconfigURL, atomically: true, encoding: .utf8)
+
+  let settings = try PropertyList<Settings>()
+  try settings.write(version: version)
+}
+catch {
+  print(error)
+}
+
```
bwaresiak, line 8: This is pretty nice implementation that uses swift, but seems a bit comlplex. Have you considered using Mac OS tools instead?

Example:
```#!/usr/bin/env bash

if [ -z "$1" ]; then
   echo Usage: ./set_version NUMBER
   echo Example: ./set_version 7.100.1
   exit 1
fi

echo "MARKETING_VERSION = $1" > Configuration/Version.xcconfig
/usr/libexec/PlistBuddy -c "Set :PreferenceSpecifiers:0:DefaultValue $1" DuckDuckGo/Settings.bundle/Root.plist
```

KevinVitale: Ah! _PlistBuddy_ is **so** much simpler—the only caveat being the hard-coded index (which could be easily forgotten about if more things get added in the future).

I ran your implementation and got the same diff output! This is awesome feedback; thanks so much. I'll squash the commits and update the PR.

KevinVitale:
@bwaresiak should be ready for another review; I squashed all script-related changes into a single commit. 

Also,  I rebased `develop` (10e736e) on my branch, so I needed to set the current version to _7.33.1_, but also adjust the four target's "Build Settings" so that they aren't overriding the value for `MARKETING_VERSION`. Please double check that's having the correct effect for your own builds.

bwaresiak:
@KevinVitale great, neat and tidy. :) Bumping version number is much easier now. Thanks for doing that.

I've tested it and it works for me. I've also made sure that app passes AppStore validation. 👍 

There are two things I'm thinking about:
1. I wonder if there is a good (simple, not hacky) way of embedding build number in the settings, so the version has it appended at the end: 7.33.0.0. For me this is optional tho.
2. We need to remember to not change version number from Xcode just use the script (as Xcode will create `MARKETING_VERSION` within project file and override config file). Should not be a problem.

@brindy Since this will become part of our release process, do you have any comments?

brindy:
I haven't ran this myself yet, but off the top of my head:

1) I don't think we need this really.  That extra version number is fairly meaningless the way we do releases.  If we had an encoded date there or something it would be more useful (e.g. 20191129)

2) If this script is part of our release process then it would be good if it failed noisily if someone has changed that by accident and then maybe can be run as part of our CI too? 

Otherwise, looking forward to to this change. It will definitely save time! :)

bwaresiak:
Let's ignore the timestamp/build number thing and rather focus on making this foolproof and failing in case:
- Someone changes the version manually. 
- Xcode somehow adds `MARKETING_VERSION` back to project file.

@KevinVitale My first idea is to simply add script before the build phase, that greps project file during build time and fails in case marketing version is present there: 
```
if grep -q MARKETING_VERSION DuckDuckGo.xcodeproj/project.pbxproj 
then
    echo "Error: 'MARKETING_VERSION' is present in project file."
    exit 1
fi
```

Other idea is to compare `MARKETING_VERSION` value with what we have in the settings plist file to make sure these two match (in case project file overrides our value) - maybe we should actually do both? 🤔 

I'm open for other (better 😄 ) solutions if anything comes to your mind.

KevinVitale:
@bwaresiak comparing the project file's override to what's in the settings plist feels brittle to me, and would be error prone once an additional _"build tag"_ (some computed value, such as an encoded date or the git short commit hash) gets included.

So, acknowledging that `Version.xcconfig` is the source of truth: I think your pre-build script step is the simplest approach. I would expect a step needs to be included on the four existing targets, and any new targets where `MARKETING_VERSION` is a valid value, correct?

I'll give this a go in bit.

KevinVitale:
Pre-build script won't work; looks like we've got an old-fashioned recursion problem 😅

```
grep MARKETING_VERSION DuckDuckGo.xcodeproj/project.pbxproj
			shellScript = "if grep MARKETING_VERSION DuckDuckGo.xcodeproj/project.pbxproj \nthen\n    echo \"Error: 'MARKETING_VERSION' is present in project file.\"\n    exit 1\nfi\n";
```

EDIT: It could be another external script. I will try that.

KevinVitale:
@bwaresiak / @brindy: Added `check_version.sh`, with accompanying _Run Script Phases_ for each relevant target. 🤞

bwaresiak (approved):
Great! This looks good to me, thank you for the contribution. 😄

PR #533, opened 2019-11-26
update to use TDS

brindy:
Task/Issue URL: https://app.asana.com/0/392891325557410/1135079790856649
Tech Design URL: https://app.asana.com/0/392891325557410/1135079790856644
CC:

**Description**:

Removes unused content blocking files and replaces them with the in-house built TDS.  Includes an embedded version which it will fallback on if there are any problems (e.g file corruption or incompatible format).

**Steps to test this PR**:

***Tracker blocking:***
1. Check trackers are blocked on all the usual suspects (cnn.com, etc)
1. Turn off site protection and check trackers are found.

***First Party:***
1. Go to Reddit.com and ensure it loads - this site breaks if we block first party trackers.

***Temporary whitelisting:***
1. Check that sites on the temporary whitelist do not block trackers.

***Embedded TDS and Downloads:***
1. The embedded version must be valid.  Without valid JSON for the TDS page breakage is likely to occur, so don't change the embedded one for any tests.
1. Prevent a new version from being downloaded and make it invalid (e.g. using a Charles Proxy Breakpoint).  The app should fall back to using the embedded version.

***Migration:***
1. Start from develop on a simulator, load the app and browse around.  Take note of the app's location in the console.
1. Install the new app.  Check the files folder and ensure that no longer relevant files are removed, including the string-cache folder.
1. The Network Leaderboard should be reset after upgrading to this version.

***D enhanced to D bug:***
1. Visit https://newatlas.com/science/psilocybin-major-depression-mdd-usona-breakthrough-therapy-fda/
1. The score will eventually change to D (previously it would say enhanced from D to D)

---
###### Internal references:


Core/StorageCache.swift
```diff
@@ -91,29 +57,22 @@ public class StorageCache: StorageCacheUpdating {
...
-            let result = entityMappingStore.persist(data: data)
-            entityMapping = EntityMapping(store: entityMappingStore)
-            return result
+        case .surrogates:
+            return fileStore.persist(data as? Data, forConfiguration: configuration)
             
-        default:
+        case .trackerDataSet:
+            if fileStore.persist(data as? Data, forConfiguration: configuration) {
+                TrackerDataManager.shared.reload()
```
bwaresiak: Basically if we are reloading here, then we should not be left with embedded data set - I'd make it so `reload()` method returns what kind of data it uses (fetched/embedded) and fire a pixel if we do not manage to load what we have just persisted.

Core/TrackerDataManager.swift
```diff
@@ -0,0 +1,95 @@
...
+        let data = FileStore().loadAsData(forConfiguration: .trackerDataSet) ?? Self.loadEmbeddedAsData()
+        do {
+            self.trackerData = try JSONDecoder().decode(TrackerData.self, from: data)
+        } catch {
+            Logger.log(text: "error loading downloaded tds [\(error.localizedDescription)], falling back to embedded")
+            
+            // The embedded data should NEVER fail, so fall back to it
+            let embeddedData = try? JSONDecoder().decode(TrackerData.self, from: Self.loadEmbeddedAsData())
+            self.trackerData = embeddedData!
+            Pixel.fire(pixel: .reloadTrackerDataFailed)
```
bwaresiak: You can use Pixel's `fire(pixel: PixelName, error: Error) ` method to embed error info as parameters.

Core/TrackerDataManager.swift
```diff
@@ -0,0 +1,95 @@
...
+    
+    private(set) public var trackerData: TrackerData!
+    
+    init() {
+        reload()
+    }
+    
+    public func reload() {
+        // valid use case is we don't have downloaded data yet
+        let data = FileStore().loadAsData(forConfiguration: .trackerDataSet) ?? Self.loadEmbeddedAsData()
```
bwaresiak: I'd consider making this flow a bit more explicit:
- Path for `trackerDataSet` only.
- Path for fallback/error handling. 

Above line sort of mixes both and we will miss any issues around storing/reading data set file (instead we will quietly fallback to embedded data set).

Core/TabInstrumentation.swift
```diff
@@ -97,11 +97,19 @@ public class TabInstrumentation {
         if #available(iOSApplicationExtension 12.0, *) {
             let currentURL = self.currentURL ?? "unknown"
             // 0 is treated as 1ms
-            let timeInNS: UInt64 = timeInMs > 0 ? UInt64(timeInMs * 1000 * 1000) : 1000000
+            let timeInNS: UInt64 = timeInMs.asNanos
```
bwaresiak, line 106: 👍

Core/TrackerData.swift
```diff
@@ -0,0 +1,120 @@
+//
+//  KnownTracker.swift
```
bwaresiak: `TrackerData.swift`

Core/TrackerDataManager.swift
```diff
@@ -0,0 +1,95 @@
+//
+//  TrackerData.swift
```
bwaresiak: `TrackerDataManager.swift`

subsymbolic:
One more test worth running is to make sure the leaderboard doesn't start adding duplicates for migrated users as the ids have changed (they need a migration), see the updated names here https://github.com/duckduckgo/duckduckgo-privacy-extension/blob/develop/shared/js/background/migrate.es6.js.

brindy:
@bwaresiak I've pushed some changes for your feedback, thanks. See what you think.

Thanks @subsymbolic - good shout.

bwaresiak:
I've noticed that since some strings have been changed, we started to get layout issues on smaller screens (possibly not only on the smallest ones), see first trackers group:

layout

brindy:
Pushed a fix that uses a predefined list of allowed categories and it takes the first one and uses that.  Should be very similar to before now.

bwaresiak (approved):
Great work @brindy :)

brindy:
Thanks @bwaresiak !

bwaresiak (approved):
LGTM :)

PR #505, opened 2019-10-21
Reveal bars when scrolling below bottom of the web page

yonihemi:
Task/Issue URL: https://github.com/duckduckgo/iOS/issues/495
CC: @bwaresiak

**Description**:
Changed behavior of BarsAnimator to reveal navigation bar and toolbar when the user 'rubberbands' the web page beyond its lower bound. 

**Steps to test this PR**:
1. Load a web page higher than a screen's height.
2. Scroll down and let the bars hide automatically.
3. After reaching the bottom, let go of the screen and scroll down again (this is a similar interaction required by Mobile Safari to reveal bars).
4. Top and bottom bars should appear automatically.

---
###### Internal references:
[Software Engineering Expectations](https://app.asana.com/0/59792373528535/199064865822552)
[Technical Design Template](https://app.asana.com/0/59792373528535/184709971311943)

Issue #495, opened 2019-10-15
Reveal bars when scrolling below bottom of the web page


bwaresiak:
Internal reference: https://app.asana.com/0/856498667320406/1145004832722770

**Goal**

Duplicate Safari's behavior: Safari shows the navigation bars when you reach the bottom of a web page and try to scroll down more.

PR #544, opened 2019-12-11
Reveal bars when scrolling below bottom of the web page

ealeksandrov:
Task/Issue URL: https://github.com/duckduckgo/iOS/issues/495
CC: @bwaresiak

**Description**:

Replicate Safari behavior - show the navigation bars when you reach the bottom of a web page and try to scroll down more.

**Technical notes**:

`contentOffsetYAtBottom` change:
This issue surfaced when I started testing bottom drag detection. UIScrollView safe area handling depends on `contentInsetAdjustmentBehavior`. To keep correct behaviour in all cases I propose changing `contentInset + safeAreaInsets` to `adjustedContentInset`.
This change contained in separate commit, feel free to cherry-pick it from PR or squash.

Bottom scroll handling:
Bars will appear only if scroll started from the very end of web page.

**Steps to test this PR**:

0. For every scenario - load webpage bigger than screen height.

##### Basic scenario
Steps:
1. Hide bars and scroll to the very bottom of the web page.
2. Scroll webpage down again (with rubber band effect).

Expectation: bars are revealed.

##### First scenario
(from https://github.com/duckduckgo/iOS/pull/505#pullrequestreview-307847045 comment)
In Safari, user is able to reach bottom of the web page and not reveal bars when actively scrolling (not decelerating).

Steps:
1. Hide bars and scroll just above bottom of the web page.
2. Start dragging the webpage towards the bottom without lifting the finger.

Expectation: bars are not revealed.

##### Second scenario
(from https://github.com/duckduckgo/iOS/pull/505#pullrequestreview-307847045 comment)
In Safari, bars will only appear when user scrolls downwards from the very bottom (end) of the webpage.

Steps:
1. Hide bars and scroll to the very end of the page in a way that bars remain hidden.
2. Start dragging the webpage a bit towards the top, then move back to the bottom.

Expectation: bars are not revealed.

##### Third scenario
In Safari, bars will not disappear when user scrolls downwards just above bottom of the web page (threshold estimated to 110, see `Constants.bottomThreshold`).

Steps:
1. Let bars appear at the bottom of webpage (*basic* scenario above)
2. Scroll up 2-3 lines
3. Start scrolling down.

Expectation: bars are revealed after 1st step and remain in place on steps 2 and 3. 

2. Scroll up >5 lines
3. Start scrolling down.

Expectation: bars are revealed after 1st step and hidden on step 3. 

---
###### Internal references:

### DIFF
--- a/DuckDuckGo/BrowserChromeManager.swift
+++ b/DuckDuckGo/BrowserChromeManager.swift
@@ -85,9 +85,6 @@ class BrowserChromeManager: NSObject, UIScrollViewDelegate {
             }
             return
         }
-        
-        let isInBottomBounceArea = scrollView.contentOffset.y > scrollView.contentOffsetYAtBottom
-        guard isInBottomBounceArea == false else { return }
 
         animator.didScroll(in: scrollView)
     }
@@ -118,7 +115,7 @@ class BrowserChromeManager: NSObject, UIScrollViewDelegate {
         guard !scrollView.isZooming else { return }
         guard canHideBars(for: scrollView) else { return }
         
-        animator.didFinishScrolling(velocity: velocity.y)
+        animator.didFinishScrolling(in: scrollView, velocity: velocity.y)
     }
 
     func scrollViewDidEndDragging(_ scrollView: UIScrollView, willDecelerate decelerate: Bool) {
@@ -155,11 +152,23 @@ private class BarsAnimator {
     var draggingStartPosY: CGFloat = 0
     
     var transitionStartPosY: CGFloat = 0
+
+    private var bottomRevealGestureState: BottomBounceRevealing = .possible
+    private var combinedBarsHeight: CGFloat {
+        guard let delegate = delegate else { return 0 }
+        return delegate.toolbarHeight + delegate.omniBar.frame.height
+    }
     
     enum State: String {
         case revealed
         case transitioning
         case hidden
+    }
+    
+    enum BottomBounceRevealing: String {
+        case possible
+        case triggered
+        case cancelled
     }
     
     func didStartScrolling(in scrollView: UIScrollView) {
@@ -193,6 +202,8 @@ private class BarsAnimator {
     
     private func revealedAndScrolling(in scrollView: UIScrollView) {
         guard scrollView.contentOffset.y > draggingStartPosY else { return }
+        guard scrollView.contentOffset.y < scrollView.contentOffsetYAtBottom - combinedBarsHeight else { return }
+        guard bottomRevealGestureState != .triggered else { return }
         
         // In case view has been "caught" in the middle of the animation above the (0.0, 0.0) offset,
         //wait till user scrolls to the top before animating any transition.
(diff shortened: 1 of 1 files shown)


DuckDuckGo/BrowserChromeManager.swift
```diff
@@ -147,6 +144,10 @@ class BrowserChromeManager: NSObject, UIScrollViewDelegate {
 
 private class BarsAnimator {
     
+    struct Constants {
+        static let bottomThreshold: CGFloat = 110
```
bwaresiak: I wonder, where does this number come from?

DuckDuckGo/BrowserChromeManager.swift
```diff
@@ -156,6 +157,9 @@ private class BarsAnimator {
     
     var transitionStartPosY: CGFloat = 0
     
+    private var scrollFromBottomTriggered = false
+    private var scrollFromBottomInvalidated = false
```
bwaresiak: Combinations of these two allow for 4 states (false false, false true, true false, true true) but I see that one combination is unused (invalid).Maybe this could be replaced with a state represented as an Enum? 

```
enum BottomTransitioning { // BottomBounceTransitioning ? BottomBounceRevealing ?
     case possible
     case triggered
     case cancelled
}
```

`possible` - equivalent to `Triggered = false ; Invalidated = false`
`triggered` - equivalent to `Triggered = true ; Invalidated = false`
`cancelled` - equivalent to `Triggered = false ; Invalidated = true`

bwaresiak (commented):
Hello!

This seems to work well 😄 I've added some comments / questions inline.

Thanks for picking this up!

DuckDuckGo/BrowserChromeManager.swift
ealeksandrov: In "third scenario" there is a comment:

> In Safari, bars will not disappear when user scrolls downwards just above bottom of the web page (threshold estimated to 110, see Constants.bottomThreshold).

I just compared same page on Safari and DDG side by side, this value looks very close.

DuckDuckGo/BrowserChromeManager.swift
ealeksandrov: Agree, will make a change.

DuckDuckGo/BrowserChromeManager.swift
bwaresiak: Wouldn't that number be equal to the combined height of the areas covered by top & bottom bars (since hiding bars within that distance of the bottom would have to readjust content offset to accommodate for expanded viewport)? 

Do you think that makes sense? If so, we could probably get rid of the constant and use `delegate` to obtain `omniBar.bounds.height` and `toolbarHeight` to use instead.

ealeksandrov: I see the idea. But won't we also need bottom safe area which is part of bottom bar but is not part of `toolbarHeight`?

The value will be 52 + 44 + 34.

BTW, In this interaction Safari freezes bottom bar in transitioning state when "bottomThreshold" entered and animates hide on touch up. I like current (this PR) implementation better 😄

bwaresiak: Yes, if device has safe area then this value will be different - what I meant was that if possible, I'd prefer to use actual values coming from the properties of a view, rather than constants. 

And yes, implementation seems to work rather smoothly. :)

ealeksandrov: Updated it. Actually it worked well without additional safe area inset.

bwaresiak (approved):
@ealeksandrov Works great, thank you for the contribution :)

PR #503, opened 2019-10-18
Refactor same day check

leschlogl:
**Description**:
I've refactored some parts that use verification of the same day. I've put into an extension to reuse code in the entire application with simple method calling, avoiding "copying and pasting" of the same day logic. Also, I've created some unit tests for this.

**Steps to test this PR**:
1 - Use the app in the parts refactored or use the Unit Tests

PR #556, opened 2020-01-15
Added extension to check same day

leschlogl:
**Description**:
I've refactored some parts that use verification of the same day. I've put into an extension to reuse code in the entire application with simple method calling, avoiding "copying and pasting" of the same day logic. Also, I've created some unit tests for this.

**Steps to test this PR**:
1 - Use the app in the parts refactored or use the Unit Tests

### DIFF
--- a/Core/DateExtension.swift
+++ b/Core/DateExtension.swift
@@ -1,5 +1,5 @@
 //
-//  Debounce.swift
+//  DateExtension.swift
 //  DuckDuckGo
 //
 //  Copyright © 2020 DuckDuckGo. All rights reserved.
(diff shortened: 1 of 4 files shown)


leschlogl:
@bwaresiak This is the new PR based on #503 

Thanks in advance!

Core/DateExtension.swift
```diff
@@ -0,0 +1,27 @@
+//
+//  Debounce.swift
```
bwaresiak: Ah I've just noticed that - can you please make this match the filename. :) Thanks!

leschlogl: Sorry, my mistake! Solved!

bwaresiak:
@leschlogl thank you! :)

This is tracked internally at: https://app.asana.com/0/856498667320406/1157331335297497

Issue #559, opened 2020-01-19
UI/Snapshot Tests are broken


albsala:
UI/Snapshot Tests are broken in develop. Something about the Onboarding, Site Ranking, and Network Offenders must have been changed and UI Tests were not updated apparently. It's a quick fix so I'm preparing a PR in case it's useful for you.

brindy:
Hello,

We are now tracking this internally at: https://app.asana.com/0/414709148257752/1158699016458265

Since there was no further PR forthcoming I'm going to close this out.

Thanks for your continued use of our app!

PR #560, opened 2020-01-19
Fix UI/Snapshot Tests

albsala:
Task/Issue URL: #559 
Tech Design URL: N/A
CC: N/A

**Description**:
UI/Snapshot tests are broken in develop branch so I fixed them up. It was just adding a couple of accesibilityIdentifiers and a couple of lines in UITests.swift.

**Steps to test this PR**:
1. Make `~/Library/Caches/tools.fastlane/screenshots/` folder
2. Select UITests Schema and test.

---
###### Internal references:

### DIFF
--- a/DuckDuckGo/HomeRow.storyboard
+++ b/DuckDuckGo/HomeRow.storyboard
@@ -219,6 +219,7 @@
                                     </view>
                                     <button opaque="NO" clipsSubviews="YES" contentMode="scaleToFill" contentHorizontalAlignment="center" contentVerticalAlignment="center" buttonType="roundedRect" lineBreakMode="middleTruncation" translatesAutoresizingMaskIntoConstraints="NO" id="2qj-n8-fMW">
                                         <rect key="frame" x="15" y="298" width="290" height="44"/>
+                                        <accessibility key="accessibilityConfiguration" identifier="Continue"/>
                                         <constraints>
                                             <constraint firstAttribute="height" constant="44" id="Llr-og-9as"/>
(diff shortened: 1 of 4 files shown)


bwaresiak:
Hello @albsala 

Thanks for looking at this. I've checked out your branch and while it correctly goes through onboarding & add to home row dialog, it still fails for me after a while (after first set of screenshots is made).

`Failed to synthesize event: Neither element nor any descendant has keyboard focus. Event dispatch snapshot: SearchField, identifier: 'searchEntry', placeholderValue: 'Search or enter address', `

Note, that I've removed the app before running the test. Does this test pass for you?

bwaresiak:
@albsala actually, please wait a bit before spending more time on this. :)

It looks like we are not using these tests anymore, so I'm going to discuss whether we should remove these instead.

Thanks!

Internal reference: https://app.asana.com/0/856498667320406/1158699016458265

albsala:
@bwaresiak Tests were successful for me before sending the PR. Since it's a keyboard issue, it might be dependant on the simulator/settings we're using.
These tests were not been used in CI but I would suggest you to maintain them because they're pretty useful :)
Cheers

bwaresiak:
@albsala We've had a chat about that and agreed that these tests are not really used anymore. Because of that, we'd like to avoid the cost of maintaining them and would prefer to remove these.

I'm specifically talking about:
- UI/Snapshot tests.
- IntegrationTests (do not confuse with AtbIntegration tests).
- TopSitesReport.

Instead of this PR, would you be willing to create one focused on removing all of the above? Thanks!

PR #574, opened 2020-02-11
add a timed pixel to measure taken to clear data and number of tabs when using fire button

brindy:
Task/Issue URL: https://app.asana.com/0/392891325557410/1161522674005625
Tech Design URL:
CC:

**Description**:

Allow us to gather metrics on how long it takes to clear data and how many tabs users have when they use the fire button.

**Steps to test this PR**:
1. Open some tabs, browse and use the fire button
1. Use the fire button with no tabs
1. Check kibana for the pixels. e.g. `pixel:/mf.dc.*ios.*/ AND request:/.*test=1.*/`

---
###### Internal references:

### DIFF
--- a/Core/Pixel.swift
+++ b/Core/Pixel.swift
@@ -28,7 +28,8 @@ public enum PixelName: String {
     case forgetAllPressedBrowsing = "mf_bp"
     case forgetAllPressedTabSwitching = "mf_tp"
     case forgetAllExecuted = "mf"
-    
+    case forgetAllDataCleared = "mf_dc"
+
     case privacyDashboardOpened = "mp"
     case privacyDashboardScorecard = "mp_c"
     case privacyDashboardEncryption = "mp_e"
@@ -230,7 +231,8 @@ public struct PixelParameters {
     static let errorCount = "c"
     static let underlyingErrorCode = "ue"
     static let underlyingErrorDesc = "ud"
-    
+
+    public static let tabCount = "tc"
 }
 
 public struct PixelValues {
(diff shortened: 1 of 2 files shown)


Core/Pixel.swift
```diff
@@ -28,7 +28,8 @@ public enum PixelName: String {
     case forgetAllPressedBrowsing = "mf_bp"
     case forgetAllPressedTabSwitching = "mf_tp"
     case forgetAllExecuted = "mf"
-    
+    case forgetAllDataCleared = "mf_dc"
```
bwaresiak, line 31: I just wonder, is it a regular, or rather a debug pixel?

Core/Pixel.swift
```diff
@@ -305,9 +307,11 @@ public class TimedPixel {
         self.date = date
     }
     
-    public func fire(_ fireDate: Date = Date()) {
+    public func fire(_ fireDate: Date = Date(), withAdditionalParmaeters params: [String: String?] = [:]) {
         let duration = String(fireDate.timeIntervalSince(date))
-        Pixel.fire(pixel: pixel, withAdditionalParameters: [PixelParameters.duration: duration])
+        var newParams = params
+        newParams[PixelParameters.duration] = duration
```
bwaresiak, line 313: Maybe we should ensure that `params` dict already does not contain `duration` key before assigning it (to avoid stupid mistakes :) )?

bwaresiak (approved):
LGTM! Two optional comments inline.

Core/Pixel.swift
brindy: Yep, I had the same thought but settled for regular because it's a little bit more than debug due to having the tab count.

Core/Pixel.swift
brindy: I think it would be unlucky to do it by accident and **should** be picked up when testing anyway. :)

brindy:
Thanks @bwaresiak !

PR #581, opened 2020-02-17
Pick App Icon during onboarding

bwaresiak:
Task/Issue URL: https://app.asana.com/0/856498667320406/1161590609052529
Tech Design URL:
CC:

**Description**:
Onboarding experiment: Added screen to present user an option to choose App Icon during onboarding.

Note: all non-supported devices (e.g. iOS <10.3) are removed from both experiment and control group pool.

**Steps to test this PR**:

Control:
1. Run onboarding.
2. Only welcome and home row screens should be shown.
3. Check if App Icon picker in settings works as expected.

Experiment:
1. Run onboarding.
2. App Icon picker should be presented as a 2nd screen.
3. Check if App Icon picker in settings works as expected.

Check iPhone SE / 11 / 11 Pro Max & iPad layout.

Non supported devices:
1. Change `isAppIconChangeSupported` so it return false.
2. Reinstall the app - no variant should be allocated.

---
###### Internal references:


DuckDuckGo/AppIconSettingsViewController.swift
```diff
@@ -79,14 +73,48 @@ class AppIconSettingsViewController: UICollectionViewController {
...
     }
+}
 
-    // MARK: UICollectionViewDelegate
+class AppIconWorker {
+    
+    enum Context: String {
+        case onboarding
+        case settings
+    }
```
brindy, line 83: Optional: Typically we'd abbreviate these parameter values, e.g. to `o` and `s` in this case.

brindy (approved):
LGTM! One optional suggestion, but feel free to merge once you get your final copy if the UI doesn't change drastically.

bwaresiak:
@brindy Thank you!

tomasstrba:
LGTM! good job! Can't wait for results :)

I have just one point for discussion, no need to do any changes. What about joining classes AppIconWorker and AppIconManager together? AppIconManager could use context as input parameter when changing the icon and also fire the pixel after successful operation. What do you think?

tomasstrba (approved):

bwaresiak:
@tomasstrba That's a good idea, `AppIconWorker` and `AppIconManager` could be merged. What I wonder about now tho, is whether we need the pixels for that at all - once the experiment is over? I'd keep it as it is for now (to not invalidate testing) but come back to this after the experiment.

tomasstrba:
@bartek, I agree with you. We can merge them, but firing a pixel should be outside the class. 👍🏼☺️

PR #595, opened 2020-03-12
Improved Broken sites process

bwaresiak:
Task/Issue URL: https://app.asana.com/0/856498667320406/1154001374942564
Tech Design URL: https://app.asana.com/0/72649045549333/1163624025952315
CC:

**Description**:
Update UI to support picking up category. Send a pixel with the broken site data as per the extensions.

**Steps to test this PR**:

Broken site form:
- Submit button should be disabled on start.
- Category entries should be randomized.
- Dark/Light theme should be supported.
- On iPhone: navigating from Privacy Screen should open modal view stacked with Privacy Screen.
- On iPhone: navigating from Menu should open full screen modal view.
- On iPad: both navigation paths should open modal view (non-full screen one).

Sending report should include:
- TDS etag (even for embedded TDS if falling back to it).
- Loaded surrogates.
- Domains of blocked trackers.
- Mobile/Desktop and HTTPS upgrade status.

---
###### Internal references:


Core/contentblocker.js
```diff
@@ -508,6 +508,7 @@ _utf8_encode : function (string) {
 	        url: trackerUrl,
 	        blocked: blocked,
 	        reason: result.reason,
+                                            isSurrogate: result.matchedRule.surrogate != null
```
brindy: Spacing is off here

DuckDuckGoTests/TrackerDataManagerTests.swift
```diff
@@ -100,10 +101,79 @@ class TrackerDataManagerTests: XCTestCase {
...
+        XCTAssertEqual(TrackerDataManager.shared.etag, TrackerDataManager.Constants.embeddedDataSetETag)
+        XCTAssertEqual(TrackerDataManager.shared.reload(etag: "new etag"), .downloaded)
+        XCTAssertEqual(TrackerDataManager.shared.etag, "new etag")
         XCTAssertNil(TrackerDataManager.shared.findEntity(byName: "Google LLC"))
         XCTAssertNotNil(TrackerDataManager.shared.findEntity(byName: "Not Real"))
 
     }
     // swiftlint:enable function_body_length
+        
+    func testWhenEmbeddedDataIsUpdatedThenUpdateSHAAndEtag() {
```
brindy, line 113: Nice! 🕺

brindy (changes requested):
Code is looking good, nice work.  A few things though:

1) If we're maintaining parity with the extensions then Other (Something else) should always be the last item in the list of reasons.

2) siteUrl should be encoded (this is a bug in the Safari code).  The backend seems to handle it but encoding it is still the correct thing to do.

3) if device model is important then we should update Device.swift.  It's not been updated for a while and I've seen "unknown" appear for the model in some cases. Sorry 🤕 - but let me know if I can help (it involves pulling upstream in to our fork and updating Carthage dependencies)

Core/contentblocker.js
bwaresiak: Indentation in that file is messed up (tabs mixed with spaces) I'm going to update it to use spaces instead, ok?

DuckDuckGo/BrokenSiteCategoryCell.swift
```diff
@@ -0,0 +1,26 @@
+//
```
brindy: Sorry I didn't spot this earlier, but this seems redundant.  Can we just use a basic cell since it's just a label?

brindy (changes requested):
Just a heads up that the design is showing SF Pro Text for the font rather than Proxima Nova - I've added a note for Robert as I presume that's wrong.

DuckDuckGo/BrokenSiteCategoryCell.swift
bwaresiak: Sure, my bad! :)

brindy (approved):
LGTM :)

bwaresiak:
Thanks @brindy !

PR #600, opened 2020-03-20
Update .gitignore

daisuke-t-jp:
Hi,

`.DS_Store` is automatically created on macOS.
I updated `.gitignore` for exclude  it.

Thanks.


brindy:
Thanks - I've never noticed .DS_Store being added to the repo personally, but this looks harmless enough so will merge it in case anyone else is having problems with it.

PR #608, opened 2020-03-28
Fix FindInPage

### LINKED ISSUE #605: The input suggestion function does not work with FindInPage.
# Summary
The input suggestion function does not work with FindInPage.
Therefore, the user cannot select Japanese Kanji.
I think the same problem occurs in Chinese.

# Information
- App version
  - 0894e4103b511c789accd09a28e4e78ba94f1cd1
- iOS version
  -  iOS 13.3
- Steps to reproduce the bug
  1. Open FindInPage.
  1. Input japanese.
- Expected behavior
  - Input suggestions are displayed on keyboard.
- Actual behavior
  - No suggestions are displayed on keyboard.

# Screenshots
The Input suggestions function working in NavigationBar.
navi

But it doesn't work on FindInPage.
findinpage

daisuke-t-jp:
Hi,

I fixed #605 

User can convert while typing in Japanese.
But, the text was forced into the textfield and could not be converted.

Thanks.

### DIFF
--- a/DuckDuckGo/FindInPage.swift
+++ b/DuckDuckGo/FindInPage.swift
@@ -54,10 +54,12 @@ class FindInPage: NSObject {
         webView.evaluateJavaScript("window.__firefox__.findPrevious()")
     }
 
-    func search(forText text: String) {
-        guard text != searchTerm else { return }
+    func search(forText text: String) -> Bool {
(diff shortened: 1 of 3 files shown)


brindy:
Thanks @daisuke-t-jp ! Looks good so have merged.

PR #653, opened 2020-06-10
Update hiring section of readme

catehstn:
Hiring section of README was out of date, referencing a job no longer on the site. Text from Zbig, more generic so as not to require regular updates.

Android change is here: https://github.com/duckduckgo/Android/pull/854


brindy (approved):
LGTM, thanks @catehstn

PR #654, opened 2020-06-12
if snapshot fails during onboarding just delegate to the actual icons

brindy:
Task/Issue URL: n/a
Tech Design URL: n/a
CC:

**Description**:

If snapshot fails during onboarding it can crash, rather just delegate to the actual icons.  Animations won't work as expected, but it looks ok and won't crash.

**Steps to test this PR**:
1.  Test onboarding with thermal state set to critical - still may not cause snapshots to update
1. Hack code to remove snapshot function to see effect without snapshots

---
###### Internal references:


bwaresiak (approved):
LGTM!

PR #655, opened 2020-06-18
Update .gitignore

catehstn:
Updates .gitignore to exclude font files to ensure they don't get checked in. Let me know if I'm missing something! Seems like the easiest way to ensure they don't accidentally end up in the repo.


catehstn:
Missed that there were empty copies in the folder, so this will not work.

PR #656, opened 2020-06-22
Update user agent to fix breakages

subsymbolic:
Task/Issue URL: https://app.asana.com/0/361428290920652/1176477052804240
Tech Design URL: https://app.asana.com/0/361428290920652/1176477052804235

**Description**:
Updates user agent to include a safari component after our application component (similar to Chrome) and to allow site specific changes.

**Steps to test this PR**:
1. Got to `https://www.whatismybrowser.com/detect/what-is-my-user-agent` and check that the user agent reported contains our DuckDuckGo component followed by the Safari component
`Mozilla/5.0 (iPhone; CPU iPhone OS 12_4 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Mobile/15E148 DuckDuckGo/7 Safari/605.1.15`
1. From an iPhone: Visit `tdbank.com` and press "Login in" and then "Online Banking" and check that you see a login box
1. From an iPhone: Visit `svb.com`, press Login (or the little profile icon) and then "SVB Online Service" and check that you see a login box
1. From an iPad: Visit `chase.com`. Depending on screen size you may need to press the "Sign in" button to see the login box or it may already be there. Either way make sure you can get to a login box. Note: This page is really slow sometimes!
1. From a iPhone or iPad: Visit `cvs.com` (use a VPN located in the US) in both Desktop and Mobile modes and ensure it loads ok. This UA is dynamically changed and will only work for iOS >=12

comparison1
comparison2

**Device Testing**:
* [x] iPhone
* [x] iPad

**OS Testing**:
* [x] iOS 10
* [x] iOS 11
* [x] iOS 12
* [x] iOS 13

---
###### Internal references:
[Software Engineering Expectations](https://app.asana.com/0/59792373528535/199064865822552)
[Technical Design Template](https://app.asana.com/0/59792373528535/184709971311943)

PR #660, opened 2020-06-28
Fix crash in UserAgentManager

bwaresiak:
Task/Issue URL: https://app.asana.com/0/856498667320406/1182309843403154
Tech Design URL:
CC:

**Description**:
Do not stall initialization, instead return immediately and update internal state as needed.

**Steps to test this PR**:

Same as in https://github.com/duckduckgo/iOS/pull/656

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [x] iPhone SE (1st Gen)
* [x] iPhone 8
* [x] iPhone X
* [x] iPad

**OS Testing**:

* [x] iOS 10
* [x] iOS 11
* [x] iOS 12
* [x] iOS 13

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:

### DIFF
--- a/Core/UserAgentManager.swift
+++ b/Core/UserAgentManager.swift
@@ -27,20 +27,20 @@ public class UserAgentManager {
     private var userAgent = UserAgent()
     
     init() {
-        DispatchQueue.main.async {
-            self.prepareUserAgent()
-        }
+        prepareUserAgent()
     }
     
     private func prepareUserAgent() {
         let webview = WKWebView()
         webview.load(URLRequest(url: URL(string: "about:blank")!))
         
-        guard let defaultAgent = UserAgentManager.getDefaultAgent(webView: webview) else {
-            return
+        getDefaultAgent(webView: webview) { [weak self] agent in
+            // Reference webview instance to keep it in scope and allow UA to be returned
+            _ = webview
+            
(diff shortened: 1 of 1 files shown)


Core/UserAgentManager.swift
```diff
@@ -23,16 +23,20 @@ import WebKit
...
     
     public static let shared = UserAgentManager()
-    
-    private var defaultAgentRetreived = false
-    private var userAgent: UserAgent
+
+    private var userAgent = UserAgent()
     
     init() {
+        DispatchQueue.main.async {
```
subsymbolic: This leads to a deadlock that then freezes browsing until the timeout of 2 seconds. I like the approach overall though and since we're using the fallback UA until we can get the real one, we could make the js code async. I have hacked out an example here https://github.com/duckduckgo/iOS/compare/develop...feature/mia/user-agent-crash-fix. What do you think?

bwaresiak: Oh yes, just using completion handler is much better 🤦‍♂️ 

I've tweaked my PR (Github has some issues - had to change the base branch in order to make it refresh this...) to match your idea.

Thanks! :)

subsymbolic (approved):
Great work!

PR #271, opened 2018-02-08
remove dependency on SwiftyJSON and use duckduckgo carthage dependencies

brindy:
Reviewer: Caine or Mia
Asana: 

* https://app.asana.com/0/392891325557410/534186388081450
* https://app.asana.com/0/392891325557410/534415163341060

CC:

**Description**:

Low hanging fruit:

* Remove dependency on SwiftyJSON and use provided JSON serialisation API.  
* Update dependencies to use the ones forked on DDG (instead of hosted on Mia's Github account).

**Steps to test this PR**:

1. Clean launch the app
1. Visit cnn.com - open the privacy grade and check the trackers are grouped by network as normal
1. Exploratory testing as usual

PR #677, opened 2020-07-18
Removes a comment referring to a now removed dependency

BasThomas:
The dependency referenced, SwiftyJSON, was removed in https://github.com/duckduckgo/iOS/pull/271

**Description**:
Removes an outdated comment in the Cartfile which could be confusing.


brindy:
Thanks for this @BasThomas

Issue #676, opened 2020-07-18
VoiceOver improvements


BasThomas:
While using the app I found some places where VoiceOver can either use some more information, or where a page isn't accessible at all using VoiceOver.

- Changing the app icon is currently not accessible through VoiceOver
- The "Add to Dock" modal has an awkward naming for its close button ("remove" instead of "close".
- Consider adding an accessibility hint to the "tab toggle" to announce which option is currently selected — or add to the label, as hints can be disabled.

brindy:
These are merged, thanks @BasThomas

BasThomas:
The tab toggle accessibility hint hasn't been added yet, actually. Up to you if that's something you'd want to add, though :)

brindy:
Ah, sorry - thanks!

brindy:
I've added the accessibility hint for the toggle on my current branch which is WIP but hopefully going out soon.  Thanks again for the heads up and contributions.

PR #678, opened 2020-07-18
Make alternate app icon buttons accessible

BasThomas:
References #676

Screenshot 2020-07-18 at 13 34 33
Screenshot 2020-07-18 at 13 34 40

**Description**:
Whereas previously the alternate app icon screen's buttons weren't accessible, this makes sure the buttons are seen as accessibility elements, adds the expected traits, and adds an accessibility label.

**Steps to test this PR**:
1. In VoiceOver, go to Settings > App Icon
1. Verify that the buttons are now accessible, show correct traits (button, button + selected if selected)

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [x] iPad

**OS Testing**:

* [ ] iOS 10
* [ ] iOS 11
* [ ] iOS 12
* [x] iOS 13

### DIFF
--- a/DuckDuckGo/AppIcon.swift
+++ b/DuckDuckGo/AppIcon.swift
@@ -27,6 +27,11 @@ enum AppIcon: String, CaseIterable {
     case purple
     case black
 
+    /// Returns a user facing string representation of the app icon.
+    var name: String {
+        rawValue
+    }
+
     static var defaultAppIcon: AppIcon {
         return .red
     }
--- a/DuckDuckGo/AppIconSettingsCell.swift
+++ b/DuckDuckGo/AppIconSettingsCell.swift
@@ -23,7 +23,19 @@ class AppIconSettingsCell: UICollectionViewCell {
 
     static let reuseIdentifier = "AppIconSettingsCell"
 
+    var appIcon: AppIcon! {
+        didSet {
+            imageView.image = appIcon.mediumImage
+            accessibilityLabel = appIcon.name
+        }
+    }
     @IBOutlet weak var imageView: UIImageView!
+
+    required init?(coder: NSCoder) {
+        super.init(coder: coder)
+        isAccessibilityElement = true
+        accessibilityTraits.insert(.button)
+    }
 
     override var isSelected: Bool {
         didSet {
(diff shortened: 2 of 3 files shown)


DuckDuckGo/AppIcon.swift
```diff
@@ -27,6 +27,11 @@ enum AppIcon: String, CaseIterable {
     case purple
     case black
 
+    /// Returns a user facing string representation of the app icon.
+    var name: String {
```
BasThomas, line 31: Added this "shadow" of `rawValue` to make it more explicit that this is a user facing representation. It could be considered to be changed to something like `"\(rawValue) app icon"` in the future, as the color itself doesn't necessarily fully describe the element; although I left this out for now as the context should be enough.

DuckDuckGo/AppIconSettingsViewController.swift
```diff
@@ -69,7 +69,7 @@ class AppIconDataSource: NSObject, UICollectionViewDataSource {
         cell.decorate(with: ThemeManager.shared.currentTheme)
 
         let appIcon = appIcons[indexPath.row]
-        cell.imageView.image = appIcon.mediumImage
+        cell.appIcon = appIcon
```
BasThomas, line 72: Rather than passing in the image directly, pass in the `appIcon` so we can access its `name` to update the accessibility label.

DuckDuckGo/AppIconSettingsCell.swift
```diff
@@ -23,8 +23,20 @@ class AppIconSettingsCell: UICollectionViewCell {
...
+            imageView.image = appIcon.mediumImage
+            accessibilityLabel = appIcon.name
+        }
+    }
     @IBOutlet weak var imageView: UIImageView!
 
+    required init?(coder: NSCoder) {
+        super.init(coder: coder)
+        isAccessibilityElement = true
+        accessibilityTraits.insert(.button)
```
BasThomas, line 37: `.selected` is one we get for free. :)

brindy:
Thanks again @BasThomas :)

BasThomas:
Is this something you'd want an entry to the release notes for, or is that something you will take care of?

brindy:
We'll take care of that, thanks @BasThomas  :)

PR #682, opened 2020-07-22
ipad improvements

brindy:
Task/Issue URL: 
* Top level task & Figma link https://app.asana.com/0/392891325557410/874480958064197 
* Implementation detail https://app.asana.com/0/392891325557410/1184791777204513

**Description**:

Improves iPad experience by adding:
* tabs
* stylised 'suggestions' tray 
* stylised favicons (grey box with white text letter)
* tweaked keyboard commands (hold Cmd to see the options, except for the arrow keys which aren't shown there but should still behave in a non-surprising way)
* better pointer support for keyboard users - this may be difficult to test without the physical keyboard for iPad peripheral 

Key concepts:

* Only users in the experiment should see any noticeable change to the UI 
* Only users on an iPad can be in the experiment
* Everyone else should see the iPhone experience on both the iPhone and the iPad
* If in doubt - see what Safari does in the same situation we should be fairly close to Safari where it makes sense

Also adds the following changes:

* Improved variant filtering
* Updated dax dialog copy
* Improved accessibility options for some buttons and rationalisation of some copy across the fire buttons
* Removes some unused code (e.g. App Review prompt in feedback)

**Steps to test this PR**:
1. Migrate from existing build - should not be added to the experiment
1. Fresh install until you get the experiment variant (or hack the code)
1. Search, browse, add favorites, add bookmarks, etc
1. Play with tabs; adding, reordering and making sure they reflect changes made elsewhere (tab switcher)
1. Use the app in "slide over" mode - it should have the "small" interface (no tabs, bottom button bar)
1. Use the app in "thin" side by side mode - it should have the iPhone interface (no tabs, bottom button bar)
1. Use the app in any of the "larger" side by side modes - it should have the larger interface (tabs, no bottom button bar)
 * note: portrait orientation on the iPad maybe surprising in that more of the screen is considered 'small' - again, check vs Safari

**Orientation Testing**:

* [x] Portrait
* [x] Landscape

**Device Testing**:

* [x] iPhone SE (1st Gen)
* [x] iPhone 8
* [x] iPhone X
* [x] iPad 12.9"
* [ ] iPad mini 

**OS Testing**:

* [x] iOS 10
* [x] iOS 11
* [x] iOS 12
* [x] iOS 13

**Theme Testing**:

* [x] Light theme
* [x] Dark theme

---
###### Internal references:


Core/UIViewControllerExtension.swift
```diff
@@ -25,15 +25,23 @@ extension UIViewController {
     var isSmall: Bool {
         return view.frame.height <= 568
     }
+    
+    var isPad: Bool {
```
brindy, line 29: previous global variable was not flexible enough when dealing with different size classes on iPad, so that has been removed and local isPad that meets the needs of the local usage more appropriately have been added.

DuckDuckGo/AutocompleteViewController.swift
```diff
@@ -195,7 +192,7 @@ extension AutocompleteViewController: UITableViewDataSource {
 
     func tableView(_ tableView: UITableView, numberOfRowsInSection section: Int) -> Int {
         if suggestions.isEmpty {
-            return minItems
+            return firstResponse ? 0 : minItems
```
brindy, line 195: Don't show the initial "no results" message until we've already shown something

DuckDuckGo/Feedback.storyboard
```diff
@@ -188,57 +188,20 @@
                                                         </attributedString>
                                                         <nil key="highlightedColor"/>
                                                     </label>
-                                                    <label opaque="NO" userInteractionEnabled="NO" contentMode="left" horizontalHuggingPriority="251" verticalHuggingPriority="251" usesAttributedText="YES" lineBreakMode="tailTruncation" numberOfLines="0" baselineAdjustment="alignBaselines" adjustsFontSizeToFit="NO" translatesAutoresizingMaskIntoConstraints="NO" id="YTz-AA-F85">
```
brindy, line 191: We hide this to get through App Store approval so just cleaning up redundant code.

DuckDuckGo/MainViewController+KeyCommands.swift
```diff
@@ -70,7 +70,15 @@ extension MainViewController {
             ]
         }
 
-        return alwaysAvailable + browsingCommands + findInPageCommands + [
+        var arrowKeys = [UIKeyCommand]()
+        if omniBar.textField.isFirstResponder {
```
brindy, line 74: If the arrow keys are always associated with a command it stops text editing in the web view - if the omni bar text field is the first responder then the user can't be editing and is likely to doing a search or something.

DuckDuckGo/PrivacyProtectionController.swift
```diff
@@ -47,7 +47,12 @@ class PrivacyProtectionController: ThemableNavigationController {
             overrideUserInterfaceStyle = .light
         }
 
-        navigationBar.isHidden = isPad
+        if DefaultVariantManager().isSupported(feature: .iPadImprovements) {
```
brindy, line 50: if in experiment let new logic determine if the navigation bar should be hidden, otherwise use the previous logic (which was wrong in some circumstances, but 🤷 )

OpenAction/ActionViewController.swift
```diff
@@ -19,41 +19,54 @@
 
 import UIKit
 import MobileCoreServices
+import Core
+import os
 
 class ActionViewController: UIViewController {
```
brindy, line 25: Add support for text and URLs, and a little clean up

adhoc-build.sh
```diff
@@ -12,8 +12,8 @@ NAME="DuckDuckGo-$SUFFIX`date "+%Y-%m-%d-%H-%M"`"
 echo Building $NAME
 echo
 
-xcodebuild -scheme DuckDuckGo clean archive -configuration release -sdk iphoneos -archivePath $1/$NAME.xcarchive
-xcodebuild -exportArchive -archivePath $1/$NAME.xcarchive -exportOptionsPlist ./adhocExportOptions.plist -exportPath $1/$NAME
+xcodebuild -allowProvisioningUpdates -scheme DuckDuckGo clean archive -configuration release -sdk iphoneos -archivePath $1/$NAME.xcarchive
```
brindy, line 15: You can add new devices to the adhoc profile and let Xcode download it from this script rather than using Xcode itself.

Core/VariantManager.swift
```diff
@@ -25,34 +25,49 @@ public enum FeatureName: String {
     // Used for unit tests
     case dummy
     
+    case iPadImprovements
     case tabSwitcherListLayout
     
 }
 
-public enum CohortFiltering {
+public enum VariantInclusion {
```
subsymbolic: Optional: The `VariantInclusion` and `Is` struct feels like a doubling up of the same concept.  Can we combine them? Then enum isn't doing much and if the `always` condition becomes a lambda/completion too, we end up with something like the following and  it simplifies logic further along.

  ```
public typealias VariantInclusion = () -> Bool

struct Inclusion {
    static let always = { return true }
    static let padDevice = { return UIDevice.current.userInterfaceIdiom == .pad }
    static let notPadDevice = { return !Inclusion.padDevice() }
}
````

File: Core/VariantManager.swift [hunk 9]
```diff
@@ -80,23 +95,30 @@ public class DefaultVariantManager: VariantManager {
...
-    private let uiIdiom: UIUserInterfaceIdiom
     
     public init(variants: [Variant] = Variant.defaultVariants,
                 storage: StatisticsStore = StatisticsUserDefaults(),
-                rng: VariantRNG = Arc4RandomUniformVariantRNG(),
-                uiIdiom: UIUserInterfaceIdiom = UI_USER_INTERFACE_IDIOM()) {
-        self.variants = variants
+                rng: VariantRNG = Arc4RandomUniformVariantRNG()) {
+        
+        self.variants = variants.filter {
```
subsymbolic: Let's randomly allocate first and THEN use the `isIncluded` logic to decide whether to keep the allocation or set it to nil. We want to throw away filtered out results rather than recycling them and changing the population distribution of other experiments.

A couple of examples of why this is important:
* if one experiment is English-only then experiments without that filter will receive more non-english devices
* if one experiment targets a higher OS level then experiments without that filter will receive a disproportionate number of users on lower OS levels than usual

Core/VariantManager.swift
```diff
@@ -80,23 +95,30 @@ public class DefaultVariantManager: VariantManager {
...
     public init(variants: [Variant] = Variant.defaultVariants,
                 storage: StatisticsStore = StatisticsUserDefaults(),
-                rng: VariantRNG = Arc4RandomUniformVariantRNG(),
-                uiIdiom: UIUserInterfaceIdiom = UI_USER_INTERFACE_IDIOM()) {
-        self.variants = variants
+                rng: VariantRNG = Arc4RandomUniformVariantRNG()) {
+        
+        self.variants = variants.filter {            
+            switch $0.isIncluded {
+            case .always:
```
subsymbolic: As per previous comment, logic here is simpler without enum as there is no need to branch just process the lambda/completion.

Core/VariantManager.swift
```diff
@@ -25,34 +25,49 @@ public enum FeatureName: String {
...
 public struct Variant {
     
+    struct Is {
+        static let padDevice = { return UIDevice.current.userInterfaceIdiom == .pad }
+        static let notPadDevice = { return !Is.padDevice() }
+    }
+    
     static let doNotAllocate = 0
     
+    // Note: Variants with `doNotAllocate` weight, should always be included so that previous installations are unaffected
```
subsymbolic, line 43: Can we add a note that all variants for an experiment must have the same filters? It would be invalid for example to have no filter on the control group and them make the experimental group English-only.

Core/VariantManager.swift
```diff
@@ -25,34 +25,49 @@ public enum FeatureName: String {
...
     public static let defaultVariants: [Variant] = [
+        
         // SERP testing
-        Variant(name: "sc", weight: 1, features: []),
-        Variant(name: "sd", weight: doNotAllocate, features: []),
-        Variant(name: "se", weight: 1, features: []),
+        Variant(name: "sc", weight: 1, features: [], isIncluded: .when(Is.notPadDevice)),
```
subsymbolic: Could we move the `isIncluded` variable to be after the weight? Seems more natural to have the weight and filtering next to each other.

Core/VariantManager.swift
brindy: Thanks, that's the bit I wasn't getting my head around.

brindy:
@subsymbolic - cleaned up the variant stuff, if you can take another peek. thanks!

subsymbolic:
@brindy that experimental code looks perfect 👍 
@tomasstrba I've finished checking the experimental section, over to you for the general review

Core/UIViewControllerExtension.swift
tomasstrba: Just checking, isn't _UIDevice.current.userInterfaceIdiom_ better iPad detection? To me, using size class to detect a device is more error prone.

[Brindy] - dunno why but I can't reply to this comment so here's my reply :) 

It used to be based on device idiom but that doesn't help on iPads when the app is in split screen modes and should be showing the phone view.  It's not ideal and once this project is released fully then I'll replace isPad with a call to the `AppWidthObserver`

DuckDuckGo/FavoriteHomeCell.swift
```diff
@@ -86,21 +85,20 @@ class FavoriteHomeCell: UICollectionViewCell {
...
-        
-        iconImage.isHidden = true
-        iconLabel.isHidden = false
+        accessibilityLabel = "\(link.displayTitle ?? "")). \(UserText.favorite)"
         
+        titleLabel.text = link.displayTitle
         iconBackground.backgroundColor = host.color
-        useImageBorder(true)
+        
+        if let domain = link.url.host?.dropPrefix(prefix: "www."),
```
tomasstrba, line 96: Optional: This is a good candidate for the _domain_ variable defined in URL extension.

DuckDuckGo/GestureToolbarButton.swift
```diff
@@ -34,36 +35,45 @@ class GestureToolbarButton: UIView {
...
     
     var image: UIImage? {
         didSet {
             iconImageView.image = image
         }
     }
     
+    let pointerView: UIView = UIView(frame: CGRect(x: 0,
+                                                   y: 0,
+                                                   width: 48,
```
tomasstrba: Did you miss to define these values as constants or is it here on purpose?

DuckDuckGo/GestureToolbarButton.swift
```diff
@@ -34,36 +35,45 @@ class GestureToolbarButton: UIView {
...
+                                                   height: 36))
+    
     override init(frame: CGRect) {
         super.init(frame: frame)
 
+        addSubview(pointerView)
         addSubview(iconImageView)
-        
+
```
tomasstrba, line 69: Optional: If you would like to get rid of whitespace-only lines, you can select one magic checkbox in Xcode Preferences.
Screenshot 2020-08-13 at 11 07 34

DuckDuckGo/OmniBar.swift
```diff
@@ -33,17 +35,28 @@ class OmniBar: UIView {
...
-
-    @IBOutlet weak var separatorHeightConstraint: NSLayoutConstraint!
+ 
+    @IBOutlet weak var bookmarksButton: UIButton!
+    @IBOutlet weak var backButton: UIButton!
+    @IBOutlet weak var forwardButton: UIButton!
+    @IBOutlet weak var shareButton: UIButton!
+
+    // Don't use weak because adding/removing them causes them to go away
+    @IBOutlet var separatorHeightConstraint: NSLayoutConstraint!
```
tomasstrba, line 50: Some devs use low/high priority of the constraint instead of activation/deactivation. Then the references could stay weak. But not sure if it is possible in this particular case

DuckDuckGo/TabSwitcherButton.swift
```diff
@@ -45,9 +46,12 @@ class TabSwitcherButton: UIView {
     var workItem: DispatchWorkItem?
 
     let anim = LOTAnimationView(name: "new_tab")
-    let tint = UIView(frame: .zero)
     let label = UILabel()
-    
+    let pointerView: UIView = UIView(frame: CGRect(x: 0,
+                                                   y: 0,
+                                                   width: 48,
```
tomasstrba: Optional: Define as contants.

DuckDuckGoTests/LargeOmniBarStateTessts.swift
```diff
@@ -0,0 +1,338 @@
+//
```
tomasstrba: The filename is LargeOmniBarStateTe**ss**ts.swift. Did you use Xcode function to refactor? :P

adhoc-build.sh
tomasstrba: Nice!

tomasstrba (approved):
Good job Brindy! 💯  I've added a few comments, but most of them are optional. 

The only thing I'm afraid of are strong references to constraints. Did you try to use constraint priorities instead of activation as I described in the comment?

DuckDuckGo/FavoriteHomeCell.swift
brindy: Going to leave this as is because www is the only subdomain we can safely remove and domain could be different things to different people, like removing other subdomains, but that logic is tricky without using a lookup.

DuckDuckGo/GestureToolbarButton.swift
brindy: Will fix

DuckDuckGo/GestureToolbarButton.swift
brindy: Good tip, thanks

DuckDuckGo/OmniBar.swift
brindy: Will bear that in mind for the future, thanks. I think this is OK for now though.

DuckDuckGo/TabSwitcherButton.swift
brindy: Will fix

DuckDuckGoTests/LargeOmniBarStateTessts.swift
brindy: Thanks, will fix

Core/UIViewControllerExtension.swift
tomasstrba: Just realized it while I'm testing. Thanks for the clarification.

brindy:
The product review was approved so am merging. Thanks for the review @tomasstrba

PR #684, opened 2020-07-22
Tab Switcher layout experiment

bwaresiak:
Task/Issue URL:
 - https://app.asana.com/0/856498667320406/1185737541299226
 - https://app.asana.com/0/856498667320406/1183760799160811

Tech Design URL:
CC:

**Description**:
Experiment around default layout of the Tab Switcher.

**Steps to test this PR**:

Migration:
No effect on tab switcher is expected.

Control group:
Default to Grid on Tab Switcher.

ControlExperiment group:
Default to List on Tab Switcher.

**Copy Testing**:

**Orientation Testing**:

* [x] Portrait
* [x] Landscape

**Device Testing**:

* [x] iPhone SE (1st Gen)
* [x] iPhone 8
* [x] iPhone X
* [x] iPad

**OS Testing**:

* [x] iOS 10
* [x] iOS 11
* [x] iOS 12
* [x] iOS 13

**Theme Testing**:

* [x] Light theme
* [x] Dark theme

---
###### Internal references:

### DIFF
--- a/Core/UserDefaultsPropertyWrapper.swift
+++ b/Core/UserDefaultsPropertyWrapper.swift
@@ -56,15 +56,25 @@ public struct UserDefaultsWrapper<T> {
 
     private let key: Key
     private let defaultValue: T
+    private let setIfEmpty: Bool
 
-    public init(key: Key, defaultValue: T) {
+    public init(key: Key, defaultValue: T, setIfEmpty: Bool = false) {
         self.key = key
         self.defaultValue = defaultValue
+        self.setIfEmpty = setIfEmpty
     }
 
     public var wrappedValue: T {
         get {
-            return UserDefaults.standard.object(forKey: key.rawValue) as? T ?? defaultValue
+            if let storedValue = UserDefaults.standard.object(forKey: key.rawValue) as? T {
+                return storedValue
+            }
+            
+            if setIfEmpty {
+                UserDefaults.standard.set(defaultValue, forKey: key.rawValue)
+            }
+            
+            return defaultValue
         }
         set {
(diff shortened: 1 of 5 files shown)


Core/UserDefaultsPropertyWrapper.swift
```diff
@@ -56,15 +56,25 @@ public struct UserDefaultsWrapper<T> {
 
     private let key: Key
     private let defaultValue: T
+    private let setIfEmpty: Bool
 
-    public init(key: Key, defaultValue: T) {
+    public init(key: Key, defaultValue: T, setIfEmpty: Bool = false) {
```
bwaresiak, line 61: This is intended for all properties that are by default initialized by experiments. 

Reason: if we don't set default here, but only read it every time this is fetched, then once experiment is over we will have to either:
- Migrate all users that did not set this option themselves to new default (and possibly annoy them).
- Based on the variant populate this property to avoid breaking UX for these people.

This option solves that problem as default can be now stored in user defaults on first access.

DuckDuckGo/TabPreviewsSource.swift
```diff
@@ -27,7 +27,7 @@ class TabPreviewsSource {
     
     private var cache = [String: UIImage]()
     
-    private let tabSettings: TabSwitcherSettings = DefaultTabSwitcherSettings()
+    private lazy var tabSettings: TabSwitcherSettings = DefaultTabSwitcherSettings()
```
bwaresiak, line 30: This delays initialization of settings so variant can be allocated in time.

brindy (approved):
LGTM!

bwaresiak:
Thanks!

PR #704, opened 2020-09-07
Migrate to Swift Package Manager

THISISDINOSAUR:
Task/Issue URL: https://app.asana.com/0/1187352151074490/1187763920072074
Tech Design URL:
CC: @brindy 

**Description**:
Change dependency management from Carthage to Swift Package Manager. This includes:

- Removing Alamofire since it was no longer required, and using URLSessionDataTasks instead
- Removing Device.swift since it doesn't currently support SPM and we don't really use it
- Necessary updates to remaining dependancies to versions that support SPM. This includes API changes to Lottie and OHHTTPStubs

**Steps to test this PR**:
1. Test building the project works without Carthage, and that all tests pass
2. Test all network calls affected by the switch from Alamo fire, including manually testing the features work, and using Charles Proxy to test that the headers etc are unchanged/as expected. The features include:

* [x] favicons
* [x] autocomplete
* [x] searches 
* [x] pixels
* [x] HTTPS upgrading

**Copy Testing**:

**Orientation Testing**:

* [x] Portrait
* [x] Landscape

**Device Testing**:

* [x] iPhone SE (1st Gen)
* [x] iPhone 8
* [x] iPhone X
* [x] iPad

**OS Testing**:

* [x] iOS 11
* [x] iOS 12
* [x] iOS 13

**Theme Testing**:

* [x] Light theme
* [x] Dark theme

---
###### Internal references:


Core/APIHeaders.swift
```diff
@@ -34,9 +37,18 @@ public class APIHeaders {
     }
 
     public var defaultHeaders: HTTPHeaders {
```
THISISDINOSAUR, line 39: The default headers should be the same as they were with Alamofire

Core/APIRequest.swift
```diff
@@ -34,43 +33,73 @@ public class APIRequest {
...
+        case trace = "TRACE"
+        case patch = "PATCH"
+    }
+    
     @discardableResult
     public static func request(url: URL,
                                method: HTTPMethod = .get,
-                               parameters: [String: Any]? = nil,
-                               completion: @escaping APIRequestCompletion) -> Request {
+                               parameters: [String: String]? = nil,
```
THISISDINOSAUR, line 61: I change this to only accept strings since we were only ever using strings, and whilst Alamofire accepted other things, as far as I could tell it wouldn't successfully add any parameter that isn't a string to the URL.

Core/APIRequest.swift
```diff
@@ -34,43 +33,73 @@ public class APIRequest {
...
+            os_log("Request for %s completed with response code: %s and headers %s",
+                   log: generalLog,
+                   type: .debug,
+                   url.absoluteString,
+                   String(describing: httpResponse?.statusCode),
+                   String(describing: httpResponse?.allHeaderFields))
+            
+            if let error = error {
+                completion(nil, error)
+            } else if let error = httpResponse?.validateStatusCode(statusCode: 200..<300) {
```
THISISDINOSAUR, line 86: An alternative to this, we could potentially do the status code validation on the appropriate URLSessionDelegate callback, but after discussions with Brindy we decided it's unnecessary.
I made this default to returning the dataTask callback error over the status code error, as I figured it was likely to be more useful than a generic HTTPURLResponseError.invalidStatusCode error, but don't have strong options about it. (I think at any rate it's unlikely that they'll be no error at this point but an invalid status code)

Core/URLExtension.swift
```diff
@@ -174,3 +174,15 @@ extension URL {
...
     }
 }
+
+extension URLComponents {
+    
+    enum URLComponentsError: Error {
+        case invalidURL(url: String?)
+    }
+
+    public func asURL() throws -> URL {
```
brindy: Can we just use the `url` property directly? Then we can just dump this extension.

Core/HTTPSUpgrade.swift
```diff
@@ -95,17 +94,27 @@ public class HTTPSUpgrade {
         serviceRequest.allHTTPHeaderFields = APIHeaders().defaultHeaders
         serviceRequest.timeoutInterval = 10
 
+        let session = URLSession.shared
         let cachedResponse = URLCache.shared.cachedResponse(for: serviceRequest)?.response as? HTTPURLResponse
-        Alamofire.request(serviceRequest).validate(statusCode: 200..<300).response { response in
-            if let data = response.data {
+
+        let task = session.dataTask(with: serviceRequest) {
```
brindy: Would it be possible (and make sense) to re-use the `APIRequest `?

Core/Pixel.swift
```diff
@@ -272,10 +271,20 @@ public class Pixel {
             .pixelUrl(forPixelNamed: pixel.rawValue, formFactor: formFactor)
             .addParams(newParams)
         
-        Alamofire.request(url, headers: headers).validate(statusCode: 200..<300).response { response in
+        let session = URLSession.shared
+        var urlRequest = URLRequest(url: url)
+        urlRequest.allHTTPHeaderFields = headers
+
+        let task = session.dataTask(with: urlRequest) {
```
brindy: Again, would it be possible (and make sense) to re-use `APIRequest`?

brindy (changes requested):
As discussed, great job!  3 comments below, and we need to be careful about handling network responses on the correct thread, which Alamofire seemed to be doing for us.

Core/Pixel.swift
```diff
@@ -257,7 +256,7 @@ public class Pixel {
     
     public static func fire(pixel: PixelName,
                             forDeviceType deviceType: UIUserInterfaceIdiom = UIDevice.current.userInterfaceIdiom,
-                            withAdditionalParameters params: [String: String?] = [:],
+                            withAdditionalParameters params: [String: String] = [:],
```
THISISDINOSAUR, line 259: I changed this interface slightly so it matches APIRequest.request() and since we weren't ever using optionals here

brindy:
@bwaresiak do you have time to give this a quick look and a smoke test, please?  It's looking good to me, but I just want your eagle eyes on it :) Thanks!

brindy (approved):
Great work @THISISDINOSAUR :) LGTM

PR #706, opened 2020-09-11
iOS 14 widgets and default browser

brindy:
Task/Issue URL: 
* https://app.asana.com/0/414235014887631/1181475254517305
* https://app.asana.com/0/414235014887631/1189028203551370
Tech Design URL:
CC:

**Description**:

Adds default browser support and widgets for iOS 14

**Steps to test this PR**:

Using Xcode 12 and an iOS 14 beta simulator or device.  However, note that the profiles are required in order run default browser which means we have to use managed code signing (for now).  As such previously unregistered devices may not work.

Note: Integration tests disabled on BitRise for now, as simulator is not showing software keyboard.

***Default Browser***

1. Install the app, go to iOS Settings app > DuckDuckGo > Default Browser and choose DuckDuckGo
1. Launch URLs from other apps (e.g. calendar)
1. May get prompted to open in DuckDuckGo, allow this.

***Widgets***

1. Add each of the home screen widgets (long press home screen, hit the plus button)
1. Launch the app and add some favorites, ensure they show in the widgets
1. Edit the favorites, ensure the changes are reflected in the widgets
1. The small widget will launch a new search
1. The medium and large widget will launch whatever favorites you tap on, otherwise will launch a new search

***Regression***

1. Running on iOS 13 or earlier should show no difference in behaviour, but default browser and home screen widgets won't be available (obviously).

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [x] iPhone 8
* [x] iPhone X
* [x] iPad

**OS Testing**:

* [ ] iOS 11
* [ ] iOS 12
* [ ] iOS 13
* [x] iOS 14

**Theme Testing**:

* [x] Light theme
* [x] Dark theme

---
###### Internal references:


Widgets/Widgets.swift
```diff
@@ -0,0 +1,154 @@
...
+    func getSnapshot(in context: Context, completion: @escaping (FavoritesEntry) -> Void) {
+        completion(createEntry(in: context))
+    }
+
+    func placeholder(in context: Context) -> FavoritesEntry {
+        return createEntry(in: context)
+    }
+
+    func getTimeline(in context: Context, completion: @escaping (Timeline<FavoritesEntry>) -> Void) {
+        let timeline = Timeline(entries: [createEntry(in: context)], policy: .atEnd)
```
samsymons: My `WidgetKit` experience is still pretty limited, but since we're refreshing the timeline manually after the favourites change it should be safe to use `.never` for the reload policy.

```suggestion
        let timeline = Timeline(entries: [createEntry(in: context)], policy: .never)
```

Widgets/Widgets.swift
```diff
@@ -0,0 +1,154 @@
...
+    private func createEntry(in context: Context) -> FavoritesEntry {
+        let favorites: [Favorite]
+
+        switch context.family {
+
+        case .systemMedium:
+            favorites = getFavorites(returningNoMoreThan: 4)
+
+        case .systemLarge:
+            favorites = getFavorites(returningNoMoreThan: 8)
```
samsymons: Should this be bumped up to `12?` I tested with the widget today with ~15 favourites and noticed the bottom row was empty, it looks like the large size can support three rows.

```suggestion
            favorites = getFavorites(returningNoMoreThan: 12)
```

Widgets/WidgetViews.swift
```diff
@@ -0,0 +1,235 @@
...
+
+}
+
+struct LargeSearchFieldView: View {
+
+    var body: some View {
+        Link(destination: DeepLinks.newSearch) {
+            ZStack {
+
+                RoundedRectangle(cornerRadius: 21)
```
samsymons: This view could use `Capsule` instead of `RoundedRectangle` if the corner radius should be perfectly rounded on all corners, though I wasn't certain if `21` has been chosen specifically – ignore if so. 😄

samsymons (commented):
Popping a few questions/comments on here after testing the widgets today!

brindy:
Thanks for the suggestions @samsymons! :)

DuckDuckGo/AppDelegate.swift
```diff
@@ -118,6 +119,39 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+
+                let paramKeys: [WidgetFamily: String] = [
+                    .systemSmall: PixelParameters.widgetSmall,
+                    .systemMedium: PixelParameters.widgetMedium,
+                    .systemLarge: PixelParameters.widgetLarge
+                ]
+
+                switch result {
+                case .failure:
+                    Pixel.fire(pixel: .appLaunch, withAdditionalParameters: [PixelParameters.widgetError: "1"])
```
bwaresiak: Maybe we should at least include error domain and code here?

Widgets/Assets.xcassets/WidgetSearchFieldBackgroundColor.colorset/Contents.json
```diff
@@ -0,0 +1,38 @@
...
+        }
+      },
+      "idiom" : "universal"
+    }
+  ],
+  "info" : {
+    "author" : "xcode",
+    "version" : 1
+  }
+}
```
bwaresiak, line 38: 1. Why background colors of the widgets differ? If I put two next to each other these look weird.
2. Dark theme background color seems really dark - I'd consider matching our home screen background color ar at least making it a bit lighter.

bwaresiak (commented):
Great job @brindy! Few comments inline. 

Also, I'd consider making sure that opening app from the (small) search widget or pressing the Search field on other ones will always bring up in-app keyboard - even if "on app launch" option is turned off in settings.

DuckDuckGo/AppDelegate.swift
brindy: Good idea!

Widgets/Assets.xcassets/WidgetSearchFieldBackgroundColor.colorset/Contents.json
brindy: 1) Hmm, they don't.  Each widget's background is set with this line:

`Rectangle().fill(Color.widgetBackground)`

There's a "WidgetBackground" color defined in the asset catalog but that doesn't appear to do anything, but even so I didn't want to delete it in case iOS was expecting it.  But `Color.widgetBackground` resolves to WidgetBackgroundColor.

I've just checked with Pixelmator and the look identical to me.  Can you share a screenshot of what you mean?

2) We tried that originally, but Robert didn't like it compared to the Apple widgets so we decided to darken it - it's been Design and Product Reviewed now so I'm not really wanting to change it at this point.

bwaresiak: Weird, I checked colors and there definitely has been a difference RGB 16/16/16 vs 17/17/17 but after restarting it looks ok... 🤷‍♂️Seems to be fine now 😅

bwaresiak:
@brindy Two more things from me:
1. Bookmark Extension cannot load data on iOS <14.
2. On iPhone SE icons seems to be quite close to the widget's border:

iPhoneSE

brindy:
Thanks @bwaresiak - have fixed 2.

1 appears to be a signing issue. Distributed builds (adhoc, etc) seem to be OK.  We will check on Test Flight during releases.

bwaresiak (approved):

brindy:
Thanks @bwaresiak

PR #709, opened 2020-09-15
Add default browser pushes

THISISDINOSAUR:
Task/Issue URL: https://app.asana.com/0/1187352151074490/1192933920238457/f
Tech Design URL:
CC: @brindy 

**Description**:
Contains three separate pushes for users to change their default browser (all three are iOS 14 only):
- Swaps out the home row onboarding flow on iOS 14 to a new default browser onboarding flow
- Adds a row to settings in the app to change your browser (opens the settings app)
- Adds a message on the home screen (that should only show for users that haven't seen the new onboarding flow)

**Steps to test this PR**:
1. Install the app from scratch and test the new onboarding flow works on iOS 14, and that it doesn't appear on older versions of iOS
2. Open settings and test the default browser button. Do like above for different iOS versions
3. Test home message appears for existing users on iOS 14.
4. Test home message doesn't appear for users that have seen the new onboarding flow
5. Test home message isn't displayed once it's been dismissed/the button has been interacted with
6. Test the home message when the user both does and doesn't have favourites
7. Test the home message sizing works on the smallest/largest devices 
8. Test the home row reminder is now shown based on when the home message was dismissed

**Copy Testing**:

**Orientation Testing**:

* [x] Portrait
* [x] Landscape

**Device Testing**:

* [x] iPhone SE (1st Gen)
* [x] iPhone 8
* [x] iPhone X
* [x] iPad

**OS Testing**:

* [x] iOS 11
* [x] iOS 12
* [x] iOS 13

**Theme Testing**:

* [x] Light theme
* [x] Dark theme

---
###### Internal references:


DuckDuckGo/HomePageConfiguration.swift
```diff
@@ -25,14 +25,26 @@ class HomePageConfiguration {
...
+            .homeMessage,
             .favorites
         ]
     }
     
+    let homeMessageStorage = HomeMessageStorage()
+    
+    func homeMessages() -> [HomeMessageModel] {
+        var messages = [HomeMessageModel]()
+        if #available(iOS 14, *), homeMessageStorage.dateDismissed(forHomeMessage: .defaultBrowserPrompt) == nil {
```
brindy: `!isDismissed(forHomeMessage: .defaultBrowserPrompt)` would be nicer :)

DuckDuckGo/HomeViewController.swift
```diff
@@ -217,6 +217,13 @@ extension HomeViewController: FavoritesHomeViewSectionRendererDelegate {
 
 }
 
+extension HomeViewController: HomeMessageViewSectionRendererDelegate {
+    
+    func homeMessageRenderer(_ renderer: HomeMessageViewSectionRenderer, didDismissHomeMessage homeMessage: HomeMessage) {
+        self.refresh()
```
brindy: `self.` is redundant

carthage-xcode12.sh
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
```
brindy: Delete :)

brindy (commented):
A few minor comments so far.

DuckDuckGo/HomePageConfiguration.swift
THISISDINOSAUR: The dateDismissed method is required when deciding if to show the Home row reminder now, do you think it's still worth having a separate isDismissed() method?

DuckDuckGo/HomeMessagesStorage.swift
```diff
@@ -0,0 +1,54 @@
...
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Core
+
+public class HomeMessageStorage {
```
brindy, line 22: Can we re-use `UserDefaultsWrapper`?

DuckDuckGo/HomePageConfiguration.swift
brindy: Just looking in more detail now I'm coming up for air, so will be a few more comments shortly.  I would encapsulate as much of this logic as makes sense in the storage class tbh.

DuckDuckGo/HomeRowReminder.swift
```diff
@@ -54,13 +60,21 @@ class HomeRowReminder {
...
-        guard let date = storage.firstAccessDate else {
-            storage.firstAccessDate = Date()
-            return false
+    private func hasReminderTimeElapsed(isDefaultBrowserSupported: Bool) -> Bool {
+        if isDefaultBrowserSupported {
+            guard let date = homeMessageStorage.dateDismissed(forHomeMessage: .defaultBrowserPrompt) else {
+                return false
+            }
+            let days = abs(date.timeIntervalSinceNow / 24 / 60 / 60)
+            return days > Constants.reminderTimeInDaysIOS14
```
brindy: Could this be encapsulated in this storage class behind a "hasExpired" function?  In similar situations I've used `shouldShow` and that would usually encapsulate both dismissed and expiry logic.

DuckDuckGo/HomeRowReminder.swift
```diff
@@ -54,13 +60,21 @@ class HomeRowReminder {
...
+        } else {
+            guard let date = storage.firstAccessDate else {
+                storage.firstAccessDate = Date()
+                return false
+            }
+            let days = abs(date.timeIntervalSinceNow / 24 / 60 / 60)
+            return days > Constants.reminderTimeInDays
```
brindy, line 71: As above, if possible.

DuckDuckGo/HomeMessagesStorage.swift
brindy, line 22: Really appreciate you thinking ahead here and separating out the model.  That'll be useful later.  However, I think should try and reuse `UserDefaultsWrapper`, especially given there's only 1 key at the moment and that we haven't got any tests for this.   

The future model I had in mind would be entirely driven by the JSON on the endpoint so I think just assuming it'll be default browser for now is good enough.

brindy (changes requested):
This is looking great - I think if we can just move that logic in to the storage class to minimise how much stuff knows about what we're doing there this will be good to merge.

DuckDuckGo/HomeRowReminder.swift
THISISDINOSAUR: This one's interesting because it has a side effect (of storing the first access date if it hasn't been stored before). I can move it, but the tests use a mocked version of the storage and test that this side effect works correctly, which if I mock will undermine the tests somewhat.
 
I could maybe restructure things, but the way `HomeRowReminder` and `HomeRowReminderStorage` are structured made me think the intention was that `HomeRowReminderStorage` would always just be a thin UserDefaults wrapper, with any logic in `HomeRowReminder`

brindy: Of course, that makes total sense.

brindy (approved):
I gave it a another test by upgrading from develop and then doing a clean install and all looks good to me.  Great work @THISISDINOSAUR ! LGTM!

PR #710, opened 2020-09-16
Add SwiftUI previews for widgets

samsymons:
CC: @brindy 

**Description**:

Note: this might be the wrong branch to PR against. Feel free to change the base branch if necessary :+1:

This PR adds previews of the widgets added in the iOS 14 update. Each widget has been added with both dark and light mode variants, and the widgets are populated with mock data to show that the randomized favourite color is working correctly. Here's a peek at a few of the widgets:

widgets

There is a minor scheme change in this PR which was needed for SwiftUI previews to work. Before this change, the unit test target was being built when the main `DuckDuckGo` scheme builds, which caused compile errors when previews were displayed.

The solution to the compile error issue is to move the unit test config into its own scheme and wire it up to the `DuckDuckGo` scheme so that Command+U works as it always has. I also unchecked the `Show` option in the new `UnitTest` scheme so that it doesn't clutter up the scheme dropdown, since it can just be run from the main app scheme like always – from a dev work perspective, nothing should have changed.

Here's a before and after of the build log to prove that test targets are no longer being built during normal builds:

**Before:**

build-process-before

**After:**

build-process-after

**Steps to test this PR**:

1. Open this branch in Xcode 12
1. Go to the `WidgetViews.swift` file and open the Canvas pane to verify that SwiftUI previews are displayed

**OS Testing**:

* [x] iOS 14

**Theme Testing**:

* [x] Light theme
* [x] Dark theme

### DIFF
--- a/DuckDuckGo.xcodeproj/xcshareddata/xcschemes/DuckDuckGo.xcscheme
+++ b/DuckDuckGo.xcodeproj/xcshareddata/xcschemes/DuckDuckGo.xcscheme
@@ -20,20 +20,6 @@
                ReferencedContainer = "container:DuckDuckGo.xcodeproj">
             </BuildableReference>
          </BuildActionEntry>
-         <BuildActionEntry
-            buildForTesting = "YES"
-            buildForRunning = "YES"
-            buildForProfiling = "NO"
-            buildForArchiving = "NO"
-            buildForAnalyzing = "YES">
-            <BuildableReference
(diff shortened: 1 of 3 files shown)


samsymons:
This PR was brought to you by Xcode 12 build `12A7209` 🕺

brindy:
Yeah, please recreate against `develop` deleting the iOS 14 branch seems to have closed this, sorry!

PR #721, opened 2020-09-22
ensure tabs bar only shows on ipads

brindy:
Task/Issue URL: https://app.asana.com/0/392891325557410/1193615744087793
Tech Design URL:
CC:

**Description**:

Ensure tabs bar only shows on iPads

**Steps to test this PR**:
1. Run on iPhone 11 Max, change to landscape, UI should not change
1. Run on iPad, try different sizes states, UI should change as expected

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone 11 Max
* [ ] iPad

**OS Testing**:

* [ ] iOS 11
* [ ] iOS 14

---
###### Internal references:

### DIFF
--- a/DuckDuckGo/AppWidthObserver.swift
+++ b/DuckDuckGo/AppWidthObserver.swift
@@ -29,16 +29,20 @@ public class AppWidthObserver {
     }
     
     public static let shared = AppWidthObserver()
-    
-    private var currentWidth: CGFloat = 0
+
+    // mutable for testing
+    var isPad = UIDevice.current.userInterfaceIdiom == .pad
+    var currentWidth: CGFloat = 0
     
     public var isLargeWidth: Bool {
-        return currentWidth >= Constants.minPadWidth
+        return isPad && currentWidth >= Constants.minPadWidth
     }
 
     private init() {}
 
     func willResize(toWidth width: CGFloat) -> Bool {
+        guard isPad else { return false }
+        
         if width != currentWidth {
             currentWidth = width
             return true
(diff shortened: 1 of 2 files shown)


DuckDuckGo/AppWidthObserver.swift
```diff
@@ -29,16 +29,20 @@ public class AppWidthObserver {
...
-    
-    private var currentWidth: CGFloat = 0
+
+    // mutable for testing
+    var isPad = UIDevice.current.userInterfaceIdiom == .pad
+    var currentWidth: CGFloat = 0
     
     public var isLargeWidth: Bool {
-        return currentWidth >= Constants.minPadWidth
+        return isPad && currentWidth >= Constants.minPadWidth
```
samsymons, line 38: Do you know if this can ever return false for iPads outside of the test suite after this change? AFAIK the [smallest dimension on any iPad](https://developer.apple.com/library/archive/documentation/DeviceInformation/Reference/iOSDeviceCompatibility/Displays/Displays.html) is 768 (iPad mini), and the `minPadWidth` here is `678`. Does it make sense to drop the width check entirely since this is now always false on iPhone and true on iPad?

That said I could see an argument to be made for keeping it just as a safeguard since there's no guarantee that 768 will always be the smallest dimension on iPad, I'm happy either way here.

samsymons (approved):
One question which isn't a blocker and won't affect app behaviour either way, otherwise LGTM! :shipit:

DuckDuckGo/AppWidthObserver.swift
brindy: The reason we size check is for the split screen scenarios.  The "thin" mode (ie < 678) needs to show the iPhone UI and we built it to basically mimic Safari's behaviour.

PR #726, opened 2020-09-29
Configuration refresh improvements

samsymons:
Task/Issue URL: https://app.asana.com/0/1193060753475688/1108597809786624
Tech Design URL: https://app.asana.com/0/1193060753475688/1195014462180105
CC: @bwaresiak 

## Description

This PR adds support for the `BackgroundTasks` framework and fixes a couple bugs found along the way.

There are three issues addressed in this PR:

**BackgroundTasks should be used on iOS 13+:**

Support for `BackgroundTasks` has been added. It uses `BGProcessingTask`, giving us more time to refresh configuration files than the standard background refresh methods. Because processing tasks are designed to be longer running and interruptible, a new task will be scheduled in the case that the existing one has been canceled.

Another change was made to check for configuration updates every time the app becomes active instead of every time it is launched (using a 12 hour refresh period).

**First launch tracker data was not refreshed correctly:**

To work around an issue with tracker data not being up to date on first launch, rules are now recompiled after tracker data is updated. Previously, we were holding onto a copy of existing `TrackerData` in-memory when rules were recompiled; this has been changed to read the latest set of rules each time `compileRules` is called.

**Cache headers were not always set:**

The final issue fixed is related to the network response caching provided by `URLRequest`. The `If-None-Match` header was not being sent for the tracker file request, causing the file to be reacquired needlessly. I don't know the logic iOS uses to determine whether the header should be automatically used, but my belief is that the OS considers the tracker list to be stale. At the time of writing this PR it was last modified almost a month ago, whereas each of the other files are much more recent and do get the correct headers provided automatically.

None of the cache policies on `URLRequest` provide the behaviour we need, so this PR updates `ContentBlockerRequest` to read the ETag for the configuration it's fetching and pass the header directly.

## Steps to test this PR

**Background task testing:**

1. Open the codebase in Xcode and set a breakpoint in `scheduleBackgroundRefreshTask`
2. Run the app and follow the instructions within that function
3. Check that the appropriate network calls are made, and that no time limit warnings are printed to the console after 30 seconds

The above steps will help you debug background tasks actively, but it's helpful to test that tasks are refreshing in real world scenarios, such as after a reboot and the device is still locked. It's possible to test this by adding extra `os_log` calls locally using the `.default` log level, which you can access from Console.app any time.

**First launch tracker data testing:**

1. Launch a clean install of the app
2. Visit https://airnow.gov
3. Search for a location to confirm that search works; for the bundled tracker data, this website's functionality is blocked, but the latest rules allow it

**Cache header testing:**

This step will require commenting out the 12 hour limit check in `AppConfigurationFetch`. 

1. Start a network traffic proxy, such as Charles or Proxyman
2. Run the app fresh to confirm that configuration files are successfully downloaded
3. Run the app again to trigger a configuration refresh, and confirm that 304 status codes are returned for all configuration files

## Device Testing

* [x] iPhone SE (1st Gen)
* [x] iPhone 8
* [x] iPhone X
* [x] iPad

**OS Testing**:

* [x] iOS 11
* [x] iOS 12
* [x] iOS 13


DuckDuckGo/AppConfigurationFetch.swift
```diff
@@ -78,9 +86,62 @@ class AppConfigurationFetch {
...
+        task.earliestBeginDate = Date(timeIntervalSinceNow: Constants.minimumConfigurationRefreshInterval)
+
+        // Background tasks can be debugged by breaking on the `submit` call, stepping over, then running the following LLDB command, before resuming:
+        //
+        // e -l objc -- (void)[[BGTaskScheduler sharedScheduler] _simulateLaunchForTaskWithIdentifier:@"com.duckduckgo.app.configurationRefresh"]
+
+        do {
+            try BGTaskScheduler.shared.submit(task)
+        } catch {
+            os_log("Error when submitting background task: %s", log: generalLog, type: .debug, error.localizedDescription)
```
samsymons: Question: do we want to report this via `Pixel`? This can throw when we've reached the maximum task limit (10), or if the background task is configured incorrectly, neither of which should be possible, so I settled on logging here.

The current set of reported stats are enough to detect issues with the new background fetch behaviour, but I'm open to changes.

DuckDuckGo/AppConfigurationFetch.swift
```diff
@@ -37,9 +40,14 @@ class AppConfigurationFetch {
...
     private struct Constants {
         static let backgroundTaskName = "Fetch Configuration Task"
+        static let backgroundProcessingTaskIdentifier = "com.duckduckgo.app.configurationRefresh"
+        static let minimumConfigurationRefreshInterval: TimeInterval = 60 * 60 * 12
     }
     
     private struct Keys {
+        static let bgFetchType = "bgft"
+        static let bgFetchTypeBackgroundTasks = "bgbt"
+        static let bgFetchTypeLegacy = "bgl"
```
samsymons, line 48: These are off-the-cuff values – feel free to suggest alternatives!

DuckDuckGo/AppConfigurationFetch.swift
```diff
@@ -108,6 +169,8 @@ class AppConfigurationFetch {
                 store.foregroundNoDataCount += 1
             }
         }
+
+        store.lastConfigurationRefreshDate = Date()
```
samsymons: I'm currently looking into how this function behaves if you have a background task which tries to run after the device has been rebooted and hasn't yet been unlocked, as it modifies `UserDefaults`.

DuckDuckGo/AppDelegate.swift
```diff
@@ -117,6 +124,12 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         if !privacyStore.authenticationEnabled {
             showKeyboardOnLaunch()
         }
+
+        AppConfigurationFetch().start { newData in
+            if newData {
+                NotificationCenter.default.post(name: ContentBlockerProtectionChangedNotification.name, object: nil)
```
samsymons, line 130: Open to suggestions for a better way to react to tracker data changes – this is just a way to have `TabViewController.onContentBlockerConfigurationChanged` called since we want rules recompiled, but I'm always skeptical about falling back on notifications since they can make code harder to follow.

(Am also open to defining a separate notification for this, since the name of this one isn't perfectly suitable.)

bwaresiak: This seems fine to me, notifications should be a valid pattern to use here as there may be multiple observers and at least few places possibly triggering this to fire.

DuckDuckGo/AppConfigurationFetch.swift
bwaresiak: I'd add debug pixel here (along with error code), we can always get rid of it after validating outcome of this project. If the expectation is that this should not happen, then it's easy to check - number of pixels should be 0. :)

DuckDuckGo/AppUserDefaults.swift
```diff
@@ -141,6 +143,14 @@ public class AppUserDefaults: AppSettings {
 }
 
 extension AppUserDefaults: AppConfigurationFetchStatistics {
+    var lastConfigurationRefreshDate: Date? {
```
bwaresiak: We could probably move to @UserDefaultsWrapper and use distantPast as a default

samsymons: Great suggestion, thanks @bwaresiak – will update.

bwaresiak (approved):
LGTM! I can confirm that background refresh is working - for me it requires device being connected to power source.

samsymons:
@bwaresiak I confirmed that background tasks do have `requiresExternalPower` set to `false` by default, good suggestion. 👍 

I also added a check for the target environment when the background task submit call throws an exception on the simulator, BackgroundTasks isn't supported on the sim so we shouldn't report those failures.

PR #727, opened 2020-09-29
Translations

bwaresiak:
Task/Issue URL: https://app.asana.com/0/856498667320406/1176971145104443
Tech Design URL: https://app.asana.com/0/856498667320406/1186381273740755
CC:

**Description**:
Translations plus some automations to make import/export easier.

Notes:
- Still waiting for translation for few strings (Welcome to... and one dax onboarding).

**Steps to test this PR**:

1. Run the app, ensure it is localized.
2. See if layout works as expected.
3. Validate new scripts. 

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPad

**OS Testing**:

* [ ] iOS 11
* [ ] iOS 12
* [ ] iOS 13

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:


THISISDINOSAUR:
1) I don't speak German, so some of these may be fine, but when testing it seems like there are some issues with missing German translations (haven't tested every language, but it looks like it was just the German that's missing). 
There's also potentially a few things that might be a problem for every language

Minor thing, currently our debug scheme is set to run the app in English, it might be worth changing it to the system language?

2) Layout looks good to me

3) I got an error when running `./loc_export.sh` (although it otherwise seemed to work):
`genstrings: error: bad entry in file /Users/ellesullivan/DuckDuckGo/iOS/DuckDuckGo/FeedbackFormViewController.swift (line = 231): Argument is not a literal string.`

German only: Possibly that "Subject Key Identifier" and two instances of "Public Key" (header and later on) need translating
Screenshot 2020-10-05 at 17 52 47
Screenshot 2020-10-05 at 17 52 54

All languages: Possible that "No tracking" needs translating. 
Screenshot 2020-10-05 at 17 53 25

German only: "private Tabs" might need translating
Screenshot 2020-10-05 at 17 53 48


All languages: Settings page "Version" text doesn't seem to be translated
Screenshot 2020-10-05 at 17 55 00

All languages (including English): on the list of trackers blocked there's a very minor layout issue with the bottoms of characters being cut off (I checked the current app and it looks like a new issue).
It looks like we don't translated the "Advertising" text , although that sounds like a tracker radar issue?
Screenshot 2020-10-05 at 17 56 56

bwaresiak:
@THISISDINOSAUR thank you for your feedback :) 

1. I know some part of translations is missing or imperfect - we are working on a solution but still will progress as we are. This will be eventually fixed. 🤞 
2. I've made it so scheme uses system language as default run option.
3. Error on export is "normal" - it is some genstrings legacy problem caused by NSLcalizedString with value parameter. This should not affect the overall process.
4. "No tracking", "Advertising" etc comes from external sources: TDS, TL;DR data and is not something I've considered changing, especially as future of Privacy Protection screen is questionable.
5. Other issues should be resolved after recent update of translation files.

THISISDINOSAUR (approved):
In that case, everything looks good to me! 
I tested the latest changes and couldn't find anything that looks unexpected

bwaresiak:
Thank you @THISISDINOSAUR !

PR #732, opened 2020-10-11
Fix fullscreen video

### LINKED ISSUE #675: Can‘t watch YoutTube videos in fullscreen (iPad)
Hi, in the DDG app it is currently not possible to watch YouTube (I know it's from Google😅) videos in full screen view. The fullscreen button is greyed out. (As you can see in picture 1)

image

Since it works fine in Safari (see picture 2), it seems to be a bug in the DDG app.

image

Device: iPad Pro 12.9" 3rd generation
Software: iPadOS 13.5.1
App-Version: 7.48.1.0

samsymons:
Task/Issue URL: https://app.asana.com/0/72649045549333/1183304843940541
Fixes: #675 
CC: @SlayterDev 

**Description**:

This PR fixes a gap in WKWebView's JS APIs, where the fullscreen APIs exist and function correctly, but [`fullscreenEnabled`](https://developer.mozilla.org/en-US/docs/Web/API/Document/fullscreenEnabled) is undefined. That property is checked by YouTube and others to determine whether fullscreen is possible.

There is a new script which checks whether `fullscreenEnabled` is defined and overrides it if it isn't. In case this gets fixed one day, the builtin implementation is checked for and used if it's found instead. An implementation of `requestFullscreen` was required to get this to work, which just calls through to `webkitEnterFullscreen` on the HTML element requesting it.

@SlayterDev I'd love to get your eyes on this to verify that the JS makes sense. 🙂

**Steps to test this PR**:
1. Open a video on YouTube.com on an iPad
1. Test that the video is able to go fullscreen

_**Note:** I tested this on iPhone and iPad with all supported iOS versions but found that iPhones running iOS 11 and 12 will exit fullscreen immediately after entering it, even on `develop`. It only happens on mobile, iPhones on 11 and 12 will work correctly if desktop mode is requested. I'd love for someone else to double check that and confirm this PR hasn't caused a regression._

**Device Testing**:

* [x] iPhone SE (1st Gen)
* [x] iPhone 8
* [x] iPhone X
* [x] iPad

**OS Testing**:

* [x] iOS 11
* [x] iOS 12
* [x] iOS 13
* [x] iOS 14

### DIFF
--- a/Core/fullscreenvideo.js
+++ b/Core/fullscreenvideo.js
@@ -1,0 +1,20 @@
+//
+//  fullscreenvideo.js
+//  DuckDuckGo
+//
+//  Copyright © 2020 DuckDuckGo. All rights reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//  http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+
--- a/Core/fullscreenvideo.js
+++ b/Core/fullscreenvideo.js
@@ -25,7 +25,7 @@
     let canEnterFullscreen = HTMLVideoElement.prototype.webkitEnterFullscreen !== undefined;
     let browserHasExistingFullScreenSupport = document.fullscreenEnabled || document.webkitFullscreenEnabled;
 
-    // YouTube Mobile won't exit fullscreen correctly if this patch is active. Reference: https://github.com/brave/brave-ios/pull/2002
+    // YouTube Mobile won't exit fullscreen correctly if requestFullscreen is overridden. Reference: https://github.com/brave/brave-ios/pull/2002
     let isMobile = /mobile/i.test(navigator.userAgent);
 
     if (!browserHasExistingFullScreenSupport && canEnterFullscreen && !isMobile) {
--- a/Core/fullscreenvideo.js
+++ b/Core/fullscreenvideo.js
@@ -25,7 +25,10 @@
     let canEnterFullscreen = HTMLVideoElement.prototype.webkitEnterFullscreen !== undefined;
     let browserHasExistingFullScreenSupport = document.fullscreenEnabled || document.webkitFullscreenEnabled;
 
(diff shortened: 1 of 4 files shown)


brindy (approved):
Great work @samsymons - our users will love this! :)

Core/fullscreenvideo.js
```diff
@@ -0,0 +1,49 @@
...
+(function() {
+    let canEnterFullscreen = HTMLVideoElement.prototype.webkitEnterFullscreen !== undefined;
+    let browserHasExistingFullScreenSupport = document.fullscreenEnabled || document.webkitFullscreenEnabled;
+
+    // YouTube Mobile won't exit fullscreen correctly if requestFullscreen is overridden. Reference: https://github.com/brave/brave-ios/pull/2002
+    let isMobile = /mobile/i.test(navigator.userAgent);
+
+    if (!browserHasExistingFullScreenSupport && canEnterFullscreen && !isMobile) {
+        Object.defineProperty(document, "fullscreenEnabled", {
+            get: function() {
```
SlayterDev: This is mainly a practices thing but if you're only returning a value here you can use the `value` property instead of a `get` function. See here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty

Core/fullscreenvideo.js
```diff
@@ -0,0 +1,49 @@
...
+                return true;
+            }
+        });
+
+        HTMLElement.prototype.requestFullscreen = function() {
+            let video = this.querySelector("video");
```
SlayterDev, line 37: Not sure what the alternative would be but my fear is the selector you're targeting may not always be called `video`.

SlayterDev (changes requested):
Hey @samsymons looks great overall. Just one minor JS practices thing and one other comment.

Core/fullscreenvideo.js
brindy: @SlayterDev While this fixes things for YouTube, would this (entire PR) likely have any negative affect elsewhere?

SlayterDev: I doubt it would have negative effects. Just if a video player calls their selector something weird it'll get missed.

brindy: Looks like it would work for `<video>` tags though, so will cover most (modern) implementations right?

SlayterDev: Ah jeez. Ignore this entire chain. I got `selector` and `name` mixed up 😅

samsymons: Yep, `<video>` tags are what I'm aiming for. 😄

Core/fullscreenvideo.js
samsymons: Oh nice, I didn't know that – thanks @SlayterDev!

SlayterDev (approved):
LGTM 👍

PR #738, opened 2020-10-12
Turn SwiftLint back on

samsymons:
Task/Issue URL: https://app.asana.com/0/414709148257752/1196995126357495
CC:

**Description**:

This PR turns SwiftLint back on and fixes existing violations.

There are a couple things to do before this PR is ready to be merged:

* [x] Test that CI is failing PRs appropriately – will introduce a failing change on this branch and see what happens
* [x] Verify that the fixes here make sense. For example, I have a couple `type_body_length` comments in test files, but it might make more sense to disable those for the `UnitTests` directly wholesale (assuming SwiftLint can configure individual rules to exclude specific directories?)

**Steps to test this PR**:
1. Build the main app target and verify that no warnings appear
1. Introduce a violation deliberately and verify that a warning appears

### DIFF
--- a/IntegrationTests/ContentBlockingRulesTests.swift
+++ b/IntegrationTests/ContentBlockingRulesTests.swift
@@ -50,10 +50,8 @@ class ContentBlockingRulesTests: XCTestCase {
 
 extension Array where Element == ContentBlockerRule {
     func findExactFilter(filter: String) -> ContentBlockerRule? {
-        for rule in self {
-            if rule.trigger.urlFilter == filter {
-                return rule
-            }
+        for rule in self where rule.trigger.urlFilter == filter {
+            return rule
         }
         
         return nil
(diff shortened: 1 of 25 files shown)


samsymons:
I've tested that SwiftLint fails on CI in this commit: https://github.com/duckduckgo/iOS/pull/738/commits/b6adfc85e5001a8c52e0fd51ceffd1bb260da895

IntegrationTests/ContentBlockingRulesTests.swift
```diff
@@ -50,10 +50,8 @@ class ContentBlockingRulesTests: XCTestCase {
 
 extension Array where Element == ContentBlockerRule {
     func findExactFilter(filter: String) -> ContentBlockerRule? {
-        for rule in self {
-            if rule.trigger.urlFilter == filter {
-                return rule
-            }
+        for rule in self where rule.trigger.urlFilter == filter {
```
brindy, line 53: forgot about this syntax 👍

samsymons: To be honest I did too! But SwiftLint actually detected a single `if` statement inside the body of the loop and recommended this, pretty impressive.

brindy (approved):
Thanks for this. LGTM!

As for the choices you made around type_body_length vs applying to all UnitTests, I think this is OK for now.  We can review that another time. :)

PR #740, opened 2020-10-14
Add configuration refresh info to the debug menu

samsymons:
CC: @brindy 

**Description**:

This PR adds info around the config refresh process to the debug menu. Specifically, it adds:

1. The last refresh date
2. A way to reset the last refresh date
3. A list of pending `BGTaskRequest` objects
4. A list of cached config files with their ETags

**Steps to test this PR**:
1. Run the app in a debug configuration
1. Go into settings and tap the Version row at the bottom
1. Tap the `Configuration Refresh Info` row

### DIFF
--- a/DuckDuckGo/ConfigurationDebugViewController.swift
+++ b/DuckDuckGo/ConfigurationDebugViewController.swift
@@ -26,13 +26,15 @@ class ConfigurationDebugViewController: UITableViewController {
 
     private let titles = [
         Sections.refreshInformation: "Background Refresh Info",
-        Sections.queuedTasks: "Queued Tasks (Earliest Execution Date)"
+        Sections.queuedTasks: "Queued Tasks (Earliest Execution Date)",
+        Sections.etags: "ETags"
     ]
 
     enum Sections: Int, CaseIterable {
 
         case refreshInformation
         case queuedTasks
+        case etags
 
     }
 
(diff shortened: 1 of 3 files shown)


brindy:
This looks cool @samsymons 😎 

1. Should there always be pending tasks?  I'm not seeing any (only tested on the simulator so far though).
2. The date string can get truncated.  How about having the date string as `detailLabel` instead?
3. If you feel you've got time and motivation to add ETags it would definitely be useful imho :)

Thanks!

samsymons:
@brindy Great call on the truncated date, I've updated it to be set on the detail text label and changed the date formatter to use a `.short` date style since the time is the most important part.

Sadly yeah it'll always be empty for pending tasks on the simulator, BackgroundTasks doesn't work on the sim. 😞

samsymons:
I ended up going with the subtitle cell style after adding ETag info, since those were gonna get extremely clipped otherwise.

**Note:** The Bloom filter ETags are always empty, they're not being cached after the download completes ([reference](https://github.com/duckduckgo/iOS/blob/develop/Core/ContentBlockerLoader.swift#L117)). This PR isn't the right spot to investigate this, I'll write up a ticket.

Simulator Screen Shot - iPhone 8 - 2020-10-14 at 19 11 55

DuckDuckGo/ConfigurationDebugViewController.swift
```diff
@@ -0,0 +1,163 @@
...
+        }
+    }
+
+    // MARK: - ETag User Defaults Wrapper
+
+    lazy var defaults = UserDefaults(suiteName: "com.duckduckgo.blocker-list.etags")
+
+    private func etag(for config: ETagRows) -> String? {
+        return defaults?.string(forKey: config.rawValue)
+    }
```
samsymons, line 161: This isn't ideal, but to use `UserDefaultsETagStorage` directly involves making a ton of stuff in `Core` public. A few lines of duplicated code is the better approach IMO since this is just a debug menu.

brindy (approved):
Noice! 👍

PR #750, opened 2020-10-19
Add widget onboarding flow and onboarding experiment

THISISDINOSAUR:
Task/Issue URL: https://app.asana.com/0/1187352151074490/1194681707917578
Tech Design URL:
CC: @brindy 

**Description**:
Adds a new widget onboarding flow, along with an experiment to determine which onboarding flow is best (default browser, home row, or widget). The experiment should only show for iOS 14 users when the app is in English.

Screenshot 2020-10-19 at 12 59 04 Screenshot 2020-10-19 at 12 57 09

**Steps to test this PR**:
1. Delete and reinstall the app repeatedly to ensure all variants are bucketed correctly
2. Delete and reinstall the app repeatedly on an device with iOS <14 to ensure it is not part of the experiment (and only sees the home row onboarding)
3. Same as above for when the app is in a language other than English 

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPad

**OS Testing**:

* [ ] iOS 11
* [ ] iOS 12
* [ ] iOS 13
* [ ] iOS 14

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:


bwaresiak:
Nice @THISISDINOSAUR 👍  I haven't got to see the code yet, but just few UX questions/things:

widgets

1. Do we need that back button? Thus far we never had that on onboarding and it seems.. weird to me. :)
2. Top label needs some scaling. :)
3. Background color of that screen does not seem to match the color of the first onboarding screen.
4. We often use custom spacing (larger) between lines for labels, I'd double check with the designer if what we have currently here matches their expectation.

THISISDINOSAUR:
Thanks a lot for reviewing! 
Re 1, good point, I've messaged Robert about it, although it's currently in Product review and has finished design review, so probably won't be worth changing anything now.
Re 2, 3 good catches, fixed!
Re 4, ah good catch, I didn't realise we did that, it wasn't what it was supposed to be so I've fixed it now

THISISDINOSAUR:
Re 1: Robert thinks we should keep it rather than forcing users through a multistep flow with no option to go back.

DuckDuckGo/OnboardingViewController.swift
```diff
@@ -21,10 +21,14 @@ import UIKit
 import Core
 
 class OnboardingViewController: UIViewController, Onboarding {
-    
-    private var controllerNames: [String] = {
-        if #available(iOS 14.0, *) {
+        
+    private lazy var controllerNames: [String] = {
+        if variantManager.isSupported(feature: .onboardingHomeRow) {
+            return ["onboardingHomeRow"]
```
brindy, line 27: Technically redundant, right?  But don't feel strongly either way, because this is also explicit which is good.

Core/VariantManager.swift
```diff
@@ -44,10 +60,14 @@ public struct Variant {
...
         
         // SERP testing
-        Variant(name: "sc", weight: 1, isIncluded: When.inRequiredCountry, features: []),
+        Variant(name: "sc", weight: doNotAllocate, isIncluded: When.inRequiredCountry, features: []),
         Variant(name: "sd", weight: doNotAllocate, isIncluded: When.always, features: []),
-        Variant(name: "se", weight: 1, isIncluded: When.inRequiredCountry, features: [])
+        Variant(name: "se", weight: doNotAllocate, isIncluded: When.inRequiredCountry, features: []),
 
+        // Onboarding
+        Variant(name: "mh", weight: 1, isIncluded: When.inEnglishAndIOS14, features: [ .onboardingHomeRow ]),
```
brindy, line 68: Noice :)

brindy (approved):
Just one optional thing, LGTM!  Feel free to merge if/when you get that updated asset.

DuckDuckGo/OnboardingViewController.swift
THISISDINOSAUR: Yeah, totally redundant, but I preferred the explicitness of it.

PR #752, opened 2020-10-20
Fix screen rotation when onboarding is being shown

THISISDINOSAUR:
Task/Issue URL: https://app.asana.com/0/inbox/1187352150811184
Tech Design URL:
CC: @brindy 

**Description**:
We have an issue in production that allows the orientation to be changed to landscape during onboarding (see Asana task), which the screens aren't designed for

**Steps to test this PR**:
1. Attempt to go through onboarding when starting in landscape
2. Go through onboarding when starting in portrait, and attempt to rotate to landscape on each screen
3. Ensure rotation still works as expected after onboarding

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPad

**OS Testing**:

* [ ] iOS 11
* [ ] iOS 12
* [ ] iOS 13
* [ ] iOS 14

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:

### DIFF
--- a/DuckDuckGo/MainViewController.swift
+++ b/DuckDuckGo/MainViewController.swift
@@ -315,6 +315,14 @@ class MainViewController: UIViewController {
 
     }
     
+    override var supportedInterfaceOrientations: UIInterfaceOrientationMask {
+        return DefaultTutorialSettings().hasSeenOnboarding ? [.allButUpsideDown] : [.portrait]
+    }
+
+    override var shouldAutorotate: Bool {
+        return true
+    }
+    
     @objc func dismissSuggestionTray() {
         dismissOmniBar()
     }


DuckDuckGo/MainViewController.swift
```diff
@@ -315,6 +315,14 @@ class MainViewController: UIViewController {
 
     }
     
+    override var supportedInterfaceOrientations: UIInterfaceOrientationMask {
+        return DefaultTutorialSettings().hasSeenOnboarding ? [.allButUpsideDown] : [.portrait]
```
THISISDINOSAUR, line 319: Question, will this ever not work? E.g. is it possible we'll have users that haven't seen the onboarding, but then won't be shown it?

brindy: It's more to cover the case that you might not finish the onboarding then leave/kill the app and start it again.

brindy (approved):
LGTM! Nice one.

DuckDuckGo/MainViewController.swift
THISISDINOSAUR: I'm a little bit weary of this solution in case we change that behaviour in the future. I could instead keep track of if onboarding is being shown, rather than checking if it has been seen. What do you think?

brindy: This is probably going to be slightly cleaner than testing to see what view controller is visible, but I don't feel strongly either way tbh.

PR #755, opened 2020-10-22
Add additional configuration refresh stats

samsymons:
Task/Issue URL: https://app.asana.com/0/1193060753475688/1198516960588141
CC: @bwaresiak 

**Description**:

This PR adds extra stats to the configuration refresh process, to better understand where errors may be occurring.

1. An early return has been added when scheduling a background task on a device which doesn't have permission, to avoid sending needless error pixels
1. A new stat has been added which is tracked when a background task is expired by the device
1. Stats have been added on a per-config file basis to help us understand if any files are causing problems

**Steps to test this PR**:
1. Manually trigger a configuration refresh and check the parameters being sent with the pixel call
1. Schedule a background task and manually force it to expire (instructions can be found via a comment in the code) to check that the expiration stat is tracked

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [x] iPad

**OS Testing**:

* [ ] iOS 11
* [ ] iOS 12
* [ ] iOS 13
* [x] iOS 14

### DIFF
--- a/Core/ContentBlockerLoader.swift
+++ b/Core/ContentBlockerLoader.swift
@@ -115,13 +115,13 @@ public class ContentBlockerLoader {
             guard case ContentBlockerRequest.Response.success(_, let data) = response,
                 let specification = try? HTTPSUpgradeParser.convertBloomFilterSpecification(fromJSONData: data)
                 else {
+                    progress?(.httpsBloomFilter)
                     semaphore.signal()
                     return
             }
             
             if let storedSpecification = self.httpsUpgradeStore.bloomFilterSpecification(), storedSpecification == specification {
                 os_log("Bloom filter already downloaded", log: generalLog, type: .debug)
-                // Mark the Bloom Filter as downloaded if the current version already exists locally.
                 progress?(.httpsBloomFilter)
                 semaphore.signal()
                 return
--- a/Core/ContentBlockerLoader.swift
+++ b/Core/ContentBlockerLoader.swift
@@ -21,7 +21,7 @@ import Foundation
 import os.log
 
 public class ContentBlockerLoader {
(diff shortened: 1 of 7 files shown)


Core/ContentBlockerLoader.swift
```diff
@@ -109,11 +121,17 @@ public class ContentBlockerLoader {
```
bwaresiak, line 122: I wonder: maybe we also should mark `httpsBloomFilter` as completed if in the above block, as effectively it won't be invoked (and should not be marked as an potential issue).

DuckDuckGo/AppUserDefaults.swift
```diff
@@ -202,4 +211,67 @@ extension AppUserDefaults: AppConfigurationFetchStatistics {
...
     }
+
+    var backgroundTaskExpirationCount: Int {
+        get {
+            return userDefaults?.integer(forKey: Keys.backgroundFetchTaskExpirationCount) ?? 0
+        }
+        set {
+            userDefaults?.setValue(newValue, forKey: Keys.backgroundFetchTaskExpirationCount)
+        }
+    }
```
bwaresiak: Can we use property wrappers here? Wouldn't that simplify the code?

samsymons: Great call, we should do that. I added it this way for consistency with the current approach, but new additions should definitely be using the more modern approach – I'll update this, thanks!

Core/ContentBlockerLoader.swift
samsymons: Great catch, thanks!

bwaresiak (approved):
LGTM! 👍

PR #757, opened 2020-10-25
Add a short threshold before the keyboard is shown on app foreground

st3fan:
> I am probably skipping a few steps with this patch, which is why I marked it as a draft. I am submitting this because it fixes a small itch that I had. If you are interesting in picking this up then I'd love to get it into a state that makes sense for the product. I've been running this in my local build.

This patch introduces a time interval that is used to determine whether the keyboard should be shown on _App Foreground_ when the _Show Keyboard on App Launch_ setting is enabled.

I like that feature, because it is what I want when I open the browser for first time. Or after not using it for a long time. 

However in my day to day use I often am in the situation where I switch back and forth between the browser and another app in a very short time interval and in those cases I do not want the keyboard to show up. My intent then is mostly is to continue browsing the same site or reading the page I already had open in the current tab - and not to visit a new page.

I settled on a pretty low threshold of 45 seconds. This works for me but may not be representative in general. I'm not sure if this application has telemetry events, but those could be used to look at user behaviour to determine if this is a change that makes sense for a bigger audience or to determine a better threshold.

No proper testing has been done - I've only used this on an iPhone X running iOS 14.1.

## TODO

Task/Issue URL: https://app.asana.com/0/392891325557410/1198923366005691
Tech Design URL:
CC:

**Description**:

**Steps to test this PR**:
1.
1.

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPad

**OS Testing**:

* [ ] iOS 11
* [ ] iOS 12
* [ ] iOS 13
* [ ] iOS 14

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:

### DIFF
--- a/DuckDuckGo/AppDelegate.swift
+++ b/DuckDuckGo/AppDelegate.swift
@@ -28,7 +28,7 @@ import BackgroundTasks
 @UIApplicationMain
 class AppDelegate: UIResponder, UIApplicationDelegate {
 
-    private static let ShowKeyboardOnLaunchThreshold = TimeInterval(45)
+    private static let ShowKeyboardOnLaunchThreshold = TimeInterval(20)
     
     private struct ShortcutKey {
         static let clipboard = "com.duckduckgo.mobile.ios.clipboard"
--- a/DuckDuckGo/AppDelegate.swift
+++ b/DuckDuckGo/AppDelegate.swift
@@ -28,6 +28,8 @@ import BackgroundTasks
 @UIApplicationMain
 class AppDelegate: UIResponder, UIApplicationDelegate {
 
+    private static let ShowKeyboardOnLaunchThreshold = TimeInterval(45)
+    
     private struct ShortcutKey {
         static let clipboard = "com.duckduckgo.mobile.ios.clipboard"
     }
@@ -47,6 +49,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     private lazy var privacyStore = PrivacyUserDefaults()
     private var autoClear: AutoClear?
     private var showKeyboardIfSettingOn = true
+    private var lastBackgroundDate: Date?
 
     // MARK: lifecycle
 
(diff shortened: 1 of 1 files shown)


brindy:
Thanks for this and thanks for making it a draft.  

Generally we prefer issues first, as per our [contributions guidelines](https://github.com/duckduckgo/iOS/blob/develop/CONTRIBUTING.md).

But let me run this past the product review team this week, I think it's a good change.

st3fan:
Would you like me to write a story in an issue?

brindy:
No, that's ok this time, thanks :) 

I've submitted the change for feedback and we're tracking it internally at: https://app.asana.com/0/392891325557410/1198923366005691

brindy:
Hi! Thanks for your patience. :)  

I took this through the product review process and we like the change in general but think a better threshold would be 20 seconds.  Would you be OK with that? 

Thanks.

st3fan:
Awesome. I think a lower delay is fine.

brindy:
Hi @st3fan - do you want to push a commit with that change, then we can get this reviewed and merged this week.  Thanks in advance!

DuckDuckGo/AppDelegate.swift
```diff
@@ -28,6 +28,8 @@ import BackgroundTasks
 @UIApplicationMain
 class AppDelegate: UIResponder, UIApplicationDelegate {
 
+    private static let ShowKeyboardOnLaunchThreshold = TimeInterval(45)
```
samsymons: This is a great quality-of-life improvement! This can be merged once the timing threshold is updated, thanks @st3fan! 🎉 

```suggestion
    private static let ShowKeyboardOnLaunchThreshold = TimeInterval(20)
```

samsymons (approved):
:shipit:

PR #767, opened 2020-10-30
Search in bookmarks

bwaresiak:
Task/Issue URL: https://app.asana.com/0/856498667320406/1178518875908026
Tech Design URL:
CC:

**Description**:
Added search to bookmarks list.

**Steps to test this PR**:
1. Open Bookmarks.
2. Search for something.

Also to look for:
- Empty state.
- Dismissal.
- Edit mode.
- iOS 12&13.

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPad

**OS Testing**:

* [ ] iOS 11
* [ ] iOS 12
* [ ] iOS 13
* [ ] iOS 14

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:


DuckDuckGo/en.lproj/Localizable.strings
```diff
@@ -244,6 +244,9 @@
 /* Empty list state placholder */
 "empty.favorites" = "No favorites yet";
 
+/* Empty search placholder on bookmarks search */
+"empty.search" = "No matches found";
+
```
bwaresiak, line 249: Will update with translations when these are ready.

DuckDuckGo/UserText.swift
```diff
@@ -175,6 +175,7 @@ public struct UserText {
     
     public static let emptyBookmarks = NSLocalizedString("empty.bookmarks", value: "No bookmarks yet", comment: "Empty list state placholder")
     public static let emptyFavorites = NSLocalizedString("empty.favorites", value: "No favorites yet", comment: "Empty list state placholder")
+    public static let noMatchesFound = NSLocalizedString("empty.search", value: "No matches found", comment: "Empty search placholder on bookmarks search")
```
samsymons:
```suggestion
    public static let noMatchesFound = NSLocalizedString("empty.search", value: "No matches found", comment: "Empty search placeholder on bookmarks search")
```

samsymons (commented):
This is really awesome 😍 The text matching works perfectly, great work @bwaresiak.

There are some edge cases I bumped into:

* The search header doesn't scale perfectly if Dynamic Type changes while it's visible
* It's possible to get an out-of-bounds crash with the iPad keyboard
* I've had two cases where the items in the bookmarks list become mostly uninteractive, tapping them won't load the bookmark and tapping their Delete button won't delete them. I have no clue how to reproduce this again, and it has only happened in the simulator – I bashed away on it on my iPad and couldn't get it to happen, it might be a non-issue

I haven't finished a proper code review yet, just wanted to write these up before I forgot.

---

### Dynamic Type header layout

This mainly affects iPhone, and it only happens if you change Dynamic Type while the search header is visible. If you change it, and then open the Bookmarks view controller, everything is fine.

This is enough of an edge case that it's not a blocker, IMO it's fine to skip – it never causes anything to break. 🙂

dynamic-type

---

### Out of bounds crash

_**Note**: I don't think this was introduced in this PR, so I can write it up as a bug instead if it's something worth fixing later, LMK. I found no out-of-bounds crashes with the search filtering._

This is a pretty sneaky edge case. You can get a crash by doing the following:

1. Connect a keyboard to your device
2. Navigate to a page that you haven't bookmarked
3. Open the Bookmarks modal
4. Use Command+D to add the current page (behind the Bookmarks modal) as a bookmark
5. Delete an entry in the bookmarks list

There's very low probability that this issue will come up, but it could be worth investigating in case there are any other ways this can be triggered.

bwaresiak:
@samsymons Thanks for the review. I've fixed the crash by adding listener to update when model is changed outside of the list view. 

For the dynamic font, I've noticed that the same issue happens in Messages app, so I think I'll leave it as it is - seems to be a corner case anyway and it's is just a glitch.

Also if you manage to reproduce that "non-responsive UI" issue on device please ping me - I've been unable to see it so far.

samsymons (approved):
Tested this again after the changes, looks really good! I only found one issue, where the search field font color on pre iOS 13 devices is incorrect when using the dark theme. It also lacks some padding at the top, but it's usable.

Beyond this issue, everything else looks good. :shipit: 

Simulator Screen Shot - iPhone 5s - 2020-11-08 at 21 20 33

bwaresiak:
@samsymons Thanks, I can't do much about the padding I'm afraid but I've tweaked the search bar style based on the current theme.

samsymons:
@bwaresiak Cool, that seems okay to me. 🙂

brindy:
Have checked the following with a fresh install, checked a few dax dialogs, checked the privacy dashboard and checked settings:

* Czech
* Swedish
* Russian
* French
* German

All seem fine. Let me know if you'd like to check any others.

bwaresiak:
Thanks @brindy, @samsymons

PR #768, opened 2020-10-31
exclude accounts.google.com UA

brindy:
Task/Issue URL: https://app.asana.com/0/inbox/392891325557408/1198978223513065/1198984670982837
Tech Design URL:
CC:

**Description**:

Exclude accounts.google.com from the ua and fix bug which prevented the UA being applied when disabling links opening in other sites.

**Steps to test this PR**:
1. Visit home.next.com and try to login with Google (doesn't matter if you have a Nest, but you do need a Google account)
1. Without the fix you'd see a message about the browser not being supported
1. With the fix you should be able to login
1. Disable "Open in Associated Apps"
1. You should still be able to login

---
###### Internal references:

### DIFF
--- a/Core/UserAgentManager.swift
+++ b/Core/UserAgentManager.swift
@@ -80,7 +80,8 @@ struct UserAgent {
     
     private static let sitesThatOmitApplication = [
         "cvs.com",
-        "sovietgames.su"
+        "sovietgames.su",
+        "accounts.google.com"
     ]
     
     private let baseAgent: String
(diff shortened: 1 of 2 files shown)


samsymons (approved):
Good catch, this looks solid. For some reason I wasn't getting the Browser Not Supported banner at all, even before this fix, so I wasn't able to QA that issue directly. That said, this change makes sense and I can see why it would help.

I tested a few different sites, Medium, Stack Overflow etc. with no issues. :+1:

PR #769, opened 2020-11-06
Migrate model objects to Codable

samsymons:
Task/Issue URL: https://app.asana.com/0/1193060753475688/791549378835838
CC: @brindy

**Description**:

This PR migrates remaining `JSONSerialization` usage over to `Codable`. There are two main changes:

1. The `Suggestion` object has been simplified a lot, and the `AutocompleteParser` has been removed entirely.
1. The `TLD` class has been updated to use `JSON(En|De)coder`, but hasn't been fully moved over to `Codable`. IMHO there's no benefit to that since this class is only initialized through a single `init` function and encoded into JSON via the `json` computed property. We could make `TLD` conform to `Codable` but would still need a way to create it from a bundled JSON file. Happy to be convinced otherwise 🙂 

**Steps to test this PR**:
1. Test the autocomplete feature of the app

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [x] iPhone X
* [ ] iPad

**OS Testing**:

* [ ] iOS 11
* [ ] iOS 12
* [ ] iOS 13
* [x] iOS 14

---
###### Internal references:


DuckDuckGo/Suggestion.swift
```diff
@@ -18,9 +18,22 @@
...
 import Foundation
+import Core
 
-struct Suggestion {
+struct Suggestion: Decodable {
 
     let type: String
     let suggestion: String
+
+    init(from decoder: Decoder) throws {
```
samsymons, line 28: Needed a custom init function here since the `type` and `suggestion` values are derived from a single key/value pair in the dictionary. Is there a cleaner way to handle this?

bwaresiak: I think this is fine :)

DuckDuckGo/Suggestion.swift
```diff
@@ -18,9 +18,22 @@
...
+        let container = try decoder.singleValueContainer()
+        let dictionary = try container.decode([String: String].self)
```
bwaresiak, line 30: Shouldn't we use `[String: String?]` to be on a safe side?

DuckDuckGo/Suggestion.swift
```diff
@@ -18,9 +18,22 @@
...
+
+        if let key = dictionary.keys.first, let value = dictionary[key] {
```
bwaresiak: You could use `let entry = dictionary.first` and just reference key/value without dictionary lookup.

bwaresiak (commented):
Good job @samsymons , few comments inline. :)

DuckDuckGo/Suggestion.swift
bwaresiak: On a second thought it is most likely fine - this is going to throw in case of such error anyway.

DuckDuckGo/Suggestion.swift
samsymons: Ah good call, thanks!

bwaresiak (approved):
LGTM 👍

PR #773, opened 2020-11-11
Prevent the blank snapshot view from appearing when the app is active

samsymons:
Task/Issue URL: https://app.asana.com/0/1193060753475688/1195019227318136
Task/Issue URL: https://app.asana.com/0/1193060753475688/1198953673392560
CC: @brindy 

**Description**:

This PR fixes two bugs related to the blank snapshot view.

1. The first is where the app will display the blank view when a tel: link is tapped and the system alert is presented.
1. The second is when system-level password UI is presented.

The core issue here was that the blank snapshot view was being displayed when the app has resigned active, which can happen in cases where the app is still in the foreground.

The fix is to move the call into `applicationDidEnterBackground`. This ensures that the snapshot view only appears when the app truly goes into the background, instead of being inactive while in the foreground.

**Steps to test basic behaviour**:

1. Enable Application Lock
1. Background the app
1. Open the app switcher
1. Verify that the app appears blank

**Steps to test tel links**:

1. Enable Application Lock
1. Tap a tel: link in the app ([this article](https://css-tricks.com/the-current-state-of-telephone-links/) contains some)
1. Verify that the phone prompt appears, but that the app state does not change (the website should still be visible)

**Steps to test login prompts**:

1. Enable Application Lock
1. Go to iCloud.com in the app
1. Wait for the system Apple ID login prompt, and verify that the app doesn't show the snapshot view in the background

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [x] iPhone X
* [x] iPad

**OS Testing**:

* [ ] iOS 11
* [ ] iOS 12
* [ ] iOS 13
* [x] iOS 14

---
###### Internal references:

### DIFF
--- a/DuckDuckGo/AppDelegate.swift
+++ b/DuckDuckGo/AppDelegate.swift
@@ -175,10 +175,6 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         showKeyboardIfSettingOn = false
     }
     
-    func applicationWillResignActive(_ application: UIApplication) {
-        displayBlankSnapshotWindow()
-    }
-    
     private func onApplicationLaunch(_ application: UIApplication) {
         beginAuthentication()
         initialiseBackgroundFetch(application)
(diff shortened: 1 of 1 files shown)


brindy (approved):
LGTM!

PR #775, opened 2020-11-12
increase hit areas for toolbar buttons

brindy:
Task/Issue URL: https://app.asana.com/0/392891325557410/1198872510442766
Tech Design URL:
CC:

**Description**:

Use a toolbar subclass to handle hit tests.

Increasing the actual size of the buttons causes the buttons to not be laid out correctly due to the custom views.

Not worrying about iPad layout at this stage.

**Steps to test this PR**:

1. Launch the app on a phone (or iPad with phone layout), ensure tab switcher and bookmarks button work as planned and are not triggered when they shouldn't be (e.g. browsing, scrolling, etc).

**OS Testing**:

* [x] iOS 11
* [x] iOS 12
* [x] iOS 13
* [x] iOS 14

---
###### Internal references:

### DIFF
--- a/DuckDuckGo/GestureToolbarButton.swift
+++ b/DuckDuckGo/GestureToolbarButton.swift
@@ -128,14 +128,7 @@ class GestureToolbarButton: UIView {
     }
 
     override func touchesEnded(_ touches: Set<UITouch>, with event: UIEvent?) {
-        guard let touch = touches.first else {
-            imposeReleaseAnimation()
-            return
-        }
-        guard point(inside: touch.location(in: self), with: event) else {
-            imposeReleaseAnimation()
-            return
-        }
+        imposeReleaseAnimation()
         delegate?.singleTapDetected(in: self)
         imposeReleaseAnimation()
     }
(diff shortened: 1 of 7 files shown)


DuckDuckGo/GestureToolbarButton.swift
```diff
@@ -128,14 +128,7 @@ class GestureToolbarButton: UIView {
...
 
     override func touchesEnded(_ touches: Set<UITouch>, with event: UIEvent?) {
-        guard let touch = touches.first else {
-            imposeReleaseAnimation()
-            return
-        }
-        guard point(inside: touch.location(in: self), with: event) else {
-            imposeReleaseAnimation()
-            return
-        }
```
brindy, line 138: All the above is largely unnecessary - the main scenario it was for is if the user presses down, then moves, then releases, but by that point the long press gesture has already kicked in.

DuckDuckGo/TabSwitcherButton.swift
```diff
@@ -155,9 +155,6 @@ class TabSwitcherButton: UIView {
 
         workItem?.cancel()
         guard workItem != nil else { return }
-
-        guard let touch = touches.first else { return }
-        guard point(inside: touch.location(in: self), with: event) else { return }
```
brindy, line 160: Same as above.

samsymons (approved):
Awesome, this works great. Tested typical usage as well as with VoiceOver on, just to be safe. 👏

brindy:
Thanks @samsymons !

PR #779, opened 2020-11-16
handle launching external apps better

brindy:
Task/Issue URL: https://app.asana.com/0/392891325557410/1198953673392557
Tech Design URL:
CC:

**Description**:

Previously launching an external app would sometimes leave the app in a weird state; a blank screen or with the wrong URL.

Now when clicking a link that opens in an external app we make sure the URL is correct for the current page.  If the launch was triggered from the home screen, we close the tab and go back to the home screen.

**Steps to test this PR**:

From the home screen:
1. Paste https://itunes.apple.com/us/app/duckduckgo 
1. Will redirect to a scheme which launches externally
1. Tab will close back to the home screen
1. Clicking No in the prompt shows the home screen
1. Clicking Yes in the prompt launches the app (or shows the message saying that it couldn't be opened)

From a web page
1. Paste https://itunes.apple.com/us/app/duckduckgo 
1. Will redirect to a scheme which launches externally
1. URL will be updated to reflect the current page, not the pasted link
1. Clicking No in the prompt returns to the page
1. Clicking Yes in the prompt launches the app 

Following links
1. Navigate here https://falkirkrpg.org.uk/noxwall/schemes.html
1. Try the links on the page, ensure browser stays in a reasonable state
1. Some links might result in a search for the URL, that's OK

Errors
1. Paste https://thisisgibberish.url in to the address bar
1. The error screen should display with the url entered
1. Visit https://badssl.com and browse around 
1. The error pages should have the URL of the URL navigated to

Smoke testing
1. Use the browser in a general way and ensure nothing out of order

**OS Testing**:

* [ ] iOS 11
* [ ] iOS 12
* [ ] iOS 13
* [x] iOS 14

---
###### Internal references:

### DIFF
--- a/DuckDuckGo/TabViewController.swift
+++ b/DuckDuckGo/TabViewController.swift
@@ -681,7 +681,7 @@ class TabViewController: UIViewController {
         delegate?.tabLoadingStateDidChange(tab: self)
         UIApplication.shared.open(url, options: [:]) { opened in
             if !opened {
-                self.view.showBottomToast(UserText.failedToOpenExternally)
+                self.view.window?.showBottomToast(UserText.failedToOpenExternally)
             }
 
             // just showing a blank tab at this point, so close it
@@ -984,13 +984,13 @@ extension TabViewController: WKNavigationDelegate {
         lastError = error
         let error = error as NSError
 
-        // prevent loops where a site keeps redirecting to itself (e.g. bbc)
         if let url = url,
             let domain = url.host,
             error.code == Constants.frameLoadInterruptedErrorCode {
+            // prevent loops where a site keeps redirecting to itself (e.g. bbc)
             failingUrls.insert(domain)
 
-            // Reset the URL
+            // Reset the URL, e.g if opened externally
             self.url = webView.url
         }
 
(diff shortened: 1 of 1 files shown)


samsymons:
I ran into an interesting issue testing this (tested on iPad, not sure how this manifests on iPhone):

1. Visit https://falkirkrpg.org.uk/noxwall/schemes.html in the app on an iPad
1. Tap and hold on the iTunes Store `itms://` link to show the preview modal
1. Hit No on the "Open in Another App?" prompt
1. Hit Copy on the preview modal to copy the `itms` link
1. Open a new tab and paste that link in there; it should go to a DuckDuckGo search with that link
1. Go back to the Falkirk RPG page and tap that `itms` link again – instead of prompting you whether you want to open the link in another app, the tab counter in the top right will increment but no new tabs will appear. When you tap on another tab, all of the new App Store tabs will appear, depending on how many times you tapped the `itms` link

No clue at all why this happens, and why it only happens if you've pasted the link in another tab. 🙂 

I'm not even certain that it's caused by the changes in this PR, I can write this up as a bug if it's not worth fixing here (will record and attach a video if so), LMK @brindy.

---

Unrelated to that bug, it looks like the https://itunes.apple.com/us/app/duckduckgo links aren't working. Possibly an error on Apple's end? I get a plaintext error page with this:

```
Your request produced an error. 
[newNullResponse]
```

samsymons (approved):
Besides the issue I mentioned, everything in here looks good. I tested this with a few links that lead to external apps and each worked as expected, and tested with the sites mentioned in the description (thanks for those!).

I'll let you decide what to do about that strange tab quirk @brindy, quite an edge case so no harm done if it gets written up as a LHF.

Note: I wasn’t able to test on iOS 11 and 12 due to an issue with my simulator not launching, but will try again first thing tomorrow if this isn’t merged by then.

brindy:
Thanks @samsymons  - I checked that bug you mentioned and couldn't replicate it so I'm going to merge. If it comes up again we can look at it then.

PR #785, opened 2020-11-20
Prevent tabs from opening all links in the background

samsymons:
Task/Issue URL: https://app.asana.com/0/414709148257752/1199224141509182

**Description**:

This PR fixes a bug where tabs could have their `tapLinkDestination` property get out of sync when the current tab changes.

If you use Command+T to open a new tab, then pressing the Command key puts a tab into the `.backgroundTab` state, and releasing it after the new tab is created means that `currentTab` has changed – the original tab never gets reset to the `.currentTab` state.

The fix is to propagate these events to all active `TabViewController` instances.

**Steps to test this PR**:
1. Open a tab on a device with a physical keyboard, for example the iPad keyboard case
1. Use Command+T to open a new tab
1. Go back to the original tab and tap any link and verify that it does not get opened as a background tab

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPad

**OS Testing**:

* [ ] iOS 11
* [ ] iOS 12
* [ ] iOS 13
* [ ] iOS 14

---
###### Internal references:

### DIFF
--- a/DuckDuckGo/AppDelegate.swift
+++ b/DuckDuckGo/AppDelegate.swift
@@ -33,13 +33,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     private struct ShortcutKey {
         static let clipboard = "com.duckduckgo.mobile.ios.clipboard"
     }
-    
-    static var shared: AppDelegate {
-        // swiftlint:disable force_cast
-        return UIApplication.shared.delegate as! AppDelegate
-        // swiftlint:enable force_cast
-    }
-    
+
     private var testing = false
     var appIsLaunching = false
     var overlayWindow: UIWindow?
--- a/DuckDuckGo/MainViewController.swift
+++ b/DuckDuckGo/MainViewController.swift
@@ -100,7 +100,7 @@ class MainViewController: UIViewController {
         return tabManager?.current
     }
 
-    var tapLinkDestination: TabViewController.LinkDestination = .currentTab
+    var keyModifierFlags: UIKeyModifierFlags?
     
     override func viewDidLoad() {
         super.viewDidLoad()
@@ -386,15 +386,7 @@ class MainViewController: UIViewController {
 
     @available(iOS 13.4, *)
     func handlePressEvent(event: UIPressesEvent?) {
-        self.tapLinkDestination = .currentTab
-
-        if event?.modifierFlags.contains(.command) ?? false {
-            if event?.modifierFlags.contains(.shift) ?? false {
-                self.tapLinkDestination = .newTab
-            } else {
-                self.tapLinkDestination = .backgroundTab
-            }
-        }
+        keyModifierFlags = event?.modifierFlags
     }
 
     override func pressesBegan(_ presses: Set<UIPress>, with event: UIPressesEvent?) {
@@ -1125,8 +1117,8 @@ extension MainViewController: TabDelegate {
         previewsSource.update(preview: preview, forTab: tab.tabModel)
     }
 
-    func tabWillRequestNewTab(_ tab: TabViewController) -> TabViewController.LinkDestination {
-        tapLinkDestination
+    func tabWillRequestNewTab(_ tab: TabViewController) -> UIKeyModifierFlags? {
+        keyModifierFlags
     }
 
     func tabDidRequestNewTab(_ tab: TabViewController) {
--- a/DuckDuckGo/MainViewController.swift
+++ b/DuckDuckGo/MainViewController.swift
@@ -99,6 +99,8 @@ class MainViewController: UIViewController {
     var currentTab: TabViewController? {
         return tabManager?.current
     }
+
+    var tapLinkDestination: TabViewController.LinkDestination = .currentTab
     
     override func viewDidLoad() {
         super.viewDidLoad()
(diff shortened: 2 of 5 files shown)


DuckDuckGo/TabManager.swift
```diff
@@ -25,7 +25,7 @@ class TabManager {
 
     private(set) var model: TabsModel
     
-    private var tabControllerCache = [TabViewController]()
+    private(set) var tabControllerCache = [TabViewController]()
```
samsymons: I originally added a new function which was basically a custom `forEach` on `TabManager` which ran a given block over each `TabViewController`, but that wasn't much different to making this readable publicly. Open to changes here though. 🙂

DuckDuckGo/MainViewController.swift
```diff
@@ -384,12 +384,13 @@ class MainViewController: UIViewController {
...
-        currentTab?.tapLinkDestination = .currentTab
+        tabManager.tabControllerCache.forEach { $0.tapLinkDestination = .currentTab }
+
         if event?.modifierFlags.contains(.command) ?? false {
             if event?.modifierFlags.contains(.shift) ?? false {
-                currentTab?.tapLinkDestination = .newTab
+                tabManager.tabControllerCache.forEach { $0.tapLinkDestination = .newTab }
             } else {
-                currentTab?.tapLinkDestination = .backgroundTab
+                tabManager.tabControllerCache.forEach { $0.tapLinkDestination = .backgroundTab }
```
bwaresiak: As I understand, problem exists for every tab, on which modifier key has ben used. This forcefully resets `tapLinkDestination` on all tabs (so when key is released on a new tab, the "previous current tab" ;) state is fixed).

This seems ok, but since you had to change visibility of `tabControllerCache` and thus kind-of break encapsulation, I wonder if we should pursue a different solution, e.g. returning `tapLinkDestination ` to default in `webView(_ webView:decidePolicyFor:decisionHandler:)` after navigation has been initiated.

bwaresiak: One additional thought - we could even move it a bit further and remove `tapLinkDestination` entirely - keep modifier status in some way in MainVC, as key modifiers to me seem as more app-wide than tab-specific.

samsymons: I like the sound of removing `tapLinkDestination` entirely :+1: You're totally right, because this is something that now has to be changed on every active tab, it doesn't make sense to maintain state individually. I'll get a commit up for this today, thanks!

DuckDuckGo/AppDelegate.swift
```diff
@@ -33,13 +33,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     private struct ShortcutKey {
         static let clipboard = "com.duckduckgo.mobile.ios.clipboard"
     }
-    
-    static var shared: AppDelegate {
-        // swiftlint:disable force_cast
-        return UIApplication.shared.delegate as! AppDelegate
-        // swiftlint:enable force_cast
-    }
```
samsymons, line 41: As a part of exploring different options for keeping the tap link destination property in one place, I noticed that this `shared` property of `AppDelegate` isn't used any more. I've removed it, but let me know if this should stay.

DuckDuckGo/TabDelegate.swift
```diff
@@ -22,6 +22,8 @@ import Core
 
 protocol TabDelegate: class {
 
+    func tabWillRequestNewTab(_ tab: TabViewController) -> TabViewController.LinkDestination
```
samsymons: @bwaresiak This is now handled by having `MainViewController` keep track of the tap link destination state, and `TabViewController` asks for the destination via this new protocol method. I want to avoid having `TabViewController` knowing about `MainViewController` directly, so this seems better even if this function is very simple. 🙂

One possible change to make is to move `LinkDestination` out of `TabViewController`, but I've kept it since that class is the only one which actually uses it to make decisions right now so in my mind it "owns" that enum. It felt more appropriate to keep in there, instead of moving it to `MainViewController`.

(I'm open to different naming here too)

bwaresiak: How about removing the `LinkDestination` enum and replacing that directly with `UIKeyModifierFlags` in MainVC, then move the logic of "unwrapping" flags in the place that is supposed to interpret these (navigation)?

samsymons: That's even better :+1: That approach will mean that `MainViewController` isn't trying to interpret what the flags are supposed to mean itself, and should make it more generalized. Thanks! I'll push a commit shortly.

samsymons:
@bwaresiak This has now been updated to store the modifier flags on `MainViewController`, and pass them to `TabViewController` when it requests them via its delegate. Now the knowledge of which flags are active is stored only in one place, and the logic of deciding what to do with them is done in `TabViewController`. 🙂

DuckDuckGo/TabViewController.swift
```diff
@@ -1046,18 +1036,18 @@ extension TabViewController: WKNavigationDelegate {
...
-                delegate?.tab(self, didRequestNewTabForUrl: url, openedByPage: false)
-                return
-
-            case .backgroundTab:
-                decisionHandler(.cancel)
-                delegate?.tab(self, didRequestNewBackgroundTabForUrl: url)
-                return
-                
-            default: break
+            let modifierFlags = delegate?.tabWillRequestNewTab(self)
```
bwaresiak: I'd put this directly to the 'if' above (line 1038), so we can have 'if let' construct and avoid working with optionals.

bwaresiak (approved):
Good job @samsymons ! Left one more comment but feel free to merge when ready :)

PR #797, opened 2020-12-13
fix long press menu on iOS 12

brindy:
Task/Issue URL: https://app.asana.com/0/414709148257752/1199561262947046
Tech Design URL:
CC:

**Description**:

Fix long press menu on iOS 12 and remove add to reading list to match iOS 13+

**Steps to test this PR**:
1. Using iOS 12, check that long press menu still works (for reason only shows after the long press event), e.g. open in new tab, etc are present
1. Regression test on iOS 13+

**Device Testing**:

* [x] iPhone 8
* [x] iPad

**OS Testing**:

* [x] iOS 12
* [x] iOS 13

---
###### Internal references:

### DIFF
--- a/DuckDuckGo/en.lproj/Localizable.strings
+++ b/DuckDuckGo/en.lproj/Localizable.strings
@@ -37,9 +37,6 @@
 /* Paste and Go action */
 "action.title.pasteAndGo" = "Paste & Go";
 
-/* Reading List action */
-"action.title.readingList" = "Add to Reading List";
-
 /* Refresh action - button shown in alert */
 "action.title.refresh" = "Refresh";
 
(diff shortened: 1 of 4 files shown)


DuckDuckGo/en.lproj/Localizable.strings
```diff
@@ -37,9 +37,6 @@
 /* Paste and Go action */
 "action.title.pasteAndGo" = "Paste & Go";
 
-/* Reading List action */
-"action.title.readingList" = "Add to Reading List";
```
samsymons, line 41: Do you know if the localized versions of this string will be removed automatically the next time we do a localization pass?

samsymons (approved):
Everything looks good, nice job! :+1:

DuckDuckGo/en.lproj/Localizable.strings
brindy: Thanks @samsymons.  We think it will.  This will be a good test anyway.  But I'm going to defer updating the translations until something needs some new ones.

PR #799, opened 2020-12-17
Feature/elle/fire button animations

THISISDINOSAUR:
Task/Issue URL: https://app.asana.com/0/72649045549333/1199145319462726/f
Tech Design URL:
CC: @brindy, @bwaresiak 

**Description**:
Changes the fire animation to use the new fire animation, and adds a settings entry to swap between three different animations (and none).

Design review is done, but product review is ongoing

Things to note are:
- This changes the timing of the various fire button actions (e.g. when the keyboard appears, when the toast appears), so we need to make sure we're happy with that.
- We reverted to an older version of Lottie for performance reasons 
- There's a minor change to how TrackerDataManager and ContentBlockerUserScript interact for performance reasons, we need to ensure that hasn't broken anything
- We've had a lot of back and forth on if the performance is good enough, so need to make sure we are happy with it.

**Steps to test this PR**:
1. Test each new fire animation (including none!), from everywhere it's possible to press the fire button, and check it actually clears what it's supposed to. 
2. Pay particular attention to if the performance is good enough
3. Check that the content blocking still works correctly, and still updates correctly
4. Check any other Lottie animations

**Copy Testing**:

**Orientation Testing**:

* [x] Portrait
* [ ] Landscape

**Device Testing**:

* [x] iPhone SE (1st Gen)
* [x] iPhone 8
* [x] iPhone X
* [x] iPad

**OS Testing**:

* [ ] iOS 11
* [x] iOS 12
* [x] iOS 13
* [x] iOS 14

**Theme Testing**:

* [x] Light theme
* [x] Dark theme

---
###### Internal references:


Core/ContentBlockerUserScript.swift
```diff
@@ -45,8 +45,7 @@ public class ContentBlockerUserScript: NSObject, UserScript {
         let surrogates = storageCache?.fileStore.loadAsString(forConfiguration: .surrogates) ?? ""
 
         // Encode whatever the tracker data manager is using to ensure it's in sync and because we know it will work
-        let encodedTrackerData = try? JSONEncoder().encode(TrackerDataManager.shared.trackerData)
-        let trackerData = String(data: encodedTrackerData!, encoding: .utf8)!
+        let trackerData = TrackerDataManager.shared.encodedTrackerData!
```
THISISDINOSAUR, line 48: This encoding was taking quite a bit of time, so it's now cached.

DuckDuckGo/FireButtonAnimationSettingsViewController.swift
```diff
@@ -0,0 +1,95 @@
...
+
+        cell.accessoryType = animationType == appSettings.currentFireButtonAnimation ? .checkmark : .none
+    }
+    
+    override func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath) {
+        tableView.deselectRow(at: indexPath, animated: true)
+        
+        let type = availableAnimations[indexPath.row]
+        appSettings.currentFireButtonAnimation = type
+        NotificationCenter.default.post(name: AppUserDefaults.Notifications.currentFireButtonAnimationChange, object: nil)
```
bwaresiak: Typically I consider it good practice to specify sender (in this case - self) as as an `object`

DuckDuckGo/FireButtonAnimator.swift
```diff
@@ -0,0 +1,160 @@
...
+        self.appSettings = appSettings
+        reloadPreLoadedComposition()
+                
+        NotificationCenter.default.addObserver(self,
+                                               selector: #selector(onFireButtonAnimationChange),
+                                               name: AppUserDefaults.Notifications.currentFireButtonAnimationChange,
+                                               object: nil)
+    }
+        
+    func animate(animationStartCompletion: @escaping () -> Void, transitionCompletion: @escaping () -> Void, completion: @escaping () -> Void) {
```
bwaresiak: Naming nitpick :) Maybe: `animate(onAnimationStart: @escaping () -> Void, onTransitionCompleted: @escaping () -> Void, completion: @escaping () -> Void)`

bwaresiak (commented):
Great job @THISISDINOSAUR ! 

I've left two comments inline, and I've found that layout is not adjusting to text length in the settings:

cell

Other than that works as expected 👍

DuckDuckGo/FireButtonAnimator.swift
THISISDINOSAUR: Makes sense to me

THISISDINOSAUR:
Thanks, good spot! It's fixed now
Simulator Screen Shot - iPhone 11 - 2020-12-21 at 14 56 53

bwaresiak (approved):
LGTM!

PR #800, opened 2020-12-17
Fix dynamic text on settings headers and footers

THISISDINOSAUR:
Task/Issue URL: https://app.asana.com/0/414709148257752/1199241778510997/f
Tech Design URL:
CC:

**Description**:
Fixes the text sizes of the headers on footers on the GPC settings page, and the Unprotected sites pages

**Steps to test this PR**:
1. Test both pages with a variety of text sizes

**Copy Testing**:

**Orientation Testing**:

* [x] Portrait
* [x] Landscape

**Device Testing**:

* [x] iPhone SE (1st Gen)
* [x] iPhone 8
* [x] iPhone X
* [x] iPad

**OS Testing**:

* [x] iOS 11
* [x] iOS 12
* [x] iOS 13
* [x] iOS 14

**Theme Testing**:

* [x] Light theme
* [x] Dark theme

---
###### Internal references:

### DIFF
--- a/DuckDuckGo/Base.lproj/Settings.storyboard
+++ b/DuckDuckGo/Base.lproj/Settings.storyboard
@@ -1,9 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.CocoaTouch.Storyboard.XIB" version="3.0" toolsVersion="17156" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" colorMatched="YES" initialViewController="w88-u3-TgK">
+<document type="com.apple.InterfaceBuilder3.CocoaTouch.Storyboard.XIB" version="3.0" toolsVersion="17506" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" colorMatched="YES" initialViewController="w88-u3-TgK">
     <device id="retina6_5" orientation="portrait" appearance="light"/>
     <dependencies>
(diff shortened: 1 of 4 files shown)


samsymons (approved):
Looks solid to me! I only ran into a couple tiny issues, neither of which are blockers:

1. The text size doesn't change in the GPC settings if you're already on that screen when Dynamic Type changes. If you back out of it to the main settings screen and then back in again, it's fine, so this is a bit of an edge case – not a blocker. 🙂 
2. I changed the Dynamic Type setting in an iOS 12 simulator while on the main settings screen and it caused a few labels to vanish. I don't think this is the fault of this PR, and it doesn't happen on later iOS versions. Dismissing the settings screen then showing it again fixes it.

Simulator Screen Shot - iPhone 5s - 2020-12-17 at 21 08 24

THISISDINOSAUR:
> Looks solid to me! I only ran into a couple tiny issues, neither of which are blockers:
> 
> 1. The text size doesn't change in the GPC settings if you're already on that screen when Dynamic Type changes. If you back out of it to the main settings screen and then back in again, it's fine, so this is a bit of an edge case – not a blocker. 🙂
> 2. I changed the Dynamic Type setting in an iOS 12 simulator while on the main settings screen and it caused a few labels to vanish. I don't think this is the fault of this PR, and it doesn't happen on later iOS versions. Dismissing the settings screen then showing it again fixes it.
> 

Thanks a lot for the review! 

Re #2 we definitely have a lot of issues around dynamic type in general, I think we have an asana task for it somewhere

Re #1 That's a good point, although in general I'm not sure it's worth worrying about changing dynamic type with the app open, at least not for cases like this where we have to do some manual management of text sizes

PR #804, opened 2021-01-08
Check if problem with Auto Clear overlay occurs and try to fix it if needed

bwaresiak:
Task/Issue URL: https://app.asana.com/0/414709148257752/1199525606019720

**Description**:
In case Blank Overlay has not been dismissed, report that as an error and try to fix this problem.

**Steps to test this PR**:
1. Tweak the code to ensure overlay is not removed (basically return early from `private func removeOverlay()`).
2. Run the app, enable Auto Clear data.
3. Background the app.
4. Put it up again - blank overlay should be displayed.
5. Press any button - that state should be fixed and pixel should be fired.

**Device Testing**:

* [ ] iPhone
* [ ] iPad

**OS Testing**:

* [ ] iOS 11
* [ ] iOS 12
* [ ] iOS 13
* [ ] iOS 14

---
###### Internal references:


DuckDuckGo/BlankSnapshotViewController.swift
```diff
@@ -20,6 +20,11 @@
 import UIKit
 import Core
 
+protocol BlankSnapshotViewRecoveringDelegate: AnyObject {
+    
+    func recoverFromPresenting(controler: BlankSnapshotViewController)
```
samsymons: Minor typo: `controler` ➞ `controller`

samsymons (approved):
Tested this locally and everything works great, nice to have a path for recovery if this state ever happens. 💯 I added one comment about a typo fix, but it's not critical.

Only one question, but not a blocker: since we're only handling the dismiss path on button interactions, do you think it's worth handling it more broadly in the view controller? For example adding a `touchesBegan` implementation/gesture recognizer to `BlankSnapshotViewController` and calling `userInteractionDetected()` inside that, so that this VC is dismissed if the user taps anywhere in the view. Maybe it's not worth it, since users are less likely to tap on the blank part of the screen – handling the buttons is good enough. 🙂

bwaresiak:
@samsymons Thanks! I considered that, but opted for buttons approach in case there is any UI hiccup and overlay removal is done with a delay - figured out it will be harder for buttons approach to cause false positives in such case.

PR #814, opened 2021-01-22
Add background refresh stat safeguards

samsymons:
Task/Issue URL: https://app.asana.com/0/1199333091098016/1199644311779247
CC: @bwaresiak 

**Description**:

This PR introduces a mechanism which checks whether a background task's completion handler has been called previously, and corrects the background refresh metrics if so.

**Steps to test this PR**:
1. Test the background refresh flow on a device through Xcode and verify that the typical flows work as expected; there are instructions inside `AppConfigurationFetch` for manually triggering background tasks

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPad

**OS Testing**:

* [ ] iOS 11
* [ ] iOS 12
* [ ] iOS 13
* [ ] iOS 14

---
###### Internal references:

### DIFF
--- a/DuckDuckGo/AppConfigurationFetch.swift
+++ b/DuckDuckGo/AppConfigurationFetch.swift
@@ -101,7 +101,7 @@ class AppConfigurationFetch {
         case newData
 
         var success: Bool {
-            self == .noData || self == .newData
+            self != .expired
         }
     }
     
--- a/DuckDuckGo/AppConfigurationFetch.swift
+++ b/DuckDuckGo/AppConfigurationFetch.swift
@@ -32,8 +32,11 @@ protocol AppConfigurationFetchStatistics {
     var backgroundStartCount: Int { get set }
     var backgroundNoDataCount: Int { get set }
     var backgroundNewDataCount: Int { get set }
+
+    var backgroundFetchTaskExpirationCount: Int { get set }
 }
 
+// swiftlint:disable type_body_length
 class AppConfigurationFetch {
     
     private struct Constants {
@@ -68,9 +71,6 @@ class AppConfigurationFetch {
     @UserDefaultsWrapper(key: .lastConfigurationRefreshDate, defaultValue: .distantPast)
     static private var lastConfigurationRefreshDate: Date
 
-    @UserDefaultsWrapper(key: .backgroundFetchTaskExpirationCount, defaultValue: 0)
-    static private var backgroundFetchTaskExpirationCount: Int
-
     @UserDefaultsWrapper(key: .backgroundFetchTaskDuration, defaultValue: 0)
     static private var backgroundFetchTaskDuration: Int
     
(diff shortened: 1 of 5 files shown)


DuckDuckGo/AppConfigurationFetch.swift
```diff
@@ -95,6 +94,16 @@ class AppConfigurationFetch {
...
         return Date().timeIntervalSince(Self.lastConfigurationRefreshDate) > Constants.minimumConfigurationRefreshInterval
     }
+
+    enum BackgroundRefreshCompletionStatus {
+        case expired
+        case noData
+        case newData
+
+        var success: Bool {
+            self == .noData || self == .newData
```
Asavarkhul: Could be simplified by `self != .expired` wdyt?

DuckDuckGo/AppConfigurationFetch.swift
```diff
@@ -313,3 +300,78 @@ class AppConfigurationFetch {
...
+                lastCompletionStatus = backgroundRefreshTaskCompletionHandler(store: store,
+                                                                              refreshStartDate: refreshStartDate,
+                                                                              task: task,
+                                                                              status: .expired,
+                                                                              previousStatus: lastCompletionStatus)
+            }
+        }
+
+        queue.async {
+            let newData = configurationFetcher.fetchConfigurationFiles(isBackground: true)
```
Asavarkhul: Since it's a Bool which is returned, why don't naming it like `let newDataExists = ...` ?
Wdyt?

samsymons: Good suggestion!

DuckDuckGo/AppConfigurationFetch.swift
samsymons: Yeah that's a fair suggestion, I wrote it this way originally in order to explicitly state the expected cases, to avoid the situation where new cases added in the future won't impact the logic – let's say a new `.error` case is added, the current implementation would still work as intended, whereas only checking for `.expired` would mean `success` returns true even for an `.error`.

All that being said, this enum is so simple that checking for `expired` makes sense. 🙂

Asavarkhul: Ho yes I agree with your point of view 😏
The only dangerous stuff with computed properties like this is when there is no unit test on it, which could involve one day the typical issue your are mentioning ; So maybe just unit test it? 🤷‍♂️

DuckDuckGo/AppConfigurationFetch.swift
```diff
@@ -68,9 +70,6 @@ class AppConfigurationFetch {
     @UserDefaultsWrapper(key: .lastConfigurationRefreshDate, defaultValue: .distantPast)
     static private var lastConfigurationRefreshDate: Date
 
-    @UserDefaultsWrapper(key: .backgroundFetchTaskExpirationCount, defaultValue: 0)
-    static private var backgroundFetchTaskExpirationCount: Int
```
bwaresiak, line 72: Shouldn't we also remove `backgroundFetchTaskExpirationCount`  Key? This seems to be defined in two places now - as an enum in `UserDefaultsWrapper.Key` and as a static let in `AppUserDefaults.Keys`? Also maybe we should be reusing the old string key as sort of migration?

bwaresiak (approved):
Great job @samsymons, one small comment inline. :)

Asavarkhul (approved):

PR #816, opened 2021-01-25
Enable GPC header adding for select sites

THISISDINOSAUR:
Task/Issue URL: https://app.asana.com/0/414235014887631/1199654549449490/f
Tech Design URL:
CC: @brindy 

**Description**:
Due to an issue with back navigation with Twitter redirects, we disabled our GPC header injection. This adds it back, but only for select sites that support it. This should minimise the effect of the twitter issue, whilst still sending the header to sites that actually support it

**Steps to test this PR**:
1. Navigate to http://global-privacy-control.glitch.me, and check that it says the header is present
2. Manually inspect the webview via safari and check the header is included in the main document request when visiting any NYTimes or Washington Post page.
3. Do the same for any other site besides the three above and check the header isn't present in the request
4. Go to Twitter, and test that when clicking links on twitter to sites other than NYTimes or Washington Post that the back button navigation doesn't have a blank t.co page on the stack
5. Go to settings and turn off the "Global Privacy Control" setting. Ensure the header isn't sent to any site

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPad

**OS Testing**:

* [ ] iOS 11
* [ ] iOS 12
* [ ] iOS 13
* [ ] iOS 14

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:

### DIFF
--- a/DuckDuckGo/TabViewController.swift
+++ b/DuckDuckGo/TabViewController.swift
@@ -1023,7 +1023,9 @@ extension TabViewController: WKNavigationDelegate {
     
     private func requestForDoNotSell(basedOn incomingRequest: URLRequest) -> URLRequest? {
         /*
-         For now, the GPC header is only applied to sites known to be honoring GPC (nytimes.com, washingtonpost.com), while the DOM signal is available to all websites. This is done to avoid an issue with back navigation when adding the header (e.g. with 't.co').
+         For now, the GPC header is only applied to sites known to be honoring GPC (nytimes.com, washingtonpost.com),
+         while the DOM signal is available to all websites.
+         This is done to avoid an issue with back navigation when adding the header (e.g. with 't.co').
          */
         guard let url = incomingRequest.url, appUrls.isGPCEnabled(url: url) else { return nil }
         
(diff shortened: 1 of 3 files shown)


DuckDuckGo/TabViewController.swift
```diff
@@ -1022,10 +1020,42 @@ extension TabViewController: WKNavigationDelegate {
         detectedNewNavigation()
         checkLoginDetectionAfterNavigation()
     }
+    
+    private func requestForDoNotSell(basedOn incomingRequest: URLRequest) -> URLRequest? {
+        // For now, only enable for select sites, due to an issue with back navigation this can introduce
```
THISISDINOSAUR: I think someone mentioned in the Asana comments we should call this out explicitly in a comment. Anything else you think we should add to it?

brindy (approved):
LGTM! Great work, thanks for jumping on this @THISISDINOSAUR !

PR #819, opened 2021-01-31
cancel show keyboard when favorite launched from home screen

brindy:
Task/Issue URL: https://app.asana.com/0/414709148257752/1199613512009688
Tech Design URL:
CC:

**Description**:

Cancel showing keyboard when favorite launched from home screen.

**Steps to test this PR**:
1. Create a favorite
1. Enable "new tab" under Keyboard in settings
1. Use the fire button and wait
1. The keyboard should launch
1. Press cancel
1. Use the fire button but quickly tap the favorite
1. The page should load but the keyboard should not be launched

---
###### Internal references:

### DIFF
--- a/DuckDuckGo/MainViewController.swift
+++ b/DuckDuckGo/MainViewController.swift
@@ -103,6 +103,7 @@ class MainViewController: UIViewController {
     }
 
     var keyModifierFlags: UIKeyModifierFlags?
+    var showKeyboardAfterFireButton: DispatchWorkItem?
     
     // Skip SERP flow (focusing on autocomplete logic) and prepare for new navigation when selecting search bar
     private var skipSERPFlow = true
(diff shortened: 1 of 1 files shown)


samsymons (approved):
Looks great! This is a nice little UX improvement.

This isn't directly related to these changes I don't think, but I did find myself bumping into a timing issue when testing this where if you manage to tap a favourite at the right time after using the fire button, it ends up loading the page while still selecting the URL bar.

Removing the extra 0.25s delay in the `highlightText` check fixes it, but that brings back the original bug that delay is intended to solve. Can't win either way 😩 



https://user-images.githubusercontent.com/183774/106400291-39446e00-63d2-11eb-92a4-46032a7fe9de.mp4

brindy:
As long as this feels like an improvement, we can tweak it more later.  :). Thanks @samsymons

PR #827, opened 2021-02-06
override lookup and open a new tab instead

brindy:
Task/Issue URL: https://app.asana.com/0/392891325557410/1199890390771052
Tech Design URL:
CC:

**Description**:

Override "lookup" in the menu that pops up when long pressing a text selection and do a search in a new tab instead.

**Steps to test this PR**:
1. Open a page
1. Select some text
1. Select lookup from the popup menu
1. New tab should open with selected text as the search

---
###### Internal references:

### DIFF
--- a/DuckDuckGo/DDGWebView.swift
+++ b/DuckDuckGo/DDGWebView.swift
@@ -1,0 +1,43 @@
+//
+//  DDGWebView.swift
+//  DuckDuckGo
+//
+//  Copyright © 2021 DuckDuckGo. All rights reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//  http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


DuckDuckGo/DDGWebView.swift
```diff
@@ -0,0 +1,43 @@
...
+
+import WebKit
+
+class DDGWebView: WKWebView {
+
+    weak var ddgWebViewDelegate: DDGWebViewDelegate?
+
+    // swiftlint:disable identifier_name
+    @IBAction func _lookup(_ sender: Any?) {
+        print("***", #function, sender)
```
samsymons, line 28: Probably worth removing this, or fixing up the compiler warning this throws.

samsymons (approved):
Nice, good work figuring this one out. Didn't see any blockers, just a couple notes:

**1.** Calling this on iframes will return an empty string, even for valid text. This opens a blank search page, which is unintuitive – it would be nice to fall back to the system behaviour, but I'm not sure how viable that is.

**2**. When using Look Up I'd expect it to always open a search tab, though if the string looks like a URL it'll get loaded directly.

This could come up in common use cases like looking up a TV show; I could see anime shows breaking this pretty often, like if you try Look Up on "Aldnoah.Zero" it'll treat it like a domain name and fail to load.

Another case is if the user tries to learn about some popular DNS server like 8.8.8.8, it won't open it as a search.

brindy:
Decided to raise feedback with Apple instead.  Thanks for looking @samsymons !

PR #829, opened 2021-02-11
sync the currently set user interface style to the selected app theme

brindy:
Task/Issue URL: https://app.asana.com/0/414709148257752/1199923842924964
Tech Design URL:
CC:

**Description**:

If the user selects light or dark mode in our app then the web view still selects the system selection, and can cause some dissonance for the user.

**Steps to test this PR**:
1. Load the SERP - it will match the current system theme.
1. Change the in-app theme to `dark`, the SERP will automatically reflect the setting
1. Change the in-app theme to `light`, the SERP will automatically reflect the setting
1. Change the in-app theme to `system`, the SERP will automatically reflect the setting
1. Change the system theme to a different color, the SERP will automatically reflect the setting

**Copy Testing**:

**Orientation Testing**:

* [x] Portrait
* [x] Landscape

**Device Testing**:

* [x] iPhone SE (1st Gen)
* [x] iPhone 8
* [x] iPhone X
* [x] iPad

**OS Testing**:

* [ ] iOS 11
* [ ] iOS 12
* [x] iOS 13
* [x] iOS 14

**Theme Testing**:

* [x] Light theme
* [x] Dark theme

---
###### Internal references:

### DIFF
--- a/DuckDuckGo/AppDelegate.swift
+++ b/DuckDuckGo/AppDelegate.swift
@@ -67,8 +67,6 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
             return true
         }
 
-        updateUserInterfaceStyle()
-
         DispatchQueue.global(qos: .background).async {
             ContentBlockerStringCache.removeLegacyData()
         }
@@ -106,7 +104,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 
     func updateUserInterfaceStyle() {
         guard #available(iOS 13.0, *) else { return }
-        switch ThemeManager.shared.currentTheme.name {
+        switch AppUserDefaults().currentThemeName {
 
         case .dark:
             window?.overrideUserInterfaceStyle = .dark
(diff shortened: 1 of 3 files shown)


samsymons (changes requested):
Found one blocker during testing:

1. Select `System Default` in the app settings
2. Change the OS theme in Control Center
3. The app's theme doesn't change

This appears to be caused by `ThemeManager` never actually returning `systemDefault`, so the `default` case in the `updateUserInterfaceStyle()` switch statement never gets tripped. Here's where `systemDefault` is used to determine whether to return dark or light: https://github.com/duckduckgo/iOS/blob/2827ac7b62959151c276bc3fe75723665d2c1506/DuckDuckGo/ThemeManager.swift#L54-L59

Maybe this could be fixed by adding an explicit check for `AppUserDefaults().currentThemeName == .systemDefault` before the switch statement in `updateUserInterfaceStyle()`, or even using `AppUserDefaults().currentThemeName` itself as the enum that's being switched over?

brindy:
Thanks @samsymons - it was a super quick PR so not surprised! Will fix later today.

samsymons (approved):
Nice, looking perfect now! :shipit:

Issue #806, opened 2021-01-13
Potential memory leak in BloomFilterWrapper


s3cur3:
The `BloomFilterWrapper` doesn't clean up the `BloomFilter` it allocates on init. As best I can tell, this is mostly a theoretical leak—in the app, the `BloomFilterWrapper` is currently retained for the lifetime of the program. If this ever were *not* the case though (as is the case in the test suite), the allocated `BloomFilter` would be leaked.

If you'd like, I can submit a pull request for [the (tiny) fix](https://github.com/s3cur3/iOS/commit/03e97382e28f3da99c4a6b7a4814273c2810d570).

brindy:
Sorry for the delay in replying.  Yes, feel free to submit a PR.  As you say the filter is tied to the the lifetime of the app anyway so no rush.  Thanks!

brindy:
Tracking internally here: https://app.asana.com/0/392891325557410/1199935106721785

samsymons:
Closing this as #833 was merged, thanks @s3cur3!

PR #833, opened 2021-02-14
Fix potential memory leak in BloomFilterWrapper

s3cur3:
**Description**:

As best I can tell, this is mostly theoretical
at this point---the BloomFilterWrapper is currently
retained for the lifetime of the program. If this
ever were *not* the case, though, we'd be leaking
the allocated BloomFilter object without a `delete`
corresponding to its `new`.

**Steps to test this PR**:

You can observe the leak in the test suite, which
*does* create temporary `BloomFilterWrapper`s: just
add a destructor to BloomFilter.hpp in the submodule
like this:

    ~BloomFilter() { printf("Destroyed Bloom\n"); }

...and observe that it's never called without this
patch, and consistently called with it.

Issue URL: https://github.com/duckduckgo/iOS/issues/806
Internal URL:  https://app.asana.com/0/392891325557410/1199935106721785

### DIFF
--- a/Core/BloomFilterWrapper.h
+++ b/Core/BloomFilterWrapper.h
@@ -21,6 +21,7 @@
 @interface BloomFilterWrapper : NSObject
 - (instancetype)initFromPath:(NSString*)path withBitCount:(int)bitCount andTotalItems:(int)totalItems;
 - (instancetype)initWithTotalItems:(int)count errorRate:(double)errorRate;
+- (void)dealloc;
 - (void)add:(NSString*) entry;
 - (BOOL)contains:(NSString*) entry;
 @end
--- a/Core/BloomFilterWrapper.mm
+++ b/Core/BloomFilterWrapper.mm
@@ -44,6 +44,10 @@
     return self;
 }
 
+- (void)dealloc {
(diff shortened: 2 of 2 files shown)


brindy (commented):
Thanks for this! Confirmed this works by using a breakpoint on the unit tests.  Quick question, for completeness is it worth making the filter pointer NULL or is that overkill?

s3cur3:
Not necessary, actually—the C++ spec (and thus Obj-C++ as well) specifies that deleting a null pointer is a no-op. [Relevant StackOverflow discussion here.](https://stackoverflow.com/a/4190737/1417451)

brindy:
Oh, I meant making `filter` NULL after deleting it.  I guess it's low risk though as `dealloc()` would have to be called directly, and that would be madness.  Thanks again! Will merge now.

brindy (approved):
Thanks again!

brindy:
Just wanted to note that this did indeed leak under the right conditions. Specifically when there was data to update from the remote endpoint, the bloom filter is instanciated/replaced so the old one would be hanging around.  So thanks again for this!

PR #837, opened 2021-02-23
Display link titles when sharing via email

samsymons:
Task/Issue URL: https://app.asana.com/0/414709148257752/1199667555122426
Tech Design URL:
CC:

**Description**:

This PR updates `Link` to provide the URL and title/subject to the activity view controller directly, by extending `Link` to conform to `UIActivityItemSource`.

The first commit on this branch achieved the same thing by setting a value for the `Subject` key on the activity view controller, but the current way is the proper way to do it.

**Steps to test this PR**:
1. Open a new tab
1. Go through Settings  -> Share and select the Mail icon
1. Verify that the link's subject is present, matching the title of the tab

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPad

**OS Testing**:

* [ ] iOS 11
* [ ] iOS 12
* [ ] iOS 13
* [ ] iOS 14

---
###### Internal references:

### DIFF
--- a/Core/UIViewControllerExtension.swift
+++ b/Core/UIViewControllerExtension.swift
@@ -51,9 +51,14 @@ extension UIViewController {
         present(controller: shareController, fromButtonItem: buttonItem)
     }
 
-    public func presentShareSheet(withItems activityItems: [Any], fromView sourceView: UIView, atPoint point: Point? = nil) {
+    public func presentShareSheet(withItems activityItems: [Any], fromView sourceView: UIView, subject: String? = nil, atPoint point: Point? = nil) {
         let activities = buildActivities()
         let shareController = UIActivityViewController(activityItems: activityItems, applicationActivities: activities)
+
+        if let subject = subject {
+            shareController.setValue(subject, forKey: "Subject")
+        }
+
         shareController.overrideUserInterfaceStyle()
         shareController.excludedActivityTypes = [.markupAsPDF]
         present(controller: shareController, fromView: sourceView, atPoint: point)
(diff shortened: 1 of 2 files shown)


Core/UIViewControllerExtension.swift
```diff
@@ -74,3 +74,21 @@ extension UIViewController {
...
+    }
+
+    public func activityViewController(_ activityViewController: UIActivityViewController,
+                                       itemForActivityType activityType: UIActivity.ActivityType?) -> Any? {
+        AppUrls().removeATBAndSource(fromUrl: url)
+    }
+
+    public func activityViewController(_ activityViewController: UIActivityViewController,
+                                       subjectForActivityType activityType: UIActivity.ActivityType?) -> String {
+        title ?? ""
```
samsymons: There are other sensible defaults that could be used here. I went with an empty string since that's what we've been doing already, but this could default to the URL itself or some predefined string like "Link shared via DuckDuckGo".

brindy (approved):
LGTM

One thought (we'll discuss later today) is that I think we should standardise on explicit returns from functions, and use implicit from vars, higher order functions or things like Swift UI.  

And just a note that it works fine with Apple Mail, but doesn't work with Fastmail or Outlook.  That's likely as good as can be expected though.

samsymons:
@brindy Yeah I've seen the same thing, in some cases email clients are taking the first line of the body text and treating that as the subject. I'd be okay with prepending the title to the body text to support those cases. 🙂

PR #839, opened 2021-03-05
Fireproofing implementation refactor

samsymons:
Task/Issue URL: https://app.asana.com/0/414709148257752/1199984099899045
Tech Design URL:
CC:

**Description**:

This PR updates the Fireproofing implementation to remove the persistence step during the Fire button process. The new implementation deletes all cached data except for cookies, and then iterates over the cookies to delete those which aren't Fireproof.

**Steps to test this PR**:
1. Test that Fireproofing works

**Device Testing**:

* [x] iPhone SE (1st Gen)
* [x] iPhone 8
* [x] iPhone X
* [x] iPad

**OS Testing**:

* [x] iOS 11
* [x] iOS 12
* [x] iOS 13
* [x] iOS 14

---
###### Internal references:


Core/CookieStorage.swift
```diff
@@ -28,57 +28,13 @@ public class CookieStorage {
...
-
     public init(userDefaults: UserDefaults = UserDefaults.standard) {
         self.userDefaults = userDefaults
     }
 
-    func clear() {
+    public func clear() {
         userDefaults.removeObject(forKey: Constants.key)
         os_log("cleared cookies", log: generalLog, type: .debug)
     }
```
bwaresiak, line 58: I'm all for removing the code - but shouldn't we try to migrate in case cookies are backed in the user defaults?

Core/WebCacheManager.swift
```diff
@@ -101,62 +64,60 @@ public class WebCacheManager {
...
+                for cookie in cookiesToRemove {
+                    group.enter()
+                    cookieStore.delete(cookie) {
+                        group.leave()
+                    }
+                }
+            }
+
+            DispatchQueue.global(qos: .background).async {
+                _ = group.wait(timeout: .now() + 5)
```
bwaresiak: Maybe we should fire debug pixel when it's timing out?

bwaresiak: Another question: is it possible that this executes before 'getAllCookies' returns (and before first 'enter' is called on a group)?

brindy (commented):
I see @bwaresiak has assigned it so I'll let him give the final approval, but I've been running it on my device for the last couple of days and it seems to be working great.

Core/CookieStorage.swift
brindy: So something occurs to me:

1) Migration is unnecessary because we always have a webview, OR 
2) We shouldn't be recreating the webview so quickly after the fire button is used (put a breakpoint on TabViewController.swift:303 and hit the fire button)

Core/WebCacheManager.swift
brindy: Yeah. probably want an `enter()` before calling `getAllCookies` and a `leave()` inside it, assuming that the completion handler for `getAllCookies` always gets called.  If it doesn't ... not sure what you could do here.

Core/CookieStorage.swift
bwaresiak: There is a slight chance cookies list this won't be empty, exactly due to #2 you raised: if cookies were stored with delay and after WebView has been created. It is unlikely, but I still wonder if we should try to preserve those in such case...

Core/WebCacheManager.swift
bwaresiak: We also could move whole `DispatchQueue.global(qos: .background).async` and nest it at the end of `getAllCookies` completion block - also assuming it is being always called. :)

samsymons: I like that suggestion, moving the `DispatchGroup` handling into the `getAllCookies` callback sounds good  – done.

Core/CookieStorage.swift
samsymons: Fair point @bwaresiak, there's certainly a slim chance where cookies have been stored but not yet consumed. I've restored the `consumeCookies` logic to ensure that any stored cookies are being restored, the same way we've always been doing. Some time in the future, we can remove it again.

brindy: I just think it's so edge case it's not worth having it in the code base.  I think we'd be better off addressing the issue that there's a webview constructed during the process of deleting data.

Core/CookieStorage.swift
```diff
@@ -57,28 +57,4 @@ public class CookieStorage {
...
-        var cookieData = [String: Any?]()
-        cookie.properties?.forEach({
-            cookieData[$0.key.rawValue] = $0.value
-        })
-
-        // Same site policy is not returned in the cookie properties so have to use iOS 13 api to get it - setting is fine.  Leaving same site policy
-        //  untouched for previous versions of iOS as changing from Strict to Lax blindly could be a security issue if sites have set it as such.
-        //  See https://www.owasp.org/index.php/SameSite
-        if #available(iOS 13, *) {
-            cookieData[HTTPCookiePropertyKey.sameSitePolicy.rawValue] = cookie.sameSitePolicy ?? HTTPCookieStringPolicy.sameSiteLax
```
brindy, line 72: So ... another problem here is that we make this change deliberately.  It's not quite the right logic though, so I'm just getting confirmation we actually need it

cc @kdzwinel

brindy: I talked to Konrad about this and we're all good given that webkit blocks 3p cookies anyway (so long as the user hasn't disabled that, which we can't do much about)

Core/CookieStorage.swift
bwaresiak: I think in such case we should value potential consistency over cleaning up codebase - we can still remove this migration as a follow up task scheduled few months from now. 

@brindy WebView construction seem to be a separate issue: as far as I see, this is about when to construct WebView for newly created tabs (and cleaning data's side effect is recreating first tab). I think we opted in to do that immediately, as doing it in a lazy manner can result is noticeable lag on first navigation - still, I'm more than happy to discuss that in Asana if you think we should revisit that and it should happen as a separate project.

brindy: Yeah, you're right, let's clean it up in a few months.

Core/WebCacheManager.swift
```diff
@@ -101,62 +103,61 @@ public class WebCacheManager {
...
+
+                for cookie in cookiesToRemove {
+                    group.enter()
+                    cookieStore.delete(cookie) {
+                        group.leave()
+                    }
+                }
+
+                DispatchQueue.global(qos: .background).async {
+                    _ = group.wait(timeout: .now() + 5)
```
bwaresiak: Wasn't we supposed to use `cookieDeletionTimedOut` here? :)

Core/WebCacheManager.swift
```diff
@@ -101,62 +103,61 @@ public class WebCacheManager {
...
+
+                    Pixel.fire(pixel: .blankOverlayNotDismissed)
```
bwaresiak: And this piece of code is pretty surprising :)

samsymons: Hmm wtf happened here 🤔

Core/WebCacheManager.swift
samsymons: Correct pixel is being used here now 🙂

Core/WebCacheManager.swift
```diff
@@ -101,62 +103,61 @@ public class WebCacheManager {
...
+                for cookie in cookiesToRemove {
+                    group.enter()
+                    cookieStore.delete(cookie) {
+                        group.leave()
+                    }
+                }
+
+                DispatchQueue.global(qos: .background).async {
+                    _ = group.wait(timeout: .now() + 5)
+                    Pixel.fire(pixel: .cookieDeletionTimedOut)
```
bwaresiak: But aren't we supposed to only fire that when there is a timeout? :)

samsymons: D'oh, yes sorry, am not able to focus properly right now. 🙂 I'll change this to check against the timeout result.

samsymons: Done

bwaresiak (approved):
Good job @samsymons !

PR #854, opened 2021-03-31
Switch from beforeload to load event handler

SlayterDev:
Task/Issue URL: https://app.asana.com/0/414709148257752/1200134317948619/f
Tech Design URL: N/A
CC: @bwaresiak @catehstn 

**Description**:
This PR fixes surrogate loading after `beforeload` was deprecated in iOS 14. This solution has been tested to work with iOS 14 and below.

**Steps to test this PR**:
1. Add log or breakpoint to this line to show when a surrogate is loaded: https://github.com/duckduckgo/iOS/blob/714450d4c59c604eb41150addda676f088beb814/Core/contentblocker.js#L490
2. Visit https://nytimes.com/wirecutter/deals/
1. Your breakpoint should trigger or log should be displayed

**Device Testing**:

* [x] iPhone SE (1st Gen)
* [x] iPhone 8
* [x] iPhone X
* [x] iPad

**OS Testing**:

* [x] iOS 11
* [x] iOS 12
* [x] iOS 13
* [x] iOS 14

---
###### Internal references:

### DIFF
--- a/Core/contentblocker.js
+++ b/Core/contentblocker.js
@@ -484,6 +484,8 @@ _utf8_encode : function (string) {
             return new URL(location.href)
         }
     }
+    
+    var loadedSurrogates = {}
 
     // private
     function loadSurrogate(surrogatePattern) {
@@ -531,7 +533,10 @@ _utf8_encode : function (string) {
         // Tracker blocking is dealt with by content rules
         // Only handle surrogates here
         if (blocked && result.matchedRule && result.matchedRule.surrogate) {
-            loadSurrogate(result.matchedRule.surrogate)
+            if (!loadedSurrogates[trackerUrl]) {
+                loadSurrogate(result.matchedRule.surrogate)
+                loadedSurrogates[trackerUrl] = true
+            }
 
             duckduckgoDebugMessaging.signpostEvent({event: "Tracker Blocked",
                                                    url: trackerUrl,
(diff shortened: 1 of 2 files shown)


Core/contentblockerrules.js
```diff
@@ -112,31 +112,18 @@
...
-        }
-
-        sendMessage(event.url, type)
-    }, false)
-
-
     function onLoadNativeCallback() {
       // Send back the sources of every script and image in the DOM back to the host application.
       [].slice.apply(document.scripts).forEach(function(el) { sendMessage(el.src, "script"); });
+      [].slice.apply(document.links).forEach(function(el) { sendMessage(el.src, "link"); });
```
kdzwinel: wait, those are links as in `<a>` not as in `<link>`. Are we sure we have the right thing here?

Core/contentblocker.js
```diff
@@ -553,29 +553,39 @@ _utf8_encode : function (string) {
...
-                duckduckgoDebugMessaging.log("blocking beforeload")
-                event.preventDefault()
-                event.stopPropagation()
-            } else {
-                duckduckgoDebugMessaging.log("don't block " + event.url);
-                return
-            }
-        }, true)
+        duckduckgoDebugMessaging.log("installing load detection")
+        window.addEventListener("load", function(event) {
```
kdzwinel, line 609: this means that we will get one snapshot and won't see other stuff loaded dynamically after page has loaded, right?

Maybe we should should check for changes every X instead?

SlayterDev: I think this is ok because there are other hooks to check for other content being loaded.

Core/contentblockerrules.js
SlayterDev: Ughhhh no. I'll fix.

Core/contentblocker.js
```diff
@@ -553,29 +553,39 @@ _utf8_encode : function (string) {
...
+            [].slice.apply(document.scripts).forEach(function(el) {
```
jonathanKingston: Does Safari support `[...document.scripts]` here?

SlayterDev: Not sure. This is in line with other parts of this codebase though. Bringing the style of the JS up to standard is out of scope here though and has been planned as a future project.

Core/contentblocker.js
```diff
@@ -553,29 +553,47 @@ _utf8_encode : function (string) {
...
+                if (shouldBlock(el.src, 'SCRIPT')) {
+                    duckduckgoDebugMessaging.log("blocking load")
```
jonathanKingston: As the load events happen this will continue logging for each load right which is much more than before.

SlayterDev: Likely. But does not appear to have any detrimental side effects in my testing. There's already a lot of logging going on.

Core/contentblocker.js
SlayterDev: Put this on an interval of x * 2 with x = 2 to start per conversation with @bwaresiak

Core/contentblocker.js
```diff
@@ -549,33 +554,61 @@ _utf8_encode : function (string) {
...
+              return
+          }
+        });
+        scheduleProcessPage()
+    }
+    
+    var interval = 1
+    function scheduleProcessPage() {
+        interval *= 2
+        setTimeout(processPage, interval * 1000)
```
jonathanKingston, line 602: Doesn't this need to debounce to limit the number of times this gets called?

jonathanKingston (approved):
LGTM on cursory glance.

Core/contentblocker.js
jonathanKingston: Ignore this, I was reading the code wrong. Thanks @SlayterDev for explaining :)

PR #856, opened 2021-04-03
Add a debug pixel to check the tab preview count

samsymons:
Task/Issue URL: https://app.asana.com/0/414235014887631/1200123805767782/f
Tech Design URL:
CC: @bwaresiak 

**Description**:

This PR adds a debug pixel to check the stored tab preview count vs the total tab count, reporting the pixel if the preview count is higher than the tab count.

One question about this change is: are there any situations where we would expect the tab preview count to be higher, even briefly? In the implementation I've added the pixel check after tab previews and the tab model are updated, but I want to be certain that we won't fire false positives.

**Steps to test this PR**:
1. Open and close tabs, checking that the pixel check is performed (but that the pixel does not fire)
1. Check that there are no other places where tab previews are stored, so that we're sure the pixel is checking all cases
1. Background and foreground the app, checking that the pixel check is performed (but that the pixel does not fire)

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPad

**OS Testing**:

* [ ] iOS 11
* [ ] iOS 12
* [ ] iOS 13
* [ ] iOS 14

---
###### Internal references:

### DIFF
--- a/DuckDuckGo/TabManager.swift
+++ b/DuckDuckGo/TabManager.swift
@@ -44,6 +44,10 @@ class TabManager {
         registerForNotifications()
     }
 
+    deinit {
+        NotificationCenter.default.removeObserver(self)
+    }
+
     private func buildController(forTab tab: Tab) -> TabViewController {
         let url = tab.link?.url
         return buildController(forTab: tab, url: url)
--- a/DuckDuckGo/TabManager.swift
+++ b/DuckDuckGo/TabManager.swift
@@ -40,6 +40,8 @@ class TabManager {
             let controller = buildController(forTab: tab)
             tabControllerCache.append(controller)
         }
+
+        registerForNotifications()
     }
 
     private func buildController(forTab tab: Tab) -> TabViewController {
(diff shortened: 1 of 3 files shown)


DuckDuckGo/TabManager.swift
```diff
@@ -224,6 +224,18 @@ class TabManager {
 
     func save() {
         model.save()
+        assertTabPreviewCount()
```
bwaresiak, line 256: The only concern I have is how often this is called (especially when loading the page). This is simple, as 'save' is a single point to catch all model changes - which is very good, but on the other hand not all of the changes result in preview creation/deletion causing potentially unnecessary checks..

brindy:
Anything happening with this PR? It's nearly a week since the last update.

samsymons:
@brindy Thanks for the reminder, Bartek and I chatted about a solution that doesn't check on every call to `save()` so I've now pushed a commit to make this more infrequent by only checking when the app becomes active. As long as the app is checking this sometimes, I'm happy – there's no reason it has to check every time data changes. 🙂

bwaresiak (approved):
LGTM!

PR #860, opened 2021-04-15
Fire education iteration

THISISDINOSAUR:
Task/Issue URL: https://app.asana.com/0/72649045549333/1179359141574181/f
Tech Design URL:
CC: @cmonfortep 

**Description**:
iOS implementation of Android Fire Button education.
We originally tried an iteration of the android one, but the result wasn't statistically significant. So this instead implements the same flow as on Android (which saw a statistically significant retention lift).

This PR also removes the previous experiment and some logic that is no longer needed in FullScreenDaxDialogViewController

Now instead of showing the fire education dialog directly, it makes the fire button pulse (for one hour, or until it's tapped on or a tab is opened).
If the user presses the fire button whilst it's pulsing, then the fire education dialog is shown.
The logic is quite complicated, so LMK if you have any questions.

Things to note: 
- Pulse should start showing at generally the same place in the flow that the previous fire education started showing
- It should time out after an hour
- It should stop showing if the user opens a new tab
- Other actions (such as opening the settings menu) shouldn't cancel the pulse
- if it's within the hour window and you close the app and open it again, the pulse should show.
- The top and bottom bars shouldn't hide when scrolling if the pulse is visible, but should behaviour normally when it's not visible.
- If you press the fire button on the tab switcher screen, the dialog should show if the pulse is active.  Android does show the dialog here, but not the pulse because it was more complicated. On iOS it was easier to make it pulse here too and we decided we'd prefer that.

**Steps to test this PR**:
1. Test experiment is bucketed correctly (I.e. test that you see the fire education roughly 50% of the time)
2. Test nothing has changed when the experiment isn't active
3. Test favourites widget onboarding still works correctly
4. Test fire education dialog shows at the correct time. The pulse should start showing at the same time as the old fire dialog showed.
5. Test that hiding dax tips before seeing the fire button education works (i.e. fire button education shouldn't be shown)
6. Test pixels are fired when expected
7. Test the things in "Things to note" above

**Copy Testing**:

**Orientation Testing**:

* [x] Portrait
* [x] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [x] iPhone 8
* [x] iPhone X
* [x] iPad

**OS Testing**:

* [x] iOS 11
* [x] iOS 12
* [x] iOS 13
* [x] iOS 14

**Theme Testing**:

* [x] Light theme
* [x] Dark theme

---
###### Internal references:


DuckDuckGo/MainViewController.swift
```diff
@@ -782,6 +814,10 @@ class MainViewController: UIViewController {
     }
 
     func newTab(reuseExisting: Bool = false) {
+        if DaxDialogs.shared.shouldShowFireButtonPulse {
+            ViewHighlighter.hideAll()
+        }
+        DaxDialogs.shared.fireButtonPulseCancelled()
```
brindy, line 825: Is any new tab a reason to cancel the pulse, or some specific path.  Reason I ask is that you can also open a new tab from the web page content menu.

Core/Pixel.swift
```diff
@@ -126,14 +126,17 @@ public enum PixelName: String {
...
-    case daxDialogsFireEducationFireButton = "m_dx_fe_f"
-    case daxDialogsHiddenBeforeFireEducation = "m_dx_h_bfe"
-
+    case daxDialogsFireEducationShown = "m_dx_fe_s"
+    case daxDialogsFireEducationConfirmed = "m_dx_fe_co"
+    case daxDialogsFireEducationCancelled = "m_dx_fe_ca"
+    case daxDialogsFireEducationCancelledOutsideOfButton = "m_dx_fe_ca_o"
+    case fireEducationPulseShown = "m_fe_p_s"
+    case fireEducationPulseCancelledBecauseTabOpened = "m_fe_p_c_ta"
+    case fireEducationPulseCancelledBecauseTimeout = "m_fe_p_c_to"
```
brindy, line 136: ❗  Did not see this one

DuckDuckGo/DaxDialogs.swift
```diff
@@ -166,14 +177,46 @@ class DaxDialogs {
...
+            Pixel.fire(pixel: .fireEducationPulseShown)
+        }
+        if fireButtonPulseTimer == nil, let date = settings.fireButtonPulseDateShown {
+            let timeSinceShown = Date().timeIntervalSince(date)
+            let timerTime = DaxDialogs.timeToFireButtonExpire - timeSinceShown
+            fireButtonPulseTimer = Timer(timeInterval: timerTime, repeats: false) { _ in
+                self.settings.fireButtonEducationShownOrExpired = true
+                ViewHighlighter.hideAll()
+                if !self.settings.fireButtonEducationShownOrExpired {
+                    Pixel.fire(pixel: .fireEducationPulseCancelledBecauseTimeout)
```
brindy, line 194: I think this didn't get called because you've already set the setting in the lines above.

brindy (commented):
Looks great, I'm excited for this because the previous fire button thing I found to be really disruptive.    Couple of comments inline - in particular one of the pixels wasn't fired.  I've tested it on an iPhone 12.  About to switch to iPad but wanted to let you have this feedback ASAP.

DuckDuckGo/DaxDialogs.swift
THISISDINOSAUR: Ah thanks, good spot! I fixed something else and in the process broke this. Will fix

DuckDuckGo/MainViewController.swift
THISISDINOSAUR: Any new tab. My assumption was that this is called for any path that opens a new tab, is that correct (it seemed to be, but I might have missed something)?

brindy: Yeah, there's the new tab stuff in the TabDelegate, I think they all ultimately go through the `loadUrlInNewTab` function  in MainViewController

brindy (commented):
I can see on BitRise there's a failing unit test when merged with develop and confirmed it locally.

brindy (commented):
iPad is generally looking good except for one edge case that I've been able to find:

* Get the pulse going
* Launch another app as the main app
* Drag our app so it's "over" 

Switching between split screen modes seems to work fine though. 

Simulator Screen Shot - iPad Pro (12 9-inch) (4th generation) - 2021-04-16 at 14 24 39

brindy (approved):
LGTM! Great work :)

PR #861, opened 2021-04-16
Prevent sharing two links from the Bookmarks share button

samsymons:
Task/Issue URL: https://app.asana.com/0/414235014887631/1200190888218943/f
Tech Design URL:
CC:

**Description**:

This PR fixes a bug where the Share button in the Bookmarks view controller would share the same link twice.

This began happening after `Core.Link` was made to conform to `UIActivityItemSource`, at which point it was capable of publishing its own URL directly.

**Steps to test this PR**:
1. Favorite or bookmark a page
1. Go the bookmarks screen and swipe right on a row to reveal the Share button
1. Tap Share and verify that only one link is present, and that sharing works as expected to various sources (iMessage, Mail, AirDrop, etc.)
1. Verify that the shared link has removed ATB and source parameters

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPad

**OS Testing**:

* [ ] iOS 11
* [ ] iOS 12
* [ ] iOS 13
* [ ] iOS 14

---
###### Internal references:

### DIFF
--- a/DuckDuckGo/BookmarksViewController.swift
+++ b/DuckDuckGo/BookmarksViewController.swift
@@ -191,9 +191,7 @@ class BookmarksViewController: UITableViewController {
     fileprivate func showShareSheet(for indexPath: IndexPath) {
 
         if let link = currentDataSource.link(at: indexPath) {
-            let appUrls: AppUrls = AppUrls()
-            let url = appUrls.removeATBAndSource(fromUrl: link.url)
-            presentShareSheet(withItems: [ link ], fromView: self.view)
+            presentShareSheet(withItems: [link], fromView: self.view)
         } else {
             os_log("Invalid share link found", log: generalLog, type: .debug)
         }
--- a/DuckDuckGo/BookmarksViewController.swift
+++ b/DuckDuckGo/BookmarksViewController.swift
@@ -193,7 +193,7 @@ class BookmarksViewController: UITableViewController {
         if let link = currentDataSource.link(at: indexPath) {
             let appUrls: AppUrls = AppUrls()
             let url = appUrls.removeATBAndSource(fromUrl: link.url)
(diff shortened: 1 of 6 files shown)


DuckDuckGo/BookmarksViewController.swift
```diff
@@ -191,9 +191,7 @@ class BookmarksViewController: UITableViewController {
     fileprivate func showShareSheet(for indexPath: IndexPath) {
 
         if let link = currentDataSource.link(at: indexPath) {
-            let appUrls: AppUrls = AppUrls()
-            let url = appUrls.removeATBAndSource(fromUrl: link.url)
-            presentShareSheet(withItems: [ url, link ], fromView: self.view)
+            presentShareSheet(withItems: [link], fromView: self.view)
```
samsymons, line 194: The `Link` instance is calling `removeATBAndSource` itself, so we no longer need to do that here. 

https://github.com/duckduckgo/iOS/blob/develop/Core/UIViewControllerExtension.swift#L86

bwaresiak: Yes, that's right, but testing that made me realize we have another bug :) 

For sharing search URLs, can you make it so `public func removeATBAndSource(fromUrl url: URL) -> URL`  also removes `Param.searchHeader` (ko) otherwise search bar won't show in other browsers.

samsymons: @bwaresiak Great find, fixed and updated tests. I renamed `removeATBAndSource` to `removeInternalSearchParameters`, happy to change this to something else though. 🙂

bwaresiak (approved):
LGTM!

PR #881, opened 2021-05-12
Correct the memory management of the launchTabObserver

ejensen:
**Asana:** https://app.asana.com/0/414235014887631/1200335662767805/f

The observer reference should be strong, while the capture of self within the closure should be weak

**Description**:

I found this issue when the memory profiler reported a leak on app start. The `LaunchTabNotification.Observer` object is being immediately deallocated since the property was marked as `weak` and there is nothing else retaining the object. Without an observer reference to unregister the observation will never stop - keeping the view controller alive forever.

memory-leak-profiler

This is technically a memory leak, but since the `MainViewController` has a lifecycle tied to the app's lifecycle, it's not a major issue. 

The leak is fixed by following the [guidelines from the `NotificationCenter.addObserver(forName:object:queue:using:)`](https://developer.apple.com/documentation/foundation/notificationcenter/1411723-addobserver#discussion)

> To avoid a retain cycle, use a weak reference to `self` inside the block when `self` contains the observer as a strong reference.
>
> You must invoke `removeObserver(_:)` or `removeObserver(_:name:object:)` before the system deallocates any object that `addObserver(forName:object:queue:using:)` specifies.

**Steps to test this PR**:
1. Start a memory leak profiling session
2. Launch the app with the profiler attached
3. Test that the memory leak caused by the observer is gone
4. Test that the launch tab observer still functions correctly (ex opening the ToS;DR link)

**OS Testing**:

* [x] iOS 13
* [x] iOS 14

---
###### Internal references:

### DIFF
--- a/DuckDuckGo/LaunchTabNotification.swift
+++ b/DuckDuckGo/LaunchTabNotification.swift
@@ -29,6 +29,10 @@ class LaunchTabNotification {
             self.observer = observer
         }
 
+        deinit {
+            remove()
+        }
+
         func remove() {
             NotificationCenter.default.removeObserver(observer)
         }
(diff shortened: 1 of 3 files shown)


samsymons (approved):
LGTM, thanks @ejensen! 👏

Issue #847, opened 2021-03-17
Excess keyboard spacing


aigarssilavs:
Unnecessary bottom content inset is applied to the web page when keyboard is presented.

**App version:** 7.62.1.0
**iOS version:** 14.2
**Device:** iPhone X
**Steps to reproduce the bug:** Select any textbox in any web page and scroll to the bottom of the page
**Expected behavior:** Scrolling stops when the bottom of the web page is reached
**Actual behavior:** Scrolling beyond web page content borders is allowed 

![ExcessKeyboardSpacing](https://user-images.githubusercontent.com/1329970/111471835-fea05600-8731-11eb-9050-da919450a65a.gif)

brindy:
Thanks for this! If it's convenient and makes sense can you add a screenshot, please?

aigarssilavs:
No problem. Here are the screenshots for of the same page as in the .gif in the original comment.

Keyboard dismisssed and view scrolled to the bottom:
![KeyboardDismissed](https://user-images.githubusercontent.com/1329970/111502962-002c4700-874f-11eb-9d20-e2ffef84ef9e.jpg)

Keyboard visible and view scrolled to the bottom:
![KeyboardVisible](https://user-images.githubusercontent.com/1329970/111502949-fc98c000-874e-11eb-939b-62140178a238.PNG)

untalfranfernandez:
I've been checking this issue and it has to be related somehow with [this content offset change](https://github.com/duckduckgo/iOS/blob/develop/DuckDuckGo/MainViewController.swift#L234) when the keyboard is displayed. I've done some quick tests with different webpages removing that line and it seems to be working fine.

I'd propose a PR for that, but [that change](https://github.com/duckduckgo/iOS/pull/411) was introduced two years ago and I have the feeling of being missing something about it.

brindy:
Thanks for diving in @untalfranfernandez

I guess that change was also made for the find in page UI we have.  If everything seems fine on iPhone 8, iPhone X and iPads without that line feel free to open a PR. :)

brindy:
Hi, I don't think we're going to address this right now as it's generally not a big issue for most users. If you feel like having another shot in the future, please feel free to reopen or create a new issue with your PR.   Thanks again for getting in touch!

PR #885, opened 2021-05-18
Remove unnecessary content offset for a webview when keyboard is displayed.

untalfranfernandez:
Task/Issue URL: 

* https://github.com/duckduckgo/iOS/issues/847
* https://app.asana.com/0/414709148257752/1200348293580049

Tech Design URL: 
CC:

**Description**:

There is an unnecessary bottom space added to any webview when the keyboard is displayed due to any textfield getting focused.

**Steps to test this PR**:
1. Open `https://nos.nl/zoeken`
2. Tap on the main textfield available.
3. Scroll the webview to the bottom.

**Copy Testing**:

**Orientation Testing**:

* [x] Portrait
* [x] Landscape

**Device Testing**:

* [x] iPhone SE (1st Gen)
* [x ] iPhone 8
* [x] iPhone X
* [x] iPad

**OS Testing**:

* [ ] iOS 13
* [x] iOS 14

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:

### DIFF
--- a/DuckDuckGo/Base.lproj/Main.storyboard
+++ b/DuckDuckGo/Base.lproj/Main.storyboard
@@ -1,9 +1,9 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<document type="com.apple.InterfaceBuilder3.CocoaTouch.Storyboard.XIB" version="3.0" toolsVersion="17701" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" useSafeAreas="YES" colorMatched="YES" initialViewController="C1X-Zd-UbA">
+<document type="com.apple.InterfaceBuilder3.CocoaTouch.Storyboard.XIB" version="3.0" toolsVersion="18122" targetRuntime="iOS.CocoaTouch" propertyAccessControl="none" useAutolayout="YES" useTraitCollections="YES" useSafeAreas="YES" colorMatched="YES" initialViewController="C1X-Zd-UbA">
     <device id="retina6_1" orientation="portrait" appearance="light"/>
     <dependencies>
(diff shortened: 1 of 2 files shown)


brindy (changes requested):
Hi - thanks for this, but it doesn't seem to work well with find in page.  On the page you suggested if you open find in page and search for NOS you can't see the items under the find in page UI / keyboard.   Any ideas?

untalfranfernandez:
Hi @brindy I came up with an improved solution that works on that scenario you've described.

Now, the container has a vertical spacing constraint with the find-in-page view, so the size for the container gets automatically reduced when this view is shown. 

Screen Shot 2021-05-24 at 19 50 35

There is one scenario pending to review, which was failing previously and now is mitigated but far from being fixed: 
If you trigger the "find-in-page" feature and without dismissing it, tap on any textfield in the website display, there is a mismatch between the sizes of `WKContentView` and the `WKScrollView`

I have some ideas about how to fix this, but those imply deeper changes that could be discussed in another PR.

brindy:
Hi - sorry for the delay in reply and thanks for sticking with this. 

I've had a think about this, and I don't want us to head in this direction because we have a bug related to resizing the web views.   You may have noticed that some pages change when the bars appear / disappear and also that the previous page's position is lost when you go back, this is all due to resizing the web view.

I think approach would likely compound that problem making it harder for us to do a fix in the future.  

I think in fact that we need to solve the web view resizing problem first before addressing the problem this PR fixes. 

Any thoughts on that?

Thanks

untalfranfernandez:
Hey, I do agree with you on that, so, we can close this PR and maybe give it another go at any given point in future.

brindy:
Thanks @untalfranfernandez !

PR #888, opened 2021-05-19
support future nav property in autocomplete results

brindy:
Task/Issue URL: https://app.asana.com/0/230839424936/1200348293580045
Tech Design URL:
CC:  @gtfiorentino

**Description**:

Changes the parsing of the autocomplete results to not break if unexpected properties are presented and also supports upcoming nav property.

**Steps to test this PR**:

1. Type something in to the address bar.  Only suggestions with isNav: true should now be navigational.  `healthcare.gov` is a good search.

Note: assumption is that this will be released on the server later today.  If not, update the in app end point to point at a dev instance (see task above).

***General***

1. Check that searching for bookmarks via the address bar still works as expected.

**Device Testing**:
* [ ] iPhone X
* [ ] iPad

**OS Testing**:

* [ ] iOS 13
* [ ] iOS 14

---
###### Internal references:


DuckDuckGo/AutocompleteViewController.swift
```diff
@@ -53,7 +53,11 @@ class AutocompleteViewController: UIViewController {
             view.backgroundColor = showBackground ? UIColor.black.withAlphaComponent(0.2) : UIColor.clear
         }
     }
-    
+
+    var selectedSuggestion: Suggestion? {
+        (selectedItem >= 0 && selectedItem < suggestions.count) ? suggestions[selectedItem] : nil
```
samsymons: One neat trick I learned recently was to use `indices` to do bounds checking, like `suggestions.indices.contains(selectedItem)`. IMO it's easier to read than comparison statements.

I've been thinking about adding an extension to `Collection` to do safe lookups, like:

```
subscript (safe index: Index) -> Element? {
    return indices.contains(index) ? self[index] : nil
}
```

None of this is a blocker, just thinking out loud. 🙂

samsymons (approved):
Awesome, really impressed by how quickly this change was proposed and implemented. 🎉 Everything looked good to me during testing.

DuckDuckGo/AutocompleteViewController.swift
```diff
@@ -53,7 +53,12 @@ class AutocompleteViewController: UIViewController {
             view.backgroundColor = showBackground ? UIColor.black.withAlphaComponent(0.2) : UIColor.clear
         }
     }
-    
+
+    var selectedSuggestion: Suggestion? {
+        let state = (suggestions: self.suggestions, selectedIndex: self.selectedItem)
```
brindy, line 58: This should hopefully make this a bit more thread safe.

DuckDuckGo/AutocompleteViewController.swift
```diff
@@ -53,7 +53,12 @@ class AutocompleteViewController: UIViewController {
...
+        return state.suggestions.indices.contains(state.selectedIndex) ? state.suggestions[state.selectedIndex] : nil
+    }
```
brindy, line 60: Good suggestion to check indices property

PR #891, opened 2021-05-27
Fix repeated webview refreshes

samsymons:
Task/Issue URL: https://app.asana.com/0/414235014887631/1200384654098883/f
Tech Design URL:
CC: @brindy

**Description**:

This PR adds a check to the content block rules manager to prevent it from compiling rules multiple times simultaneously.

**Steps to test this PR**:

The best way to test this is to reproduce the bug on `develop` and then run the same steps on this branch and compare.

To reproduce this on `develop`:

1. Update `etag(for list: ContentBlockerRequest.Configuration)` in `UserDefaultsETagStorage` to return nil every time, to force the app to refresh even if data has been downloaded prior
1. Update `shouldRefresh` in `AppConfigurationFetch` to always return `true`

From here, launch the app and open a number of tabs. Re-launch the app and quickly go to 3-4 tabs via the tab switcher and then wait on the last one. Every time you open a new tab when new rules have been downloaded but compilation hasn't completed will trigger the start of a new compilation run asynchronously, and they will all finish in sequence at the end, trigger many refreshes in a row.

Once you've done that, swap to this branch and do the same thing.

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPad

**OS Testing**:

* [ ] iOS 13
* [ ] iOS 14

---
###### Internal references:

### DIFF
--- a/Core/ContentBlockerRulesManager.swift
+++ b/Core/ContentBlockerRulesManager.swift
@@ -30,15 +30,13 @@ public class ContentBlockerRulesManager {
 
     private init() {}
 
-    private var isCompilingRules: Bool = false
+    private var isCompilingRules = false
 
     public func recompile() {
         guard let store = WKContentRuleListStore.default() else {
             fatalError("Failed to access the default WKContentRuleListStore")
         }
 
-        // The `compiledRules` function has this check internally, but it needs to be checked here so that `removeContentRuleList` doesn't get
-        // called accidentally.
         guard !isCompilingRules else {
             return
         }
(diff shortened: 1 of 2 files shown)


Core/ContentBlockerRulesManager.swift
```diff
@@ -48,6 +56,12 @@ public class ContentBlockerRulesManager {
 
     /// Return compiled rules for the current content blocking configuration.  This may return a precompiled rule set.
     public func compiledRules(completion: ((WKContentRuleList?) -> Void)?) {
+        guard !isCompilingRules else {
```
samsymons, line 57: One thing I'm worried about here is this scenario:

1. App launches for the very first time, and starts compiling the tracker data bundled with the app
2. App downloads the new tracker data
3. App tries to compile the freshly downloaded data, but the first compilation is still running so it hits the early return

@brindy Not sure about the best way to solve for that, what do you think? I'm not _too_ concerned about it since we have a larger refactor to this area of the codebase coming soon.

brindy (approved):
Nice work @samsymons !

PR #894, opened 2021-06-02
Replace lookup button with SERP based definition

brindy:
Task/Issue URL: https://app.asana.com/0/42792087274227/1200417326972258
Tech Design URL:
CC:

**Description**:

Replace lookup button with SERP based definition

**Steps to test this PR**:

1.   Open a page, select some text when the menu appears choose lookup.

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [x] iPhone X
* [x] iPad

**OS Testing**:

* [x] iOS 13
* [x] iOS 14

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:

### DIFF
--- a/Core/WebView.swift
+++ b/Core/WebView.swift
@@ -1,0 +1,41 @@
+//
+//  WebView.swift
+//  DuckDuckGo
+//
+//  Copyright © 2021 DuckDuckGo. All rights reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//  http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import WebKit
+
+public protocol WebViewDelegate: AnyObject {
+
+    func webView(_ :WebView, didRequestDefinitionOfTerm term: String)
+
+}
+
+public class WebView: WKWebView {
+
+    public weak var webViewDelegate: WebViewDelegate?
+
+    // swiftlint:disable identifier_name
+    @IBAction func _lookup(_ sender: UIMenuController?) {
+        evaluateJavaScript("window.getSelection().toString()") { result, _ in
+            guard let term = result as? String else { return }
+            self.webViewDelegate?.webView(self, didRequestDefinitionOfTerm: term)
+        }
+    }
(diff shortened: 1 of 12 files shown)


DuckDuckGo/DefineTermViewController.swift
```diff
@@ -0,0 +1,112 @@
...
+        webView.navigationDelegate = self
+        view.addSubview(webView)
+
+        if let term = term {
+            load(term: term)
+        }
+    }
+
+    func load(term: String) {
+        print("***", #function, term)
```
samsymons: Debugging call leftover here 🙂

Core/WebView.swift
```diff
@@ -0,0 +1,41 @@
...
+
+}
+
+public class WebView: WKWebView {
+
+    public weak var webViewDelegate: WebViewDelegate?
+
+    // swiftlint:disable identifier_name
+    @IBAction func _lookup(_ sender: UIMenuController?) {
+        evaluateJavaScript("window.getSelection().toString()") { result, _ in
```
samsymons: A potential issue with this API is that it returns an empty string if you call it on any element within an iframe ([MDN](https://developer.mozilla.org/en-US/docs/Web/API/Window/getSelection)):

>When called on an <iframe> that is not displayed (eg. where display: none is set) Firefox will return null, whereas other browsers will return a Selection object with Selection.type set to None.

I took a quick go at checking whether `term` is empty and calling the superclass's implementation of this if it is, but it didn't show the typical lookup view.

samsymons (approved):
LGTM! Two comments here, not certain how much of an issue the `iframe` thing is.

Core/WebView.swift
brindy: If the frame is not displayed how would the user select text and then show the popover menu? 🤔 

I think this should be fine, but nice attention to detail.

samsymons: The case I tested with is [here](https://www.w3schools.com/html/html_iframe.asp), if you use Lookup on a string within that nested page it'll come back with an empty string. But again I really don't know how much of a problem this is. 😄

brindy: I see now - yeah I need to either do nothing if the text is empty or have another method for getting the selection :/

brindy: I've added another user script to detect selections.

@SlayterDev would you mind giving this a quick review as well, please?

brindy:
@samsymons I ended up doing some minor refactoring around user scripts so I could re-use the logic.  Can you give that a bit of test when you come to re-review this, please? (in particular check I'm not leaking memory)

Core/WebView.swift
samsymons: This looks good to me! :+1:

SlayterDev (approved):
JS looks good 👍

brindy:
We decided this isn't a great change and could have wider implications.  Thanks for the reviews though.

PR #895, opened 2021-06-06
Update Turkish localization strings

nesimtunc:


brindy:
Hi @nesimtunc - thanks for this but we don't accept translation PRs as we use a third party to do all our translations.  We need to update our contributions doc to reflect this, so sorry about that.  I will also add a note to feedback these to our translators next time the app goes through a translation cycle.

nesimtunc:
Hi,

Yeap, no worries. They may use it as a reference. 

Thank you

PR #897, opened 2021-06-06
Improve the quality of fetched favicons

samsymons:
Task/Issue URL: https://app.asana.com/0/414235014887631/1198210503453245/f
Tech Design URL:
CC:

**Description**:

This PR improves the quality of fetched favicons. It does so by augmenting the existing favicon fetching code to use the `LinkPresentation` first, which is a framework designed to fetch metadata for URLs.

**Steps to test this PR**:
1. Run the app on develop and bookmark as many domains as you can think of. See the Asana task for a shortlist of domains which currently display in lower resolutions in the app.
1. Clear the domain cache (there's a new option in the debug menu for this) and load favicons again, checking that they increase in quality.

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPad

**OS Testing**:

* [ ] iOS 13
* [ ] iOS 14

---
###### Internal references:

### DIFF
--- a/DuckDuckGo/Favicons.swift
+++ b/DuckDuckGo/Favicons.swift
@@ -22,7 +22,7 @@ import UIKit
 import os
 import LinkPresentation
 
-// swiftlint:disable type_body_length
+// swiftlint:disable type_body_length file_length
 public class Favicons {
 
     public struct Constants {
@@ -412,4 +412,4 @@ public class Favicons {
     }
 
 }
-// swiftlint:enable type_body_length
+// swiftlint:enable type_body_length file_length
--- a/DuckDuckGo/Favicons.swift
+++ b/DuckDuckGo/Favicons.swift
@@ -280,7 +280,6 @@ public class Favicons {
 
         if #available(iOS 13.0, *) {
             retrieveLinkPresentationImage(from: domain) {
-
                 guard let image = $0 else {
                     self.retrieveBestImage(bestSources: bestSources, additionalSources: additionalSources, completion: completion)
                     return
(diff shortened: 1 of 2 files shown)


DuckDuckGo/Favicons.swift
```diff
@@ -292,6 +304,26 @@ public class Favicons {
         }
     }
 
+    @available(iOS 13.0, *)
```
brindy: Unneeded?

DuckDuckGo/Favicons.swift
```diff
@@ -292,6 +304,26 @@ public class Favicons {
...
+    private func retrieveLinkPresentationImage(from domain: String, completion: @escaping (UIImage?) -> Void) {
+        guard let url = URL(string: "http://\(domain)") else {
```
brindy: Should we stick to https? http is handled in the fallback already, better to have a bad icon than potentially leak over http.

DuckDuckGo/ImageCacheDebugViewController.swift
```diff
@@ -18,6 +18,7 @@
 //
 
 import UIKit
+import Core
```
brindy, line 21: Unneeded version check on this class

brindy (approved):
Few minor things, but otherwise looks good. :)

DuckDuckGo/ImageCacheDebugViewController.swift
samsymons: Thanks, fixed - this shows how old this branch actually is. 🙂

DuckDuckGo/Favicons.swift
samsymons: Yeah good call. I made this change and tested it on a number of cases, didn't notice any impact on favicon quality, so I've pushed the HTTPS change up. :+1:

PR #922, opened 2021-07-08
Properly load unprotected list

SlayterDev:
Task/Issue URL: https://app.asana.com/0/414709148257752/1200561434937075/f
Tech Design URL:
CC:

**Description**:
This PR fixes privacy dashboard performance issues for sites with privacy protection disabled.

**Steps to test this PR**:
1. Visit: foxnews.com, td.com, https://www.nintendo.com/switch/oled-model/ (with protections off)
1. Browser and privacy dashboard should remain responsive

**Orientation Testing**:

* [x] Portrait
* [x] Landscape

**Device Testing**:

* [x] iPhone SE (1st Gen)
* [x] iPhone 8
* [x] iPhone X
* [x] iPad

**OS Testing**:

* [x] iOS 13
* [x] iOS 14

**Theme Testing**:

* [x] Light theme
* [x] Dark theme

---
###### Internal references:

### DIFF
--- a/Core/ContentBlockerConfigurationUserDefaults.swift
+++ b/Core/ContentBlockerConfigurationUserDefaults.swift
@@ -26,11 +26,31 @@ public class ContentBlockerProtectionUserDefaults: ContentBlockerProtectionStore
         static let unprotectedDomains = "com.duckduckgo.contentblocker.whitelist"
         static let trackerList = "com.duckduckgo.trackerList"
     }
+    
+    private let lock = NSLock()
 
     private let suiteName: String
+    
+    private var _storageCache: StorageCacheProvider?
+    private var storageCache: StorageCacheProvider? {
+        lock.lock()
+        if _storageCache == nil {
+            _storageCache = StorageCacheProvider()
+        }
+        lock.unlock()
+        return _storageCache
+    }
 
     public init(suiteName: String = ContentBlockerStoreConstants.groupName) {
         self.suiteName =  suiteName
+        NotificationCenter.default.addObserver(self,
+                                               selector: #selector(onStorageChange),
+                                               name: StorageCacheProvider.didUpdateStorageCacheNotification,
+                                               object: nil)
+    }
+    
+    deinit {
+        NotificationCenter.default.removeObserver(self, name: StorageCacheProvider.didUpdateStorageCacheNotification, object: nil)
     }
 
     private var userDefaults: UserDefaults? {
@@ -52,10 +72,7 @@ public class ContentBlockerProtectionUserDefaults: ContentBlockerProtectionStore
         }
     }
     
-    private var tempUnprotectedDomains: [String] {
-        return StorageCacheProvider().current.fileStore.loadAsArray(forConfiguration: .temporaryUnprotectedSites)
-            .filter { !$0.trimWhitespace().isEmpty }
-    }
+    private var tempUnprotectedDomains: [String]?
 
     public func isProtected(domain: String?) -> Bool {
         guard let domain = domain else { return true }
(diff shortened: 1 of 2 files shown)


Core/ContentBlockerConfigurationUserDefaults.swift
```diff
@@ -26,11 +26,31 @@ public class ContentBlockerProtectionUserDefaults: ContentBlockerProtectionStore
         static let unprotectedDomains = "com.duckduckgo.contentblocker.whitelist"
         static let trackerList = "com.duckduckgo.trackerList"
     }
+    
+    private let lock = NSLock()
 
     private let suiteName: String
+    
+    private var _storageCache: StorageCacheProvider?
+    private var storageCache: StorageCacheProvider? {
```
bwaresiak: Since this always returns an instance, it shouldn't be an optional.

bwaresiak (approved):
Looks good to me!

Core/ContentBlockerConfigurationUserDefaults.swift
```diff
@@ -66,10 +72,15 @@ public class ContentBlockerProtectionUserDefaults: ContentBlockerProtectionStore
     public func isTempUnprotected(domain: String?) -> Bool {
         guard let domain = domain else { return false }
         
+        if tempUnprotectedDomains == nil {
+            tempUnprotectedDomains = FileStore().loadAsArray(forConfiguration: .temporaryUnprotectedSites)
+                .filter { !$0.trimWhitespace().isEmpty }
```
brindy, line 77: @bwaresiak / @SlayterDev 

I know this is merged but why wasn’t a lazy var good enough here? 

The comment says that var should be constrained to main thread but nothing is constraining this assignment to the main thread.

Core/ContentBlockerConfigurationUserDefaults.swift
```diff
@@ -66,10 +72,15 @@ public class ContentBlockerProtectionUserDefaults: ContentBlockerProtectionStore
...
+        }
+        
         // Break domain apart to handle www.*
         var tempDomain = domain
         while tempDomain.contains(".") {
-            if tempUnprotectedDomains.contains(tempDomain) {
+            if tempUnprotectedDomains!.contains(tempDomain) {
```
brindy, line 83: 😬 

don’t like this force unwrap, would be better to keep it optional and either default to false using ?? false or check explicitly against the bool using == true

Core/ContentBlockerConfigurationUserDefaults.swift
```diff
@@ -29,8 +29,19 @@ public class ContentBlockerProtectionUserDefaults: ContentBlockerProtectionStore
 
     private let suiteName: String
 
+    // This variable should be confined to the Main Thread
+    private var tempUnprotectedDomains: [String]?
```
brindy, line 33: I think we should consider changing this to a computed property which uses the app dependency provider, and modify the storage cache to have a cached version of this list which it reads lazily from file storage.

We also inject the storage cache in to this class and load the temp list there so could do the same there too.

  https://github.com/duckduckgo/iOS/blob/develop/Core/ContentBlockerRulesUserScript.swift#L51

brindy: @samsymons pointed out this class is in Core.

I think we should consider flattening core into the app and remodularising to put the right access levels on these config classes.

Extensions that use core classes can just reference those files directly via their target. 

I’ll have a think about that.

Core/ContentBlockerConfigurationUserDefaults.swift
bwaresiak: This is used from UI only (at the moment) so I just put a comment there to remember when we refactor it (which we will have to do).

Core/ContentBlockerConfigurationUserDefaults.swift
bwaresiak: I agree with all of your comments - this was a quick fix to a performance problem. When we started diving into that we figured out there is more to that, which I pointed in in the original task and even created a follow up one: https://app.asana.com/0/414709148257752/1200571576333117 Basically we need to change how we inject these dependencies and how they relate to each other.

PR #911, opened 2021-06-27
Fix height calculation on onboarding messages

Bunn:
Task/Issue URL: Fix issue https://github.com/duckduckgo/iOS/issues/910
Asana: https://app.asana.com/0/414709148257752/1200623609427972

**Description**:
The storyboard containing the layout of the onboarding screen was considering the minimum width to be 375 instead of 320. The Fix was just a change on 2 views on the storyboard width's constant.

Code reviews for StoryBoards are generally not straightforward so in case the reviewer wants to just close the PR and do the changes on the reviewer's machine here are the only changes made:

Screen Shot 2021-06-27 at 19 45 54

**Steps to test this PR**:
- On an iPhone SE or iPod touch
- Open the app for the first time
- See first onboarding message
- Open a website
- Open a new tab
- See the subsequent onboarding message

Here's a video illustrating the steps to reproduce:

https://user-images.githubusercontent.com/782316/123559230-af3ade80-d792-11eb-8f32-c9fd24f29c52.mp4

**Orientation Testing**:

* [x] Portrait
* [x] Landscape

**Device Testing**:

* [x] iPhone SE (1st Gen)
* [x] iPhone 8
* [x] iPhone 11
* [x] iPod touch (7th Gen)

**Test screenshots and video**

https://user-images.githubusercontent.com/782316/123559316-283a3600-d793-11eb-913d-1946eaa91f79.mp4

Screen Shot 2021-06-27 at 19 48 37
Screen Shot 2021-06-27 at 19 48 50
Screen Shot 2021-06-27 at 19 49 11
Screen Shot 2021-06-27 at 19 49 28

PR #930, opened 2021-07-20
Revert "Fix height calculation on onboarding messages"

THISISDINOSAUR:
Task/Issue URL: https://app.asana.com/0/414709148257752/1200630479174169/f
Tech Design URL:
CC:

**Description**:
Reverts https://github.com/duckduckgo/iOS/pull/911
As a side effect of the above, it shrinks dax dialogs from <=375px to <= 320px, which we don't want to do. It also changes from Proxima Nova to the system font in some places.
I think fixing what the above aims to fix would require a more nuanced solution and for now I think we should just revert it.

**Steps to test this PR**:
1. Go through onboarding with a larger device (e.g. iPhone 12) and check that the dax dialogs are the correct width (i.e. >320px)

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPad

**OS Testing**:

* [ ] iOS 13
* [ ] iOS 14

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:

Reverts duckduckgo/iOS#911


brindy (approved):
LGTM

PR #934, opened 2021-07-21
Remote Configuration

SlayterDev:
Task/Issue URL: https://app.asana.com/0/0/1200434558302342/f
Tech Design URL: https://app.asana.com/0/481882893211075/1200380127788005/f
CC: @brindy @bwaresiak @catehstn 

**Description**:
This PR integrates the new [remote configuration](https://github.com/duckduckgo/privacy-configuration) for privacy features and deprecates the legacy temp unprotected list.

**Steps to test this PR**:
1. Visit well known sites with trackers
2. Tracker count should match before and after changes
3. Visit https://privacy-test-pages.glitch.me/privacy-protections/fingerprinting/
4. Make sure `Include all headers and all properties of window, navigator and screen objects` is checked and run the test
5. Values overridden by `fingerprint.js` should match the results
6. Visit https://global-privacy-control.glitch.me/
7. Header and DOM signal should be present
8. Visit a [temp unprotected site](https://github.com/duckduckgo/privacy-configuration/blob/main/exception-lists/trackers-unprotected-temporary.json)
9. Site should display as "temporarily unprotected" in the privacy dashboard and should show trackers detected.

**Orientation Testing**:

* [x] Portrait
* [x] Landscape

**Device Testing**:

* [x] iPhone SE (1st Gen)
* [x] iPhone 8
* [x] iPhone X
* [x] iPad

**OS Testing**:

* [x] iOS 13
* [x] iOS 14

---
###### Internal references:


Core/StorageCache.swift
```diff
@@ -64,17 +64,20 @@ public class StorageCache: StorageCacheUpdating {
...
+            if fileStore.persist(data as? Data, forConfiguration: configuration),
+               TrackerDataManager.shared.reload(etag: etag) != .downloaded {
+                Pixel.fire(pixel: .trackerDataReloadFailed)
+                return false
             }
-            return false
+            return true
             
-        case .temporaryUnprotectedSites:
-            return fileStore.persist(data as? Data, forConfiguration: configuration)
```
samsymons, line 77: Do you know if the old `temporaryUnprotectedSites` will get cleaned up? Ideally we should remove it if it isn't being used any more.

Core/StorageCache.swift
```diff
@@ -64,17 +70,23 @@ public class StorageCache: StorageCacheUpdating {
...
-                return true
+            if fileStore.persist(data as? Data, forConfiguration: configuration),
+               TrackerDataManager.shared.reload(etag: etag) != .downloaded {
+                Pixel.fire(pixel: .trackerDataReloadFailed)
+                return false
             }
-            return false
+            return true
+            
+        case .privacyConfiguration:
```
SlayterDev, line 80: @bwaresiak there is now a cyclomatic complexity warning in this method. I did some things to reduce this (see the if statements) but the lowest I can get it is 11. Not sure if you have any ideas here.

Core/PrivacyConfigurationManager.swift
```diff
@@ -0,0 +1,125 @@
...
+        return embeddedData.config
+    }
+
+    init(configData: PrivacyConfiguration) {
+        _fetchedData = (configData, "")
+    }
+
+    init() {
+        reload(etag: UserDefaultsETagStorage().etag(for: .privacyConfiguration))
+    }
```
bwaresiak, line 87: Is it intended that these inits work in a different way - one calls reload, but second doesn't?

Core/StorageCache.swift
```diff
@@ -57,24 +63,27 @@ public class StorageCache: StorageCacheUpdating {
...
-                    return false
-                }
-                return true
+            if fileStore.persist(data as? Data, forConfiguration: configuration),
+               TrackerDataManager.shared.reload(etag: etag) != .downloaded {
+                Pixel.fire(pixel: .trackerDataReloadFailed)
+                return false
             }
-            return false
+            return true
```
bwaresiak: Doesn't that change mean we will be returning true even if persist operation fails?

Core/StorageCache.swift
```diff
@@ -57,24 +63,27 @@ public class StorageCache: StorageCacheUpdating {
...
             
-        case .temporaryUnprotectedSites:
-            return fileStore.persist(data as? Data, forConfiguration: configuration)
+        case .privacyConfiguration:
+            if fileStore.persist(data as? Data, forConfiguration: configuration),
+               PrivacyConfigurationManager.shared.reload(etag: etag) != .downloaded {
+                Pixel.fire(pixel: .privacyConfigurationReloadFailed)
+                return false
+            }
+            return true
```
bwaresiak: Same here

Core/PrivacyConfigurationManager.swift
SlayterDev: I was using another class as a reference. The alternate init is never called so I removed it.

Core/ContentBlockerUserScript.swift
```diff
@@ -58,7 +61,8 @@ public class ContentBlockerUserScript: NSObject, UserScript {
         return Self.loadJS("contentblocker", from: Bundle.core, withReplacements: [
             "${unprotectedDomains}": unprotectedDomains,
             "${trackerData}": trackerData,
-            "${surrogates}": surrogates
+            "${surrogates}": surrogates,
+            "${blockingEnabled}": PrivacyConfigurationManager.shared.privacyConfig.isEnabled(featureKey: .contentBlocking).description
```
bwaresiak: Accessing `description` or calling `String(describing:)` is actually discouraged in Swift - better explicitly use `isTrue ? "true" : "false"`

Core/ContentBlockerRulesSource.swift
```diff
@@ -41,11 +41,15 @@ class DefaultContentBlockerRulesSource: ContentBlockerRulesSource {
     }
 
     var tempListEtag: String? {
-        return UserDefaultsETagStorage().etag(for: .temporaryUnprotectedSites)
+        return UserDefaultsETagStorage().etag(for: .privacyConfiguration)
```
bwaresiak: Since we now have interface for that through `PrivacyConfigurationManager` (fetched/embedded fields contain pair config:etag) we should use that abstraction instead of reaching to storage directly.

Core/ContentBlockerRulesSource.swift
```diff
@@ -41,11 +41,16 @@ class DefaultContentBlockerRulesSource: ContentBlockerRulesSource {
     }
 
     var tempListEtag: String? {
-        return UserDefaultsETagStorage().etag(for: .temporaryUnprotectedSites)
+        return PrivacyConfigurationManager.shared.fetchedData?.etag
+            ?? PrivacyConfigurationManager.shared.embeddedData.etag
```
bwaresiak, line 45: I have one follow up question: how often can configuration change and how many of these changes can be unrelated to content blocker (temp list, etc)? Right now any change to that config will force every device to rebuild content blockingrules, so if we are going to have frequent updates that are unrelated to that, I'd rather see some other UID to be provided for that subset of data.

SlayterDev: I think the most common updates will be to the `unprotectedTemporary` or `contentBlocking` exemptions. Both would require a rebuild of the CBRs. Feature toggles won't change much and so far we haven't seen many instances of needing exception rules for the other features. I don't imagine it changing much more often than the old list did.

bwaresiak: Ok, so that's current state and it is fine. Now - outside of this PR - can you think of steps to guarantee that any changes to that (e.g. introducing new processes around its release or adding new functionalities) will take into account possible impact of any updates?

Core/PrivacyConfiguration.swift
```diff
@@ -0,0 +1,99 @@
...
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+
+public struct PrivacyConfiguration: Codable {
```
bwaresiak, line 22: Since we are embedding configuration now, along with etag, we should add unit tests that:
- Validate if etag matches embedded file using SHA - see `testWhenEmbeddedDataIsUpdatedThenUpdateSHAAndEtag()` for reference.
- Validate if embedded config is what we expect e.g. if tracker blocking and gpc is enabled.

Core/ContentBlockerRulesSource.swift
SlayterDev: Remote Config will likely become an AoR. I'll make sure these considerations are documented and considered as things change.

Core/StorageCache.swift
```diff
@@ -57,24 +63,29 @@ public class StorageCache: StorageCacheUpdating {
...
-            if fileStore.persist(data as? Data, forConfiguration: configuration) {
-                if TrackerDataManager.shared.reload(etag: etag) != .downloaded {
-                    Pixel.fire(pixel: .trackerDataReloadFailed)
-                    return false
-                }
+            if fileStore.persist(data as? Data, forConfiguration: configuration),
+               TrackerDataManager.shared.reload(etag: etag) == .downloaded {
                 return true
+            } else {
+                Pixel.fire(pixel: .trackerDataReloadFailed)
```
bwaresiak: I'm not sure about this change: previously we reported when file has been malformed, but with this it is also all issues related to storage (persistence). I don't think we need this, as amount of free space on the device is outside of our control and not actionable - I'd rather see we stick to old behavior here and adapt it to new privacy config.

SlayterDev: The main reason behind this change is the cyclomatic complexity of the method was getting too high. I can revert it to the old logic but we may need to break each case into its own method? Or perhaps just the more complex ones?

Core/StorageCache.swift
```diff
@@ -38,11 +38,17 @@ public class StorageCache: StorageCacheUpdating {
...
+        // Remove legacy data
+        _ = fileStore.removeData(forConfiguration: .temporaryUnprotectedSites)
     }
     
     public init(tld: TLD, termsOfServiceStore: TermsOfServiceStore) {
         self.tld = tld
         self.termsOfServiceStore = termsOfServiceStore
+        
+        // Remove legacy data
+        _ = fileStore.removeData(forConfiguration: .temporaryUnprotectedSites)
```
bwaresiak: Can we get rid of `temporaryUnprotectedSites` enum (and related functions) and just make explicit function to remove that legacy file?

Core/StorageCache.swift
bwaresiak: Then we definitely should keep old behavior. I'd just silence the warning as I prefer to have this in one block of code.

Core/FileStore.swift
```diff
@@ -36,6 +36,21 @@ public class FileStore {
         }
     }
     
+    func removeData(forFile file: String) -> Bool {
+        var path = FileManager.default.containerURL(forSecurityApplicationGroupIdentifier: groupIdentifier)
+        path = path!.appendingPathComponent(file)
+        guard let path = path else { return false }
+        guard FileManager.default.fileExists(atPath: path.absoluteString) else { return true }
```
bwaresiak: Please call `path` a `fileURL` and use `fileURL.path` instead of `absoluteString` as the latter doesn't work with `removeItem(at:)` api.

bwaresiak (approved):
Great job!

PR #937, opened 2021-07-23
(Hopefully) Fix tracker reports sometimes being incorrect due to race condition

THISISDINOSAUR:
Task/Issue URL: https://app.asana.com/0/414709148257752/1200297306793824/f
Tech Design URL:
CC:

**Description**:
A potential fix, but I haven't been able to recreate the original issue to know for sure, so will have to wait for Bartek's return

**Steps to test this PR**:
1. Disable the dax logo on the SERP so you can see the site rating
1. Attempt to make it so a tracker is (incorrectly) reported on the SERP. If this works you shouldn't succeed.

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [x] iPhone 8
* [x] iPhone X
* [ ] iPad

**OS Testing**:

* [ ] iOS 13
* [x] iOS 14

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:


Core/DetectedTracker.swift
```diff
@@ -27,12 +27,14 @@ public struct DetectedTracker {
     public let knownTracker: KnownTracker?
     public let entity: Entity?
     public let blocked: Bool
+    public let pageUrl: String?
     
-    public init(url: String, knownTracker: KnownTracker?, entity: Entity?, blocked: Bool) {
+    public init(url: String, knownTracker: KnownTracker?, entity: Entity?, blocked: Bool, pageUrl: String? = nil) {
```
bwaresiak: Shouldn't pageURL be non-optional?

THISISDINOSAUR: ContentBlockerUserScript (i.e. the other user script) also uses DetectedTracker but doesn't receive the page URL from the JS script, maybe it should? I don't 100% know what is done as a result of any message from that script.

I can see SiteRating tracks installed surrogates for broken site reports, I guess they probably suffer from the same race condition so should have the same check.

bwaresiak: Yes, I think it should - otherwise `guard tracker.pageUrl == url.absoluteString else { return }` won't let the code through as pageURL is nil - it happens if you break on that return and visit page with surrogates.

Core/TrackerResolver.swift
```diff
@@ -74,7 +74,7 @@ class TrackerResolver {
             return nil
         }
 
-        return DetectedTracker(url: trackerUrlString, knownTracker: tracker, entity: entity, blocked: blocked)
+        return DetectedTracker(url: trackerUrlString, knownTracker: tracker, entity: entity, blocked: blocked, pageUrl: pageUrlString)
```
THISISDINOSAUR, line 77: I've just realised that pageUrlString doesn't even come from the WKMessage like I thought it did, but is from message.webview.url, so good chance it doesn't even fix the race condition anyway 🤦‍♀️

THISISDINOSAUR: Existing code is
`guard let pageUrlStr = message.webView?.url?.absoluteString ?? dict[ContentBlockerKey.pageUrl] as? String else { return }`
So I'm not sure in what conditions we don't get ` dict[ContentBlockerKey.pageUrl]`, but I think we should just do 
`guard let pageUrlStr = dict[ContentBlockerKey.pageUrl] as? String else { return }`

Core/ContentBlockerRulesUserScript.swift
```diff
@@ -67,7 +67,7 @@ public class ContentBlockerRulesUserScript: NSObject, UserScript {
         guard let blocked = dict[ContentBlockerKey.blocked] as? Bool else { return }
         guard let trackerUrlString = dict[ContentBlockerKey.url] as? String else { return }
         let resourceType = (dict[ContentBlockerKey.resourceType] as? String) ?? "unknown"
-        guard let pageUrlStr = message.webView?.url?.absoluteString ?? dict[ContentBlockerKey.pageUrl] as? String else { return }
+        guard let pageUrlStr = dict[ContentBlockerKey.pageUrl] as? String else { return }
```
THISISDINOSAUR, line 111: Do you know of any time we don't get `dict[ContentBlockerKey.pageUrl]`? Removing the `message.webView?.url?.absoluteString` part seems necessary to me to avoid the race condition, but I wonder why it was there in the first place, so assume it was probably to fix something else 🙈

Core/DetectedTracker.swift
THISISDINOSAUR: I've added the page url to the messages sent from the other script and made pageURL not optional now

DuckDuckGoTests/DaxDialogTests.swift
```diff
@@ -18,7 +18,6 @@
 //
 
 import XCTest
```
THISISDINOSAUR, line 20: Apologies for polluting the diff with seemingly unrelated changes, but I noticed we had two sets of dax dialog tests left over from the fire button experiment, so I either needed to fix the extra ones to cope with the changes to DetectedTracker, or clean them up, so went for the latter

Core/ContentBlockerRulesUserScript.swift
bwaresiak: @SlayterDev I think that (adding `message.webView?.url?.absoluteString`) is a change you've made - do you remember why?

SlayterDev: It was related to a bug fix but I'm not sure this ended up affecting anything. It can be removed if needed.

bwaresiak (approved):
Great work, sorry it took me a while to get to this. I've tested it and it works as expected, I can confirm misreports are solved by URL checks.

Issue #901, opened 2021-06-18
History API is ignored if a page is opened in a new tab (target "_blank")


SimonBackx:
## Environment
App version: 7.63.7.0
Device model: iPhone 12 pro
iOS version: 14.6

## Steps to reproduce the bug

- Start with a simple website (the initial website) that opens a different website (the opened website) in a new tab when clicking on a link:
```html
<a href="https://example.com" target="_blank">Open in new tab</a>
```
- On the opened website, use the History API's pushState method (e.g. when clicking on a button) to navigate to a new page
```js
history.pushState({ currentPage: "page1" }, "Page 1")
```
- Press the back button in the browser to go to the previous page

## Expected behavior

It should go back to the first page of the opened website.

## Actual behavior

It goes back to the initial website and skips over the first page (or more intermediate pages) on the opened website.

## Example

I've made a small repository with example code and hosted it so you can try it yourself without having to write any code.

- Open the following URL (= the initial website) in DuckDuckGo's browser for iOS:
`https://simonbackx.github.io/push-state-example/open-in-tab.html`
- Click on the link `Open in new tab`
- Hit one or more buttons `Home`, `Page1`, or `Page2`
- Go back using the browser buttons. You'll see that any intermediary pages are skipped

Now compare this with the normal behavior:
- Open the URL of the opened website by entering it directly in DuckDuckGo's browser for iOS, instead of opening it in a new tab using the initial website:
`https://simonbackx.github.io/push-state-example/`
- Hit one or more buttons `Home`, `Page1`, or `Page2`
- Go back using the browser buttons. You'll see that everything works as expected

You can check the source at https://github.com/SimonBackx/push-state-example

brindy:
Hi @SimonBackx - thanks for submitting this and sorry for not replying sooner.  We determine back/forward using the signals we get from WKWebView.   

I'm not sure there's much we can do about this, unless you have any ideas? It might better to create a bug for this in the WebKit issue tracker. https://webkit.org/reporting-bugs/

SimonBackx:
As far as I know, WKWebView doesn't have any support for target _blank out of the box, so the current behaviour is probably custom. I'll look into it and if you are open for it, open a PR with a fix.

brindy:
Yep, check these lines for starters:

* https://github.com/duckduckgo/iOS/blob/develop/DuckDuckGo/TabViewController.swift#L1268

And:
[...]

PR #952, opened 2021-08-23
Fix history API goBack not working (#901)

SimonBackx:
(sorry, had to remove the template since it was all internal, please feel free to ask if you need more information).

Task/Issue URL: https://github.com/duckduckgo/iOS/issues/901

This PR fixes a bug with opening new tabs, and then clicking the back button for websites that use the History API (pushState, replaceState). It closed the tab instead of going back. You can read the details in #901

The issue orginates in the following part of the code (TabViewController.swift):
```swift
    func goBack() {
        if isError {
            hideErrorMessage()
            url = webView.url
            onWebpageDidStartLoading(httpsForced: false)
            onWebpageDidFinishLoading()
        } else if webView.canGoBack && webView.goBack() != -1 {
            webView.backForwardList.backList
            chromeDelegate?.omniBar.resignFirstResponder()
        } else if openingTab != nil {
            delegate?.tabDidRequestClose(self)
        }
    }
```

During testing with the steps described in #901 (so pressing the back button when there was some navigation on a website using pushState instead of normal anchor text):
- the backlist (`webView.backForwardList.backList`) is filled
- `webView.canGoBack` returns `true`.
- `webView.goBack()` returns nil.

In every case, `webView.goBack()` is always executed. Which is good. The webView does go to the previous page in the new tab. The problem is, when the current tab has been opened by another one, it skips to the next `else if` statement and also closes the tab. 

I don't know why `webView.goBack()` returns nil when `webView.canGoBack` returns `true`. This might be a bug or inconsitency in the API or it could just mean than when the History API is used, there is no `WKNavigation` instance that could be returned (since there is no navigation). In fact, the only thing that has happend is some JavaScript that is executed on the page. This might be intended behaviour.

So I propose the following change. In case of `webView.canGoBack` returning true, `webView.goBack()` should always get executed and handled as if it succeeded. This should not have any impact on other parts of the code. For now, I still execute `chromeDelegate?.omniBar.resignFirstResponder()` afterwards, you might want to correct that if needed (I think it should work that way). For consistency, you might also want to change the `goForward` method.

```swift
    func goForward() {
        if webView.goForward() != nil {
            chromeDelegate?.omniBar.resignFirstResponder()
        }
    }
```

to

```swift
    func goForward() {
        if webView.canGoForward {
            webView.goForward()
            chromeDelegate?.omniBar.resignFirstResponder()
        }
    }
```

### DIFF
--- a/DuckDuckGo/TabViewController.swift
+++ b/DuckDuckGo/TabViewController.swift
@@ -479,7 +479,8 @@ class TabViewController: UIViewController {
             url = webView.url
             onWebpageDidStartLoading(httpsForced: false)
             onWebpageDidFinishLoading()
-        } else if webView.canGoBack && webView.goBack() != nil {
+        } else if webView.canGoBack {
+            webView.goBack()
             chromeDelegate?.omniBar.resignFirstResponder()
         } else if openingTab != nil {
             delegate?.tabDidRequestClose(self)


Anuwat11 (approved):

brindy:
Thanks @SimonBackx - someone from DDG will review this soon (hopefully this week).

brindy:
This is looking good @SimonBackx - I’m going to run it on my personal device over the weekend then will approve and merge on Monday if there’s no issues.   Thanks again for this.

Anuwat11 (approved):

brindy:
All seemed fine @SimonBackx, thanks again!

PR #968, opened 2021-09-16
Fix iOS 15 image favicon rendering

samsymons:
This is to work around issues with indexed color spaces failing to render correctly in widgets.

Task/Issue URL: https://app.asana.com/0/414235014887631/1201002203177706/f
Tech Design URL:
CC: @bwaresiak 

**Description**:

This PR fixes an issue where favicons in widgets on iOS 15 can sometimes fail to render correctly. 

**Steps to test this PR**:
1. Favorite some websites that exhibit this issue and verify that they're broken before swapping to this branch (see task for examples and steps on how to reproduce the issue)
1. Swap to this branch, and then clear the image cache
1. Visit each favorite website to re-fetch their favicons
1. Try to reproduce the issue again

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [x] iPhone X
* [x] iPad

**OS Testing**:

* [ ] iOS 13
* [x] iOS 14

---
###### Internal references:

### DIFF
--- a/DuckDuckGo/Favicons.swift
+++ b/DuckDuckGo/Favicons.swift
@@ -392,7 +392,6 @@ public class Favicons {
         return [
             .downloader(Constants.downloader),
             .requestModifier(Constants.requestModifier),
-            .cacheSerializer(FormatIndicatedCacheSerializer.jpeg),
             .targetCache(cache),
             expiry,
             .alternativeSources(sources)
--- a/DuckDuckGo/Favicons.swift
+++ b/DuckDuckGo/Favicons.swift
@@ -392,6 +392,7 @@ public class Favicons {
         return [
             .downloader(Constants.downloader),
             .requestModifier(Constants.requestModifier),
+            .cacheSerializer(FormatIndicatedCacheSerializer.jpeg),
             .targetCache(cache),
             expiry,
             .alternativeSources(sources)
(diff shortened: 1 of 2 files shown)


DuckDuckGo/Favicons.swift
```diff
@@ -392,6 +392,7 @@ public class Favicons {
         return [
             .downloader(Constants.downloader),
             .requestModifier(Constants.requestModifier),
+            .cacheSerializer(FormatIndicatedCacheSerializer.jpeg),
```
bwaresiak: What about entries that exist in the cache? Is migration possible/needed?

samsymons: Yes, we'll need to do something to handle existing cached images. `cacheSerializer` will convert images that haven't already been converted when reading them from the cache, but that doesn't apply to widgets as we don't go through Kingfisher to read them – currently looking into a way to solve this.

samsymons: Images are now converted to sRGB directly before rendering to avoid this issue. They're small and there will only be a fixed number of them, so it should be okay. Technically doing this negates the need to have this `cacheSerializer`, so it could be removed.

bwaresiak (approved):
Great job, LGTM!

PR #1014, opened 2021-11-30
Add Bookmark Folders

THISISDINOSAUR:
Task/Issue URL: https://app.asana.com/0/72649045549333/1143141356446970/f
Tech Design URL: https://app.asana.com/0/481882893211075/1200283968383818/f, https://app.asana.com/0/481882893211075/1201023911968241/f
CC:

**Description**:
Adds bookmark folders, including migrating bookmark and favourite storage from UserDefaults to CoreData.

**Steps to test this PR**:
Bookmarks/favorites appear in many parts of the app, so the test plan itself below should be scrutinised for anything missing. I've broken it down by area of the app to hopefully make that easier

**Test Migration**
1. To start, before you run this build, make sure you have favorites and bookmarks on a previous version of the app, and screenshot them, so you can compare their existence/position both pre and post migration
2. Install this PR
3. Open the app and go to the bookmark screen and check all favourites and bookmarks are in the right place (their ordering should be preserved)
 ---
1. Start with a previous build of the app again, with favs and bookmarks 
2. Install this PR 
3. Before you go to the bookmarks screen, add some favorites and bookmarks
4. Then go to the bookmarks screen, and check previous favs and bookmarks are in the right positions, and any new ones were correctly added to the end.
 ---
1. Delete any existing DDG app
2. Install this PR
3. Go to the bookmarks screen, ensure there is nothing there
4. Add a folder/bookmark/whatever, check it works correctly

**The browsing menu**
1. Go to a website that isn't a fav or bookmark
2. Check that the browsing menu shows the options to add it as a bookmark and a favorite
3. Add it as a bookmark from the browsing menu
4. Add it as a favourite from the browsing menu
5. Now when opening the browsing menu it should show the options to edit the bookmark, and remove the favorite
6. Edit the bookmark from the browsing menu, make sure the edit screen is the correct one for the bookmark and not the favourite.
7. Make some change (e.g. to the title) and save
8. Go to the bookmark screen to check the change saved correctly 
9. Go back to the browsing menu and remove the favourite
10. Open the browsing menu and check that the remove favourite option has changed to add favourite, and the bookmark option still says edit bookmark.
 ---
1. Create a bookmark from the browsing menu again
2. When the toast appears at the bottom of the screen to say the bookmark is created, press the edit button
3. Edit the bookmark and check it works as above
---
1. Create a favorite from the browsing menu 
2. When the toast appears, press undo. Check that the favorite doesn't exist
3. Do the reverse (remove a favourite from the browsing menu and press undo)

**The bookmarks screen**
1. Check that bookmarks and favourites show the correct favicons
2. Press on both a favorite and a bookmark, check they open the sites correctly
3. Swipe right on both a bookmark and a favorite and share them
4. Swipe left on a bookmark, a favorite and a folder and delete them. Check an empty folder doesn't show the confirmation prompt
5. Attempt to delete a folder with contents, it should show a confirmation prompt, check both options work as expected.
6. Press on a folder and check it shows the correct subfolder screen. 
---
1. Press the edit button, the cells should change to show the edit view
2. Rearrange some bookmarks and favorites 
3. Rearrange a bookmark into a favourite and vice versa
4. Attempting to do likewise with a folder, it shouldn't let you convert it to a folder.
5. Also test arranging bookmarks/folders from a sub folder.
---
1. When there are no bookmarks or favorites, check it shows the correct empty state
2. When a subfolder has no items, check it shows the correct empty state (different from the top level empty state)
---
1. Edit a bookmark, select a new parent folder
2. Save it, and check all views etc updated correctly 
---
1. Edit a bookmark, but press add a new folder
2. After creating a new folder, it should be selected automatically yon the edit bookmark screen
---
1. Edit a favorite, checks it doesn't show the select folder view
2. Make some changes and test saving it works 
---
1. Edit a favorite or bookmark, but don't save the changes. Check that no changes are saved
---
1. Edit a folder
2. Check that in the select folder view itself or any of its children don't appear (i.e. you shouldn't be able to add a folder to itself)

**Favorites in various places**
1. On iPad, check that the favourites overlay appears and its contents is still correct when using the url bar
---
1. Check that favorutes appear on the home screen
2. Long press and check that both editing and deleting favorites works from the home screen.
3. Whist on the Home Screen, open the bookmarks view and delete or edit a favorite, check this change is reflected in the Home Screen view

**Search**
1. From the bookmarks screen, scroll down to reveal the search bar
2. Try search terms that should show a) a bookmark b) a favorite, c) both, d) neither, and check the results are correct
3. Tap on a result to check that you can open a bookmark from the search results
4. Try to swipe to delete a search result, it shouldn't work
5. Try to swipe to share a search result, it _should_ work
Note: Folders should not appear in search results 
---
1. Type in URL bar search terms as above, check they are correct and the sites open correctly.

**Widgets**
1. When you have favorites already, add either the medium or large widget (i.e. one with favorites)
2. Check that the favorites appear there as expected (in the right order, with the right favicons etc)
3. Tap on a favourite and check it opens the site as expected
---
1. When you have no favorites and have not already been through favorites onboarding, add the widget
2. It should show the onboarding prompt, press on it
3. The app should show the favourites onboarding 
4. Once you've added a favorite, check the widget shows the new favorite
5. Then delete the fav, and check the change is propagated to the widget
---
1. Convert a bookmark to a favorites, or vice versa, and check the change is propagated to the widget

**Other**
1. Press the bookmark all tabs button from tabs screen, check that it does that.

**Pixels**
Test the following pixels fire appropriately
- case bookmarksFolderCreated = "m_bookmarks_folder_created"
- case bookmarkCreatedAtTopLevel = "m_bookmark_created_at_top_level"
- case bookmarkCreatedInSubfolder = "m_bookmark_created_in_subfolder"
- case bookmarkEditedAtTopLevel = "m_bookmark_edited_at_top_level"
- case bookmarkEditedInSubfolder = "m_bookmark_edited_in_subfolder"

**Copy Testing**:

**Orientation Testing**:

* [x] Portrait
* [x] Landscape

**Device Testing**:

* [x] iPhone SE (1st Gen)
* [ ] iPhone 8
* [x] iPhone X
* [x] iPad

**OS Testing**:

* [ ] iOS 13
* [x] iOS 14
* [x] iOS 15

**Theme Testing**:

* [x] Light theme
* [x] Dark theme

---
###### Internal references:


DuckDuckGo/BookmarksSearch.swift
```diff
@@ -19,34 +19,49 @@
...
+    func bookmarksAndFavorites(completion: @escaping ([Bookmark]) -> Void)
+}
+
+extension BookmarksCoreDataStorage: BookmarksSearchStore {
+    var hasData: Bool {
+        !topLevelBookmarksItems.isEmpty || !favorites.isEmpty
+    }
+    
+    func bookmarksAndFavorites(completion: @escaping ([Bookmark]) -> Void) {
+        allBookmarksAndFavoritesShallow(completion: completion)
```
bwaresiak: Why 'shallow'? Comment there suggests it is only used for deletion.

Core/BookmarksCoreDataStorage.swift
```diff
@@ -0,0 +1,722 @@
...
+    }
+    
+    private func updateParentAndSave(of item: BookmarkItemManagedObject,
+                                     parentID: NSManagedObjectID?,
+                                     context: NSManagedObjectContext,
+                                     completion: BookmarkItemSavedCompletion? = nil) {
+        
+        func updateParentAndSave(parent: BookmarkFolderManagedObject) {
+            item.parent = parent
+            parent.addToChildren(item)
```
bwaresiak: This is not needed (and I've seen this pattern in other places) - since relationships are bi-directional Core Data automatically manages graph consistency, so setting relation from one side is enough ( `item.parent = parent` )

Core/BookmarksCoreDataStorage.swift
```diff
@@ -0,0 +1,722 @@
...
+    // Since we'll never need to update children at the same time, we only support changing the title and parent
+    public func update(folderID: NSManagedObjectID,
+                       newTitle: String,
+                       newParentID: NSManagedObjectID,
+                       completion: BookmarkItemUpdatedCompletion? = nil) {
+        
+        let privateContext = getTemporaryPrivateContext()
+        privateContext.perform {
+
+            let mo = privateContext.object(with: folderID)
```
bwaresiak: We should not be using `object(with:)` but rather `existingObject(with:)` as the former one always returns object - even if it has ben removed from the store, which can result in the crash if we reach try to unfault such object.

DuckDuckGo/BookmarksSearch.swift
bwaresiak: Also note how this method is called for every character typed - be it in url field or in bookmarks search - this can lead to performance issues.

DuckDuckGo/BookmarksSearch.swift
```diff
@@ -92,21 +107,29 @@ class BookmarksSearch {
...
-        let results = bookmarksStore.favorites.map { ScoredLink(link: $0)} + bookmarksStore.bookmarks.map { ScoredLink(link: $0, score: -1) }
-        
-        let trimmed = query.trimWhitespace()
-        score(query: trimmed, results: results)
-        
-        var finalResult = results.filter { $0.score > 0 }
-        if sortByRelevance {
-            finalResult = finalResult.sorted { $0.score > $1.score }
+        bookmarksStore.bookmarksAndFavorites { bookmarks in
+            let results: [ScoredBookmark] = bookmarks.map {
```
bwaresiak: Hm, it feels a bit awkward that search ignores folders.

Core/BookmarksCoreDataStorage.swift
```diff
@@ -0,0 +1,722 @@
...
+    
+    public func containsFavorite(url: URL, completion: @escaping (Bool) -> Void) {
+        containsBookmark(url: url, searchType: .favoritesOnly, completion: completion)
+    }
+    
+    public func bookmark(forURL url: URL, completion: @escaping (Bookmark?) -> Void) {
+        viewContext.perform {
+    
+            let fetchRequest = NSFetchRequest<BookmarkManagedObject>(entityName: Constants.bookmarkClassName)
+            fetchRequest.predicate = NSPredicate(format: "url == %@ AND isFavorite == false", url as CVarArg)
```
bwaresiak: To make this less error-prone and avoid writing property names by hand this could be written like:

`fetchRequest.predicate = NSPredicate(format: "%K == %@ AND %K == false", #keyPath(BookmarkManagedObject.url), url as NSURL, #keyPath(BookmarkManagedObject.isFavorite))`

You also can cast to actual Foundation object instead of CVarArg.

Core/BookmarksCoreDataStorage.swift
```diff
@@ -0,0 +1,722 @@
...
+                fetchRequest.predicate = NSPredicate(format: "url == %@ AND isFavorite == true", url as CVarArg)
+            case .bookmarksAndFavorites:
+                fetchRequest.predicate = NSPredicate(format: "url == %@", url as CVarArg)
+            }
+            
+            guard let results = try? self.viewContext.fetch(fetchRequest) else {
+                completion(false)
+                return
+            }
+            completion(results.count > 0)
```
bwaresiak: If you are only interested in the count, then call `viewContext.count(for: fetchRequest)` instead of `viewContext.fetch(fetchRequest)`. Moreover, if it's only about existence (count > 0) then you can add limit to the fetch request: `request.fetchLimit = 1`

DuckDuckGo/TabSwitcherViewController.swift
```diff
@@ -224,19 +224,25 @@ class TabSwitcherViewController: UIViewController {
...
+                        newBookmarksCount += 1
+                    }
+                    group.leave()
                 }
             } else {
                 os_log("no valid link found for tab %s", log: generalLog, type: .debug, String(describing: tab))
             }
         }
         
+        group.wait()
```
bwaresiak: This will cause deadlock as bookmarksManager is tied to main thread and this is also invoked on main thread -  completion block of `bookmarksManager.contains` will be always waiting for execution on the thread that is currently blocked.

These kind of deadlocks (hard to foresee in a perfectly reasonably written code)  are the reason why it is advocated to return completion blocks on private queues unless there is a good reason not to, or that class is directly tied to main thread and it is made obvious to the developer (e.g. UIKit classes, animations, possibly ViewModels).

Ideally Bookmarks Manager should be thread-agnostic - but I understand how dispatching all completion blocks back to main thread could be a pain. What do you think?

Core/BookmarksCoreDataStorage.swift
```diff
@@ -0,0 +1,722 @@
...
+    
+    public func saveNewBookmark(withTitle title: String, url: URL, parentID: NSManagedObjectID?, completion: BookmarkItemSavedCompletion? = nil) {
+        createBookmark(url: url, title: title, isFavorite: false, parentID: parentID, completion: completion)
+    }
+    
+    // Since we'll never need to update children at the same time, we only support changing the title and parent
+    public func update(folderID: NSManagedObjectID,
+                       newTitle: String,
+                       newParentID: NSManagedObjectID,
+                       completion: BookmarkItemUpdatedCompletion? = nil) {
```
bwaresiak: One nitpick about completion handlers in this class: sometimes these will be invoked on main thread, sometimes on private queue of a private context - this isn't really consistent and may cause issues, as we cannot make any assumptions about the API.

Core/BookmarksCoreDataStorage.swift
```diff
@@ -0,0 +1,722 @@
...
+
+// MARK: respond to data updates
+extension BookmarksCoreDataStorage {
+            
+    @objc func contextDidSave(notification: Notification) {
+        guard let sender = notification.object as? NSManagedObjectContext else { return }
+        
+        if sender.name == Constants.privateContextName {
+            viewContext.perform { [weak self] in
+                self?.viewContext.mergeChanges(fromContextDidSave: notification)
```
bwaresiak, line 449: Very good. 😄

bwaresiak (changes requested):
Great job @THISISDINOSAUR it really works well. I've dropped some comment in the code, will be giving it another pass Tomorrow. 😄

DuckDuckGo/BookmarksSearch.swift
THISISDINOSAUR: Good point RE shallow, I can see how that's misleading. What do you think of replacing it with flat, since that's what I actually wanted to convey?
Re performance, my strategy was to try the simplest solution first and then test to see if the performance was okay, (which it seemed to be when I did my scale tests), to minimise the chance of inadvertently changing the search algorithm.
Do you have an opinion on what the best way to improving the performance might be? I can think of a few options, but I don't love any of them.

Core/BookmarksCoreDataStorage.swift
THISISDINOSAUR: I didn't think it was but was seeing some issue a long time ago related to these saves. I think that was an unrelated problem though that I resolved so I'll change these

DuckDuckGo/BookmarksSearch.swift
THISISDINOSAUR: Yeah, we discussed it and decided to exclude them, which is what Android did.

DuckDuckGo/TabSwitcherViewController.swift
THISISDINOSAUR: I'm really not sure on this one, in other cases I just ended up converting previously synchronous functions to async (e.g. building the browsing menu). This one will be a bit more of a PITA to do I think, but I don't have a better idea, do you?

Core/BookmarksCoreDataStorage.swift
THISISDINOSAUR: What would your preference be given objects being somewhat tied to threads?

DuckDuckGo/TabSwitcherViewController.swift
bwaresiak: I think that answer to https://github.com/duckduckgo/iOS/pull/1014#discussion_r770510085 may be helpful - if we were to go with the "completion on private queue only" then this issue should disappear, but you'd need to add `DispatchQueue.main` calls in other places. 

Alternatively, we could go with runloop juggling (ugh) or change this method to also have completion and invoke it on background queue.

Core/BookmarksCoreDataStorage.swift
bwaresiak: I don't think any objects that are retrieved from this API are tied to any particular thread - or am missing something?

DuckDuckGo/BookmarksSearch.swift
bwaresiak: I understand, thanks for confirmation!

DuckDuckGo/BookmarksSearch.swift
bwaresiak: Yes, naming it Flat should bring clarity.

Idea that comes to my mind: we could create transactional-like object to manage the state of search:

1. User starts "operation" (url typing, search in bookmarks) that can lead to multiple searchers.
2. We spawn 'BookmarksCachingSearch' object.
3. That object retrieves current state and caches it - now we use search algo on these objects only.
4. When we complete "operation" (leave url/text field) we discard the `BookmarksCachingSearch` object.

Assumption is that state is not changed between the searches so we don't need to re-fetch data. Object lifetime is tied to the cache so it is easy to use.

DuckDuckGo/TabSwitcherViewController.swift
THISISDINOSAUR: So you mean keep the dispatch group, but ensure all the bookmakarks manager calls complete on a private queue?

bwaresiak: Yes, this way there is no way for it to deadlock in such manner.

Core/BookmarksCoreDataStorage.swift
THISISDINOSAUR: I mean with e.g. `public func bookmark(forURL url: URL, completion: @escaping (Bookmark?) -> Void)`, it's completion will be invoked on the main thread, which sort of makes sense to me since the object is from the view context. Doesn't the fact that it's from the view context (and e.g. if I tried to check the parent or child and it's faulted on a background thread it could crash) mean we have to make assumptions about the API?

bwaresiak: Sorry, I got totally confused looking at the API. My mind got the impression that `Bookmark` is not a managed object - and thus it is not tied to any thread.

Now I have few trains of thought:
1. We could make API more explicit about what is returned: seeing `BookmarkManagedObject` in the completion would make people think about thread affiliation of particular instances. It would also disclose the information about inner implementation, which in this case is not bad as API has limitation around threading related to Core Data. I know it's not ideal as `Bookmark` abstraction is elegant, but at the same time it does not warn the user of this API about its limitations (I've assumed it's a thread-safe struct 🤦‍♂️and was ready to just pass it between threads... )
2. For methods that return object IDs in completion, I'd consider if we want to make it also return on the Main Thread, to keep API consistent. That would tie the whole object to Main Thread, but not in the performance-impairing way.
3. If users of this API understand relation to Main thread (based on above two points), then it is less likely any locking they implement won't take that into account - DispatchGroup from the above would have to be transformed to be async call based on that.

It would make this object fall in the "there is a good reason not use the private queue for completions" bucket. Does that makes sense to you?

THISISDINOSAUR: Yeah, that does make sense. Agreed on 1, I'd rather we had the abstraction, but, well, we just don't as long as we're using core data properly. At least at the UI layer etc we can still have it be more generic 
Do you think I should change BookmarksManager to return a `BookmarkManagedObject` instead too? 
2 also makes sense, but it still wouldn't feel super obvious from the outside to me. How would you feel about something in the completion type alias name to indicate the thread?

bwaresiak: Yes to all of your points, we should make it as clear as possible. :)

DuckDuckGo/TabSwitcherViewController.swift
THISISDINOSAUR: I've implemented and pushed a recursive function for this, but it means the tabs aren't necessarily saved in order. LMK if you can think of an easy way to resolve this

THISISDINOSAUR: Okay, ignore me, It was a very simple fix, it's late here...

DuckDuckGo/BookmarksCachingSearch.swift
```diff
@@ -92,21 +124,28 @@ class BookmarksSearch {
     }
     // swiftlint:enable cyclomatic_complexity
     
-    func search(query: String, sortByRelevance: Bool = true) -> [Link] {
+    func search(query: String, sortByRelevance: Bool = true, completion: @escaping ([Bookmark]) -> Void) {
         guard hasData else {
-            return []
+            completion([])
```
bwaresiak, line 129: I'm curious if this will happen - debug pixel maybe?

DuckDuckGo/BookmarksCachingSearch.swift
```diff
@@ -18,35 +18,67 @@
...
         }
     }
     
-    private let bookmarksStore: BookmarkStore
+    private let bookmarksStore: BookmarksSearchStore
     
-    init(bookmarksStore: BookmarkStore = BookmarkUserDefaults()) {
+    init(bookmarksStore: BookmarksSearchStore = BookmarksCoreDataStorage.shared) {
         self.bookmarksStore = bookmarksStore
+        loadCache()
```
bwaresiak, line 54: When I open Bookmarks and search for something, cache is not reset when leaving the text field - thus searching after deleting bookmark still returns it.

DuckDuckGo/AddOrEditBookmarkViewController.swift
```diff
@@ -0,0 +1,158 @@
...
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import UIKit
+import Core
+import CoreData
+
+class AddOrEditBookmarkViewController: UIViewController {
```
bwaresiak, line 24: When Editing bookmarks I get some layout constraints warnings (running on iPhone SE 1st gen).

Core/BookmarksCoreDataStorage.swift
```diff
@@ -0,0 +1,762 @@
...
+    case contextSaveError
+}
+
+public typealias BookmarkItemSavedMainThreadCompletion = ((NSManagedObjectID?, BookmarksCoreDataStorageError?) -> Void)
+public typealias BookmarkExistsMainThreadCompletion = ((Bool) -> Void)
+
+public typealias BookmarkItemDeletedBackgroundThreadCompletion = ((Bool, BookmarksCoreDataStorageError?) -> Void)
+public typealias BookmarkItemUpdatedBackgroundThreadCompletion = ((Bool, BookmarksCoreDataStorageError?) -> Void)
+public typealias BookmarkItemIndexUpdatedBackgroundThreadCompletion = ((Bool, BookmarksCoreDataStorageError?) -> Void)
+public typealias BookmarkConvertedBackgroundThreadCompletion = ((Bool, BookmarksCoreDataStorageError?) -> Void)
```
bwaresiak, line 51: This looks pretty clear :)

Core/BookmarksCoreDataStorage.swift
```diff
@@ -0,0 +1,762 @@
...
+            let results = try? self.viewContext.fetch(fetchRequest)
+            completion(results?.first)
+        }
+    }
+    
+    // Just used for favicon deletion and search
+    public func allBookmarksAndFavoritesFlat(completion: @escaping ([BookmarkManagedObject]) -> Void) {
+        viewContext.perform { [weak self] in
+        
+            let fetchRequest: NSFetchRequest<BookmarkManagedObject> = BookmarkManagedObject.fetchRequest()
```
bwaresiak, line 211: I'd consider adding parameter to configure whether we want to obtain bookmarks objects that are already prefetched (non-faults) to improve performance of search.

Core/BookmarksCoreDataStorage.swift
```diff
@@ -0,0 +1,762 @@
...
+            return nil
+        }
+        return folder
+    }
+    
+    public var topLevelBookmarksItems: [BookmarkItemManagedObject] {
+        guard let folder = cachedReadOnlyTopLevelBookmarksFolder else {
+            return []
+        }
+        return folder.children?.array as? [BookmarkItemManagedObject] ?? []
```
bwaresiak, line 154: Not for now, but generally: above is a strong reference (and it has to be to serve its purpose) but that may lead to Core Data keeping objects in memory - need to remember that if this turns out to be a problem someday.

DuckDuckGo/BookmarkSectionDataSources.swift
```diff
@@ -0,0 +1,251 @@
...
+        if let delegate = delegate,
+            let folder = item as? BookmarkFolder,
+            (folder.children?.count ?? 0) > 0 {
+            let title = String(format: UserText.deleteBookmarkFolderAlertTitle, folder.title ?? "")
+            let count = folder.children?.count ?? 0
+            let messageString: String
+            if count == 1 {
+                messageString = UserText.deleteBookmarkFolderAlertMessageSingular
+            } else {
+                messageString = UserText.deleteBookmarkFolderAlertMessagePlural
```
bwaresiak, line 192: Shouldn't this use strings dict to properly translate plural form? See `TabSwitcherViewController:refreshTitle()` for reference.

Also, if I have:
Bookmarks -> Folder A (one subfolder) -> Folder B (10 subfolders) message for removal of 'Folder A' will only warn me about removing one item, not a bunch of these from a subtree.

DuckDuckGo/BookmarksViewController.swift
```diff
@@ -21,26 +21,29 @@ import UIKit
 import Core
 import os.log
 
+// swiftlint:disable file_length
+
 class BookmarksViewController: UITableViewController {
 
+    @IBOutlet weak var addFolderButton: UIBarButtonItem!
     @IBOutlet weak var editButton: UIBarButtonItem!
```
bwaresiak, line 30: We probably should disable Add Folder & Edit buttons when in search mode (keyboard can be invisible when using external one).

bwaresiak (commented):
Great job, I've left few comments inline but overall it looks really solid and work really well. I went through the testing scenarios (including migration) and nothing looks broken, well done. :)

Feel free to merge when you're done.

DuckDuckGo/BookmarksCachingSearch.swift
THISISDINOSAUR: It should just happen if the user hasn't added anything, it's not checking the cache is populated or anything like that

THISISDINOSAUR: Although there'd be a decent argument for getting rid of this check as unnecessary, but I'm somewhat indifferent about it

Core/BookmarksCoreDataStorage.swift
THISISDINOSAUR: Sorry can you clarify what you mean?

bwaresiak: By default `fetchRequest` does not populate content of objects from the DB - so when you iterate over the result set (e.g. in search logic) it causes N (where N is number of objects) DB queries - one for each object. Instead, you can configure [`returnsObjectsAsFaults`](https://developer.apple.com/documentation/coredata/nsfetchrequest/1506756-returnsobjectsasfaults) to make things more efficient. It would be good to have that as an option here.

THISISDINOSAUR: Should it be an option? Sounds like I should just disable faulting for this method tbh

bwaresiak: I'd say yes - to not bind this API with certain usage. But I don't have strong opinion, you could just set that property to false here to improve performance of current use cases.

PR #1016, opened 2021-12-06
Brad/tracking params

SlayterDev:
Task/Issue URL: https://app.asana.com/0/72649045549333/1201286795398788/f
Tech Design URL: https://app.asana.com/0/481882893211075/1201299313936485/f
CC: @bwaresiak 

**Description**:
This PR adds support for removing URL parameters known to track users cross site.

**Steps to test this PR**:
**Note: Privacy Test Page is pending merge so will need to be tested locally**
1. Disable the experiment in `VariantManager`
1. Visit https://privacy-test-pages.glitch.me/privacy-protections/query-parameters/
2. Click the links on the page
3. The displayed "Results" on the landing page should match the "Expected" from the previous page

**Device Testing**:

* [x] iPhone SE (1st Gen)
* [x] iPhone 8
* [x] iPhone X
* [x] iPad

**OS Testing**:

* [x] iOS 13
* [x] iOS 14
* [x] iOS 15

---
###### Internal references:


brindy:
Hey @SlayterDev - Looks like there's test failures. I'm just re-running the build to check they're not something flakey, but if not can you have a look and I'll pick up this PR in the morning my time.  Thanks!

SlayterDev:
@Brindy looks like a regression when the privacy ref tests updated. Looking in to it.

Core/LinkCleaner.swift
```diff
@@ -88,4 +88,42 @@ public class LinkCleaner {
...
+        guard let url = url, !isURLExcluded(url: url, config: config, feature: .trackingParameters) else { return nil }
+        if let initiator = initiator, isURLExcluded(url: initiator, config: config, feature: .trackingParameters) {
+            return nil
+        }
+        
+        guard let urlsComps = URLComponents(url: url, resolvingAgainstBaseURL: false) else {
+            return nil
+        }
+        guard let queryParams = urlsComps.queryItems, queryParams.count > 0 else {
+            return nil
```
bwaresiak: Since this is "clean tracking params and return clean url" type of method, shouldn't this just return url if it has no params at all - meaning it is "clean"?

Core/LinkCleaner.swift
```diff
@@ -88,4 +88,42 @@ public class LinkCleaner {
         
         return nil
     }
+    
+    public func cleanTrackingParameters(initiator: URL?, url: URL?,
+                                        config: PrivacyConfiguration = PrivacyConfigurationManager.shared.privacyConfig) -> URL? {
```
bwaresiak, line 95: This seems like a good candidate for some unit tests, that test border conditions, config exceptions etc).

Core/LinkCleaner.swift
```diff
@@ -88,4 +88,42 @@ public class LinkCleaner {
...
+        guard let url = url, !isURLExcluded(url: url, config: config, feature: .trackingParameters) else { return nil }
+        if let initiator = initiator, isURLExcluded(url: initiator, config: config, feature: .trackingParameters) {
+            return nil
+        }
+        
+        guard let urlsComps = URLComponents(url: url, resolvingAgainstBaseURL: false) else {
```
bwaresiak: You could make this a `var`, then remove creation of new URLComponents instance and just assign `urlsComps.queryItems = preservedParams` at the end.

DuckDuckGo/TabViewController.swift
```diff
@@ -1146,6 +1151,13 @@ extension TabViewController: WKNavigationDelegate {
             return true
         }
         
+        if let newUrl = linkCleaner.cleanTrackingParameters(initiator: webView.url,
+                                                                   url: navigationAction.request.url) {
+            decisionHandler(.cancel)
+            load(newUrl: newUrl, forNavigationAction: navigationAction)
+            return true
+        }
```
bwaresiak, line 1171: Hmm can we have amp link that has tracking parameter? :) If so:
- I wonder how safe is having a multiple 'cancel' decisions.
- Do we have such test case in privacy pages?

bwaresiak: Also do we have an idea how many urls will be stripped of these links? I'm still concerned about possible impact of cancellation on redirects, reliability and history state of the tab...

Core/LinkCleaner.swift
SlayterDev: There is a test suite in the privacy reference tests that is included with this PR.

DuckDuckGo/TabViewController.swift
SlayterDev: Since we only clean URLs on deliberate navigation actions (typing URL, link click, open app by url, etc) we don't need to worry about redirects. Additionally history state should be unaffected as these actions happen before URL load.

> Do we have such test case in privacy pages?

I'll add a test case for this but my inclination is there should be no problem. It should follow:
1. AMP link with tracking params is clicked
2. AMP link is rewritten and a new load request is issued
3. AMP link is not detected but tracking parameters are detected and subsequently rewritten.

SlayterDev: > Do we have such test case in privacy pages?

you can try this URL for the time being and verify the parameter at the end is removed: https://www.google.com/amp/s/www.nytimes.com/2021/09/20/business/jeff-bezos-earth-fund.amp.html?utm_source=test

Core/VariantManager.swift
```diff
@@ -66,7 +67,10 @@ public struct Variant {
         Variant(name: "se", weight: doNotAllocate, isIncluded: When.always, features: []),
         
         Variant(name: "ma", weight: 1, isIncluded: When.isIOS15EnglishWithOnDeviceSpeech, features: [.voiceSearch]),
-        Variant(name: "mb", weight: 1, isIncluded: When.isIOS15EnglishWithOnDeviceSpeech, features: [])
+        Variant(name: "mb", weight: 1, isIncluded: When.isIOS15EnglishWithOnDeviceSpeech, features: []),
+        
+        Variant(name: "ba", weight: 1, isIncluded: When.always, features: [.parameterStripping]),
+        Variant(name: "bb", weight: 1, isIncluded: When.always, features: [])
```
bwaresiak: Is there and experiment template for this? Why not start variant name from "m" - this is typically done on mobile?

Core/LinkCleaner.swift
```diff
@@ -39,12 +40,12 @@ public class LinkCleaner {
...
     }
     
-    public func isURLExcluded(url: URL, config: PrivacyConfiguration) -> Bool {
+    public func isURLExcluded(url: URL, config: PrivacyConfiguration, feature: PrivacyFeature = .ampLinks) -> Bool {
         guard let host = url.host else { return true }
         
         if config.isTempUnprotected(domain: host)
             || config.isUserUnprotected(domain: host)
-            || config.isInExceptionList(domain: host, forFeature: .ampLinks) {
+            || config.isInExceptionList(domain: host, forFeature: feature) {
```
bwaresiak, line 49: We should also be checking whether feature itself is enabled.

DuckDuckGo/TabViewController.swift
```diff
@@ -1153,6 +1160,14 @@ extension TabViewController: WKNavigationDelegate {
             return true
         }
         
+        if AppDependencyProvider.shared.variantManager.isSupported(feature: .parameterStripping),
+           let newUrl = linkCleaner.cleanTrackingParameters(initiator: webView.url,
+                                                            url: navigationAction.request.url) {
+            decisionHandler(.cancel)
```
bwaresiak: Maybe we should double-check that new url is actually different from the old one before cancelling?

bwaresiak (commented):
@SlayterDev I've made a pass through this and left some comments. I'm out next week so in case you need final review, reach out to someone from iOS team. :)

Core/VariantManager.swift
SlayterDev: https://app.asana.com/0/500019618144197/1201665681727134/f

Changed the variant names. Was unaware of the convention.

Core/LinkCleaner.swift
```diff
@@ -88,4 +90,38 @@ public class LinkCleaner {
...
+        
+        let trackingParams = TrackingLinkSettings(fromConfig: config).trackingParameters
+        
+        let preservedParams: [URLQueryItem] = queryParams.filter { param in
+            for trackingParam in trackingParams {
+                if param.name.matches(pattern: "^\(trackingParam)$") {
+                    urlParametersRemoved = true
+                    return false
+                }
+            }
```
samsymons: Tiny Swift syntax note, not a blocker: we can use `for x in y where z` syntax to reduce some nesting:

```suggestion
            for trackingParam in trackingParams where param.name.matches(pattern: "^\(trackingParam)$") {
                urlParametersRemoved = true
                return false
            }
```

That said, since we're returning on the first match what we're really doing here is equivalent to `contains`:

```suggestion
            if trackingParams.contains(where: { param.name.matches(pattern: "^\($0)$") }) {
                urlParametersRemoved = true
                return false
            }
```

samsymons (approved):
Looks good. Bartek's comments are resolved, feature flag usage looks good, tested this against the privacy test pages and a bunch of typical web browsing as well – no issues found. 🙂

samsymons (approved):
Tested this again with the GPC fix, looks good :+1:

Issue #1023, opened 2021-12-23
[iPadOS] Back button is missing in Privacy Protection Practices


tsekityam:
## What did I do
1. Go to https://apple.com
2. Check SSL information of the page

## What did I see
Back button in the top left corner is missing, however, clicking that area will be sent back to the previous page.
![Simulator Screen Shot - iPad mini (6th generation) - 2021-12-23 at 23 42 12](https://user-images.githubusercontent.com/3814422/147263058-660096a9-aa79-4038-88b0-777956ae6b13.png)

## What do I expect
There is a back button

## iPad version
Simulator
* Version 13.2 (972.2)
* SimulatorKit 613.1
* CoreSimulator 783.5
![Simulator Screen Shot - iPad mini (6th generation) - 2021-12-23 at 23 47 43](https://user-images.githubusercontent.com/3814422/147263289-dec94005-940d-47b6-a20d-0c3953b44c4a.png)

## App version
6ece6fb5

UnfixedSun:
@tsekityam @roop @dav @ayoy @st3fan Please assign me this. Thanks.

samsymons:
This will be fixed by https://github.com/duckduckgo/iOS/pull/1373 once it's merged

PR #1024, opened 2021-12-23
Fix iPad detection logic

tsekityam:
Task/Issue URL: #1023 
Tech Design URL:
CC:

**Description**:
The back button is set to be hidden if `isPad()` return true. The button is hidden because the function return `false` even if the device is an iPad.

The current `isPad()` implementation use `traitCollection.horizontalSizeClass` to check if the device is an iPad or not, however, this property doesn't tell if the device is an iPad or not. So the function returns false negative value.

Change to use `UIDevice.current.userInterfaceIdiom` in `isPad()` to check if the device is an iPad, such that the function can return `true` for iPad correctly.

**Steps to test this PR**:
1. Go to https://apple.com
2. Open SSL certification information
3. Back button is shown on the top left corner

Simulator Screen Shot - iPad mini (6th generation) - 2021-12-23 at 23 40 52

**Copy Testing**:

**Orientation Testing**:

* [x] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [x] iPad

**OS Testing**:

* [ ] iOS 13
* [ ] iOS 14
* [ ] iOS 15
* [x] iPadOS 15

**Theme Testing**:

* [x] Light theme
* [ ] Dark theme

---
###### Internal references:

### DIFF
--- a/Core/UIViewControllerExtension.swift
+++ b/Core/UIViewControllerExtension.swift
@@ -27,7 +27,7 @@ extension UIViewController {
     }
     
     var isPad: Bool {
-        return traitCollection.horizontalSizeClass == .regular
+        return UIDevice.current.userInterfaceIdiom == .pad
     }
 
     @objc func buildActivities() -> [UIActivity] {


Bunn:
Hello @tsekityam 
thank you very much for taking a look at this, unfortunetly this change introduces new bugs when the iPad app is running on multitask mode since `UIDevice.current.userInterfaceIdiom` doesn't take into consideration that the app might be running on a compact size class.
bug

I'll close this for now, but feel free to reopen if you want to take a second look.
Thanks again!

PR #1035, opened 2022-01-14
Fix dax dialog presentation 

Bunn:
Add a check to make sure the dax dialog is only displayed if the current URL is the same as the one who originated it

Task/Issue URL: https://app.asana.com/0/414709148257752/1201620790053163/f

**Description**:

**Steps to test this PR**:

1. Clean install
2. Perform a search
3. Click a link that's likely got some trackers
4. Time it so that the page has loaded a bit but before the dialog shows
5. Swipe back
6. Dax dialog should not be displayed 

ps. After testing with a clean install, if you want to force the dialogs to be displayed without reinstalling the app I recommend removing the following lines:
https://github.com/duckduckgo/iOS/blob/123a583a08a313e0ef680824e036a1faaef1fc58/DuckDuckGo/DaxDialogs.swift#L205

https://github.com/duckduckgo/iOS/blob/123a583a08a313e0ef680824e036a1faaef1fc58/DuckDuckGo/DaxDialogs.swift#L288

https://github.com/duckduckgo/iOS/blob/123a583a08a313e0ef680824e036a1faaef1fc58/DuckDuckGo/FullscreenDaxDialogViewController.swift#L155

**Device Testing**:

* [x] iPhone 13
* [x] iPad

**OS Testing**:

* [ ] iOS 13
* [x] iOS 14
* [x] iOS 15

---
###### Internal references:

### DIFF
--- a/DuckDuckGo/DaxDialogs.swift
+++ b/DuckDuckGo/DaxDialogs.swift
@@ -44,7 +44,7 @@ class DaxDialogs {
         let accessibilityLabel: String?
     }
     
-    func overrideShownFlagForSpec(spec: BrowsingSpec, flag: Bool) {
+    func overrideShownFlagFor(_ spec: BrowsingSpec, flag: Bool) {
         switch spec.type {
         case .withMultipleTrackers, .withOneTracker :
             settings.browsingWithTrackersShown = flag
--- a/DuckDuckGo/DaxDialogs.swift
+++ b/DuckDuckGo/DaxDialogs.swift
@@ -46,7 +46,7 @@ class DaxDialogs {
     
     func overrideShownFlagForSpec(spec: BrowsingSpec, flag: Bool) {
         switch spec.type {
-        case .withMutipleTrackers, .withOneTracker :
+        case .withMultipleTrackers, .withOneTracker :
             settings.browsingWithTrackersShown = flag
         case .afterSearch:
             settings.browsingAfterSearchShown = flag
@@ -67,7 +67,7 @@ class DaxDialogs {
             case siteIsMajorTracker
             case siteOwnedByMajorTracker
             case withOneTracker
-            case withMutipleTrackers
+            case withMultipleTrackers
         }
         // swiftlint:enable nesting
 
@@ -96,10 +96,10 @@ class DaxDialogs {
                                                  highlightAddressBar: true,
                                                  pixelName: .daxDialogsWithTrackers, type: .withOneTracker)
         
-        static let withMutipleTrackers = BrowsingSpec(message: UserText.daxDialogBrowsingWithMultipleTrackers,
+        static let withMultipleTrackers = BrowsingSpec(message: UserText.daxDialogBrowsingWithMultipleTrackers,
                                                       cta: UserText.daxDialogBrowsingWithMultipleTrackersCTA,
                                                       highlightAddressBar: true,
-                                                      pixelName: .daxDialogsWithTrackers, type: .withMutipleTrackers)
+                                                      pixelName: .daxDialogsWithTrackers, type: .withMultipleTrackers)
         
         let message: String
         let cta: String
@@ -323,7 +323,7 @@ class DaxDialogs {
             
         default:
             settings.browsingWithTrackersShown = true
-            return BrowsingSpec.withMutipleTrackers.format(args: entitiesBlocked.count - 2,
+            return BrowsingSpec.withMultipleTrackers.format(args: entitiesBlocked.count - 2,
                                                            entitiesBlocked[0].displayName ?? "",
                                                            entitiesBlocked[1].displayName ?? "")
         }
--- a/DuckDuckGo/DaxDialogs.swift
+++ b/DuckDuckGo/DaxDialogs.swift
@@ -21,6 +21,8 @@ import Foundation
 import Core
 import TrackerRadarKit
 
+// swiftlint:disable type_body_length
+
 class DaxDialogs {
     
     struct MajorTrackers {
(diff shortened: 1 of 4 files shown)


DuckDuckGo/TabViewController.swift
```diff
@@ -1000,7 +1000,14 @@ extension TabViewController: WKNavigationDelegate {
         
         isShowingFullScreenDaxDialog = true
         scheduleTrackerNetworksAnimation(collapsing: !spec.highlightAddressBar)
+        let daxDialogSourceURL = self.url
+        
         DispatchQueue.main.asyncAfter(deadline: .now() + 0.5) { [weak self] in
+            // https://app.asana.com/0/414709148257752/1201620790053163/f
+            if self?.url != daxDialogSourceURL {
+                return
+            }
```
bwaresiak, line 1011: Question: won't that simply omit the Dax dialog, so given message (part of the flow) just disappears and is never shown?

Bunn: You're right,  the main problem is that DaxDialogs is deciding if itself was shown or not, for example:
https://github.com/duckduckgo/iOS/blob/123a583a08a313e0ef680824e036a1faaef1fc58/DuckDuckGo/DaxDialogs.swift#L296

This concept is wrong because DaxDialogs doesn't know if it was shown or not, it only creates the spec. Either `TabViewController` or `FullscreenDaxDialogViewController` knows that info, but not the DaxDialog since they're only holding data, not displaying it.


I can think about 2 different solutions that doesn't require a bigger refactor:
### 1 - Only update the flag once the view controller has shown the items.

On DaxDialogs remove all the shown flags = true on the methods where they are created, then add the following method:

```
    func updateShownFlag(_ spec: BrowsingSpec) {
        switch spec.type {
        case .withMutipleTrackers, .withOneTracker :
            settings.browsingWithTrackersShown = true
        case .afterSearch:
            settings.browsingAfterSearchShown = true
        case .withoutTrackers:
            settings.browsingWithoutTrackersShown = true
        case .siteIsMajorTracker, .siteOwnedByMajorTracker:
            settings.browsingMajorTrackingSiteShown = true
            settings.browsingWithoutTrackersShown = true
         }
    }

```
    
which would require a new type to be added since currently the same struct "type" can have different messages so we can't rely on the Equatable for the actual "type"


```
        enum BrowsingSpecType  {
        case afterSearch
        case withoutTrackers
        case siteIsMajorTracker
        case siteOwnedByMajorTracker
        case withOneTracker
        case withMutipleTrackers
    }

```
    
 and then on `FullScreenDaxDialogViewController`  can call:     `DaxDialogs.shared.updateShownFlag(spec)`

This approach makes sure the flag is only updated at the correct time, BUT, it breaks all the tests (which are quite a lot), and it's more disruptive since it's a paradigm shift.


### 2 - Add a method that overrides the flag

Similar to approach 1, we'd have a method on DaxDialogs:
```
    func overrideShownFlagForSpec(spec: BrowsingSpec, flag: Bool) {
        switch spec.type {
        case .withMutipleTrackers, .withOneTracker :
            settings.browsingWithTrackersShown = flag
        case .afterSearch:
            settings.browsingAfterSearchShown = flag
        case .withoutTrackers:
            settings.browsingWithoutTrackersShown = flag
        case .siteIsMajorTracker, .siteOwnedByMajorTracker:
            settings.browsingMajorTrackingSiteShown = flag
            settings.browsingWithoutTrackersShown = flag
         }
    }

```

But we'd leave the assigments to the flags on the methods that creates them like we have today.
And on TabViewController we add:
```
            if self?.url != daxDialogSourceURL {
                DaxDialogs.shared.overrideShownFlagForSpec(spec: spec, flag: false)
                return
            }


```
Which would cause the flag to be removed.

I'll go with solution number 2 because It doesn't change the paradigm of how the flags are set and doesn't break the tests. 

Please let me know if you disagree and/or have a suggestion for this.
Thank you!

DuckDuckGo/DaxDialogs.swift
```diff
@@ -33,61 +35,85 @@ class DaxDialogs {
...
         let message: String
         let accessibilityLabel: String?
-
+    }
+    
+    func overrideShownFlagForSpec(spec: BrowsingSpec, flag: Bool) {
+        switch spec.type {
+        case .withMultipleTrackers, .withOneTracker :
+            settings.browsingWithTrackersShown = flag
+        case .afterSearch:
```
bwaresiak, line 51: Instead of using enum (that is basically just denormalizing type - instance - information), we could just check against BrowsingSpec as it's equatable: `spec == BrowsingSpec.afterSearch` but at the same time I get how using enums enforce developer to look at this code path when expanding dax dialogs, so no strong feelings here. :)

DuckDuckGo/DaxDialogs.swift
```diff
@@ -33,61 +35,85 @@ class DaxDialogs {
...
         static let subsequent = HomeScreenSpec(message: UserText.daxDialogHomeSubsequent, accessibilityLabel: nil)
         static let addFavorite = HomeScreenSpec(message: UserText.daxDialogHomeAddFavorite,
                                                 accessibilityLabel: UserText.daxDialogHomeAddFavoriteAccessible)
 
         let message: String
         let accessibilityLabel: String?
-
+    }
+    
+    func overrideShownFlagForSpec(spec: BrowsingSpec, flag: Bool) {
```
bwaresiak: Naming nitpick: `For(spec:` or `ForSpec(_ spec:` or even `For(_ spec:` since we have type info in signature.

bwaresiak (approved):
Great job!

DuckDuckGo/DaxDialogs.swift
Bunn: Thanks. Going with:
`func overrideShownFlagFor(_ spec: BrowsingSpec, flag: Bool) {`

PR #1037, opened 2022-01-19
Add keychain items to debug menu

THISISDINOSAUR:
Task/Issue URL: -
Tech Design URL: -
CC:

**Description**:
Simulator Screen Shot - iPhone 8 - 2022-01-19 at 17 50 20

**Steps to test this PR**:
1. Go to debug menu, see if keychain items are there

### DIFF
--- a/DuckDuckGo/KeychainItemsDebugViewController.swift
+++ b/DuckDuckGo/KeychainItemsDebugViewController.swift
@@ -1,0 +1,132 @@
+//
+//  KeychainItemsDebugViewController.swift
+//  DuckDuckGo
+//
+//  Copyright © 2021 DuckDuckGo. All rights reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//  http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


DuckDuckGo/KeychainItemsDebugViewController.swift
```diff
@@ -0,0 +1,132 @@
...
+    let valueData: Data?
+    
+    var value: String? {
+        guard let valueData = valueData else { return nil }
+        return String(data: valueData, encoding: .utf8)
+    }
+    
+    // swiftlint:disable line_length
+    var displayString: String {
+        return "Service: \(service ?? "nil")\nAccount: \(account ?? "nil")\nValue as String: \(value ?? "nil")\nValue data: \(String(describing: valueData))"
```
Bunn: One small suggestion, you could do something like:
```
    var displayString: String {
        return """
            Service: \(service ?? "nil")
            Account: \(account ?? "nil")
            Value as String: \(value ?? "nil")
            Value data: \(String(describing: valueData))
            """
    }
    
```

and remove the lint disable rule

Bunn (approved):
Added one suggestion to the code style but LGTM.

DuckDuckGo/KeychainItemsDebugViewController.swift
THISISDINOSAUR: Definitely better ;)

PR #1038, opened 2022-01-20
Widget crash fix

samsymons:
Task/Issue URL: https://app.asana.com/0/414235014887631/1201654842382336/f
Tech Design URL:
CC: @THISISDINOSAUR 

**Description**:

This PR fixes a crash in the widget extension, where users who update to a version of the app that supports bookmarks folders but don't yet launch it will have their widget crash repeatedly without realizing.

This workaround is just to stop the crash, it doesn't restore favourites in the widget - users need to launch the app so that they can be migrated before favourites will appear again.

**Steps to test this PR**:
1. Install a pre-folders build of the app to the simulator, i.e. `git checkout 7.65.2`
2. Run the app, add some favorites
3. Add a widget to the Home Screen, verify that favourites appear
4. Kill the app and swap over to this branch with `git checkout sam/widget-crash-fix`
5. Run the new `WidgetsExtension` scheme that was added in this branch. This will launch the widget with the debugger attached but will **not** run the app, so the migration won't take place (which is what we want)
6. Verify that there are no favourites on the widget, and also that nothing has crashed. Add a breakpoint in `hasTopLevelFolder()` to make sure the debugger is actually attached, I had some issues with reliability here – keep trying if you can't hit a breakpoint in this function. Make sure the breakpoint is hit, and that no exceptions pop up after you resume.
7. Launch the app, verify that your favourites show up in the Bookmarks screen
8. Go back to the Home Screen, verify that favourites now show up in the widget

Also, test this on the on a physical device over top of the latest release if you can, just to make sure nothing's broken.

**Copy Testing**:

**OS Testing**:

* [ ] iOS 13
* [ ] iOS 14
* [ ] iOS 15

---
###### Internal references:

### DIFF
--- a/Core/BookmarksCoreDataStorage.swift
+++ b/Core/BookmarksCoreDataStorage.swift
@@ -424,6 +424,9 @@ extension BookmarksCoreDataStorage {
     
     public func favoritesUncachedForWidget(completion: @escaping ([BookmarkManagedObject]) -> Void) {
         guard BookmarksCoreDataStorageMigration.migratedFromUserDefaults else {
+            // Users with a pre-folders install who have then installed a build with folders and have not yet migrated it will
+            // not be able to read folders from Core Data. This will be resolved the first time they launch the app, so an empty
+            // set of favorites is returned here to avoid the widget crashing.
             completion([])
             return
         }
--- a/Core/BookmarksCoreDataStorage.swift
+++ b/Core/BookmarksCoreDataStorage.swift
@@ -717,7 +717,7 @@ extension BookmarksCoreDataStorage {
 public class BookmarksCoreDataStorageMigration {
     
     @UserDefaultsWrapper(key: .bookmarksMigratedFromUserDefaultsToCD, defaultValue: false)
-    fileprivate static var migratedFromUserDefaults: Bool
+    private static var migratedFromUserDefaults: Bool
     
     public static func migrate(fromBookmarkStore bookmarkStore: BookmarkStore, context: NSManagedObjectContext) {
         if migratedFromUserDefaults {
--- a/Core/BookmarksCoreDataStorage.swift
+++ b/Core/BookmarksCoreDataStorage.swift
@@ -719,7 +719,7 @@ public class BookmarksCoreDataStorageMigration {
     @UserDefaultsWrapper(key: .bookmarksMigratedFromUserDefaultsToCD, defaultValue: false)
     private static var migratedFromUserDefaults: Bool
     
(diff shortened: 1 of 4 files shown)


Core/BookmarksCoreDataStorage.swift
```diff
@@ -422,18 +422,34 @@ extension BookmarksCoreDataStorage {
...
+                    } else {
+                        fatalError("Favourites shouldn't contain folders")
+                    }
                 }
+                completion(favorites)
+            }
+        }
+    }
+    
+    private func hasTopLevelFolder() async -> Bool {
```
samsymons, line 446: This isn't the cleanest way to detect whether users have migrated, but I think it's the only choice:

1. We can't check `migratedFromUserDefaults` as it's not stored in a user defaults suite that is accessible to the widget
2. We can't check whether favourites data has been removed from the bookmarks user defaults suite as we're setting that to an empty array post-migration instead of removing it entirely; no way to tell if users are on an older version and simply don't have any favourites, or if they have no favourites because of the migration

DuckDuckGo/BookmarksCachingSearch.swift
```diff
@@ -1,5 +1,5 @@
 //
-//  BookmarksSearch.swift
+//  BookmarksCachingSearch.swift
```
samsymons, line 2: Tiny SwiftLint warning fix 🙂

Core/BookmarksCoreDataStorage.swift
THISISDINOSAUR: I think this is the best we can do. Love the use of async!

THISISDINOSAUR:
Screenshot 2022-01-20 at 12 45 07
bitrise isn't happy, we probably need to update the version of Xcode it's using to be able to use async (I've commented on the CI AoR about it)

THISISDINOSAUR:
> Screenshot 2022-01-20 at 12 45 07
> 
> bitrise isn't happy, we probably need to update the version of Xcode it's using to be able to use async (I've commented on the CI AoR about it)

Brindy has resolved!

samsymons:
@THISISDINOSAUR @brindy Nice, thanks!

THISISDINOSAUR:
I still have to test on my actual phone, but I'm finding step 8 doesn't work unless I make a change to a favourite. Nothing seems to refresh the widget until then (ofc the os probably will at some point? idk when).
We could solve this by refreshing the widget right after the migration?

THISISDINOSAUR (approved):
Looks great! You might want to quickly review what I added

THISISDINOSAUR (approved):

PR #1039, opened 2022-01-24
Add JS Alert block button

Bunn:
Task/Issue URL: https://app.asana.com/0/414235014887631/1201481513493461/f

**Description**:
Add block action to JS alerts

**Steps to test this PR**:
1. Go to https://privacy-test-pages.glitch.me/features/js-alerts.html
2. Tap on any alert
3. Select Block Alert option
4. Tap on other alerts. 
5. They should not be displayed again.
6. Open another tab
7. Open https://privacy-test-pages.glitch.me/features/js-alerts.html
8. Tap on any alert
9. They should be displayed

**Device Testing**:

* [x] iPhone 13
* [x] iPad

**OS Testing**:

* [x] iOS 14
* [x] iOS 15
* [x] iPadOS 15

---
###### Internal references:

### DIFF
--- a/Core/Pixel.swift
+++ b/Core/Pixel.swift
@@ -170,6 +170,9 @@ public enum PixelName: String {
     
     case textSizeSettingsShown = "m_text_size_settings_shown"
     case textSizeSettingsChanged = "m_text_size_settings_changed"
+    
+    case jsAlertShown = "m_js_alert_shown"
+    case jsAlertBlocked = "m_js_alert_blocked"
 
     // MARK: SERP pixels
     
--- a/DuckDuckGo/TabViewController.swift
+++ b/DuckDuckGo/TabViewController.swift
@@ -175,6 +175,7 @@ class TabViewController: UIViewController {
     private var textSizeUserScript = TextSizeUserScript()
     private var autofillUserScript = AutofillUserScript()
     private var debugScript = DebugUserScript()
+    private var shouldBlockJSAlert = false
 
     lazy var emailManager: EmailManager = {
         let emailManager = EmailManager()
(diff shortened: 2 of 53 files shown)


THISISDINOSAUR:
If I have the logic right of this: once a user blocks alerts it's not based on the domain or anything, that tab will never show alerts again? I could see this being a problem.

This is in the simulator, the blocking seems to work, but when I press on the recursive alerts after blocking it seems to lock up everything for a bit (the time seems quite variable, sometimes barely perceptible, sometimes quite a few seconds). Do you know what's happening there, is it expected, and could this still cause lockup problems for people?

I tried it on my actual device and couldn't replicate, so might just be crappy unrelated simulator performance? My device is pretty new though

Bunn:
@THISISDINOSAUR you're correct. The app is supposed to block all alerts for that tab, it's not related to the domain. This is the intended behavior.

As for the simulator issue, I've also noticed it, it works fine on all my test devices, including iPhone SE 1st gen and iPhone 6S

What happens is that if the delegate
`- (void)webView:(WKWebView *)webView runJavaScriptAlertPanelWithMessage:(NSString *)message initiatedByFrame:(WKFrameInfo *)frame completionHandler:(void (^)(void))completionHandler`

is called recursively WebKit kills it caller after a X number of calls where X seems to be very small on the actual device vs the simulator. (8x less, on a very unscientific test) 

I addressed the performance issue on this [comment](https://app.asana.com/0/0/1201481513493461/1201675903391678/f), maybe one alternative would be to inject some JS to prevent the site to call the alert again and stop the delegate from being spammed. 

Do you think I should pursue this alternative?

THISISDINOSAUR:
Okay, I think the performance issue is fine then, but I'm not super comfortable with blocking all alerts on the tab, has there been any discussion around it?

Bunn:
@THISISDINOSAUR It was on the proposal for the [PFR](https://app.asana.com/0/1175293949586521/1201692089019804/f)

Second bullet point: `After tapping on Disable Alerts, that tab would never display any JS Alert again`

THISISDINOSAUR (approved):
Ok thanks! I could see this causing some user confusion, but given the PFR decided that and it's a stop gap solution I can live with it.

PR #1040, opened 2022-01-26
Widget Education

jaceklyp:
Task/Issue URL: https://app.asana.com/0/72649045549333/1200189110180410/f

**Description**:
Introduce a new Home screen message that informs users on how to install the app widget within iOS system.

**Steps to test this PR**:
1. Make sure you run at least iOS 14.0.
2. Since the feature is part of the experiment make sure to comment out the variantManager code in HomeMessageStorage and in Settings.
3. You should be able to see the home message on the Home screen.
4. Once entered you can see the modal with instructions on how to add the app widget.
5. Same screen can be seen from the Settings under the "Add Widget to Home Screen" cell.
3. Once dismissed it should never appear again.

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPad

**OS Testing**:

* [ ] iOS 13 (the feature shouldn't be visible there)
* [ ] iOS 14
* [ ] iOS 15

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme


Core/VariantManager.swift
```diff
@@ -68,9 +75,12 @@ public struct Variant {
         
         Variant(name: "ma", weight: 1, isIncluded: When.isIOS15EnglishWithOnDeviceSpeech, features: [.voiceSearch]),
```
brindy: I spoke to @Bunn - he's got enough installs now so you could set the weight of `ma` and `mb` to zero in order to maximise the number of installations you get when your experiment starts rolling out.

DuckDuckGo/HomeMessageView.swift
```diff
@@ -0,0 +1,183 @@
...
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import SwiftUI
+
+struct RoundedRectStyle: ButtonStyle {
+    var foregroundColor: Color
+    var backgroundColor: Color
```
brindy: Can these be `let` instead?

DuckDuckGo/HomeMessageView.swift
```diff
@@ -0,0 +1,183 @@
...
+            .padding(.horizontal, Const.Padding.buttonHorizontal)
+            .padding(.vertical, Const.Padding.buttonVertical)
+            .foregroundColor(configuration.isPressed ? foregroundColor.opacity(0.5) : foregroundColor)
+            .background(backgroundColor)
+            .cornerRadius(Const.Radius.corner)
+    }
+}
+
+struct HomeMessageView: View {
+    var viewModel: HomeMessageViewModel
```
brindy: `let`

DuckDuckGo/HomeMessageView.swift
```diff
@@ -0,0 +1,183 @@
...
+    static let button = Color(UIColor.cornflowerBlue)
+    static let background = Color("HomeMessageBackgroundColor")
+    static let shadow = Color("HomeMessageShadowColor")
+}
+
+private extension Image {
+    static let dismiss = Image("HomeMessageDismissIcon")
+}
+
+private enum Const {
```
brindy, line 136: We've avoided using `enum` like this, and generally prefer a `struct`.  A enum without a case feels wrong.  Is there a technical reason for this or just preference?

If it's just preference, let's use a struct and consider adding it to our style guide.

DuckDuckGo/HomeMessageViewModel.swift
```diff
@@ -0,0 +1,41 @@
...
+
+struct HomeMessageButtonViewModel {
+    enum ActionStyle {
+        case `default`
+        case cancel
+    }
+    
+    let title: String
+    let action: () -> Void
+    var actionStyle: ActionStyle = .default
```
brindy, line 39: Make the action closure the last thing in the list so that you can use the closure style when constructing.

DuckDuckGo/WidgetEducationView.swift
```diff
@@ -0,0 +1,137 @@
...
+                }
+                .padding(.horizontal)
+                .padding(.top, Const.Padding.top)
+            }
+        }
+    }
+    
+    private var secondParagraphText: Text {
+        // https://stackoverflow.com/questions/62168292/what-s-the-equivalent-to-string-localizedstringwithformat-for-swiftuis-lo
+        Text("addWidget.settings.secondParagraph.\(Image.add)")
```
brindy, line 54: This seems like magic, but OK! :)

brindy (approved):
Test on a couple of different simulators, including iPad, and with double length text enabled.

A few minor comments, but generally LGTM! Nice work.

Feel free to merge once you fix them and get a thumbs up on your product review.

DuckDuckGo/WidgetEducationView.swift
jaceklyp: I wish it worked and looked different :(

DuckDuckGo/HomeMessageView.swift
jaceklyp: Yeah, so I was following the RW's guidelines in that regard: https://github.com/raywenderlich/swift-style-guide#constants

There are couple of benefits which are listed here in people's answers (which the biggest are: you don't have to maintain empty private init and structs can have private stored properties)
https://forums.swift.org/t/static-let-in-enum-vs-struct

Let me know what you think.

brindy: Ah ok, that makes sense.  I'll start doing that too moving forward. :)

PR #1048, opened 2022-02-04
Feature/bunn/download manager

Bunn:
Task/Issue URL: https://app.asana.com/0/1201188456623839/1201600067271023/f
Tech Design URL: https://app.asana.com/0/481882893211075/1201601635648875/f
CC:

**Description**:
Add DownloadsManager to the project. Note, this PR is not to me merged on develop, but on feature/downloads

**Steps to test this PR**:
1. Tap on a Wallet item like [this one](https://www.passsource.com/pass/register.php?hashedSerialNumber=eNortjIysVIKcAmuKPD3Ci8KDyjMCwzwT8-2qCwOtLVVsgZcMKbZCjU,&psRegisteredFlag=1&psMessage=&) and see if the browser presents the wallet pass
2. Tap on downloads like this [example](https://www.thinkbroadband.com/download) and check the logs to see if the download is working (Also check tmp and documents folder to see the download)
3. Tap on multiple downloads with the same name to see if the app correctly appends a unique number to the file
4. Run tests and see if they are passing

**Device Testing**:

* [x] iPhone 13
* [x] iPad Pro

**OS Testing**:

* [x] iOS 13
* [x] iOS 14
* [x] iOS 15

---
###### Internal references:


DuckDuckGo/DownloadsManager.swift
```diff
@@ -0,0 +1,168 @@
...
+
+extension DownloadsManager: DownloadDelegate {
+    func downloadDidFinish(_ download: Download, error: Error?) {
+        moveToDownloadFolderIfNecessary(download)
+        var userInfo: [AnyHashable: Any] = [UserInfoKeys.download: download]
+        if let error = error {
+            userInfo[UserInfoKeys.error] = error
+        }
+        downloadList.remove(download)
+        notificationCenter.post(name: .downloadFinished, object: nil, userInfo: userInfo)
```
miasma13, line 163: Not sure if it will impact the later implementation but I see a small inconsistency here. We include the `download` object as part of the userInfo dict, but if the file move in l:155 was performed the `location` in the `download` object will still point at the temporary location.

DuckDuckGo/Download.swift
```diff
@@ -0,0 +1,126 @@
...
+                  mimeType: MIMEType,
+                  fileName: String,
+                  temporary: Bool,
+                  delegate: DownloadDelegate? = nil) {
+      
+        self.delegate = delegate
+        self.filename = fileName
+        self.mimeType = mimeType
+        self.temporary = temporary
+        self.session = downloadSession
```
miasma13: It would be cleaner to maintain order/ priority of the vars (both for arguments and for setting them) as they are defined in lines 32-39

DuckDuckGo/Download.swift
```diff
@@ -0,0 +1,126 @@
...
+    
+    private func updateState() {
+        self.state = self.session.task?.state ?? .completed
+    }
+    
+    deinit {
+        print("Download \(id) \(filename) deinit")
+    }
+    
+    private func renameFile(_ oldPath: URL, name: String) -> URL? {
```
miasma13: Maybe `func renameFile(_ oldPath: URL, newFilename: String)`?

DuckDuckGo/Download.swift
```diff
@@ -0,0 +1,126 @@
...
+            return newPath
+        } catch {
+            return nil
+        }
+    }
+}
+
+extension Download: DownloadSessionDelegate {
+  
+    func urlSession(_ session: URLSession, downloadTask: URLSessionDownloadTask, didFinishDownloadingTo location: URL) {
```
miasma13: Just a suggestion for implementation here (and the DownloadSessionDelegate definition) to placing the methods in a chronological order (As they occur for the download) - first method 3 followed by 1 and 2. It is usually easier to follow in that manner.

DuckDuckGo/Download.swift
```diff
@@ -0,0 +1,126 @@
...
+    }
+    
+    @Published private(set) var state: URLSessionTask.State = .suspended
+    @Published private(set) var bytesWritten: Int64 = 0
+    @Published private(set) var totalBytesWritten: Int64 = 0
+    @Published private(set) var totalBytesExpectedToWrite: Int64 = 0
+    
+    required init(downloadSession: DownloadSession,
+                  mimeType: MIMEType,
+                  fileName: String,
```
miasma13: `fileName` -> `filename`. We should be consistent and always use one form. `Filename` seems to be aligned with approach found across iOS APIs.

DuckDuckGo/DownloadMetaData.swift
```diff
@@ -0,0 +1,46 @@
...
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+
+struct DownloadMetaData {
+    let fileName: String
```
miasma13: `fileName` -> `filename`

DuckDuckGo/DownloadMetaData.swift
```diff
@@ -0,0 +1,46 @@
...
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+
+struct DownloadMetaData {
```
miasma13: It should be spelled `DownloadMetadata`. Please check all other introcuded uses of (e.g. in variable names) of `metadata` keyword.

DuckDuckGo/DownloadSession.swift
```diff
@@ -0,0 +1,93 @@
...
+import Foundation
+import WebKit
+
+protocol DownloadSessionDelegate: AnyObject {
+    
+    func urlSession(_ session: URLSession, downloadTask: URLSessionDownloadTask, didFinishDownloadingTo location: URL)
+    
+    func urlSession(_ session: URLSession, task: URLSessionTask, didCompleteWithError error: Error?)
+    
+    func urlSession(_ session: URLSession,
```
miasma13: Noted in earlier comment. I'd suggest chronological order in how they are invoked: 3, 1, 2.

DuckDuckGo/DownloadSession.swift
```diff
@@ -0,0 +1,93 @@
...
+                    downloadTask: URLSessionDownloadTask,
+                    didWriteData bytesWritten: Int64,
+                    totalBytesWritten: Int64,
+                    totalBytesExpectedToWrite: Int64)
+}
+
+class DownloadSession: NSObject {
+    private(set) var session: URLSession?
+    private(set) var cookieStore: WKHTTPCookieStore?
```
miasma13: Couldn't they be fully `private`?

DuckDuckGo/DownloadsManager.swift
```diff
@@ -0,0 +1,168 @@
...
+                temporaryFile = false
+            }
+        }
+        
+        let session: DownloadSession
+        if let downloadSession = downloadSession {
+            session = downloadSession
+        } else {
+            session = DownloadSession(metaData.url, cookieStore: cookieStore)
+        }
```
miasma13: I know it is probably a cosmetic thing but I'd kept order of the function's arguments tied to their priority and keep that order they are refereed in body (if possible).

DuckDuckGo/DownloadsManager.swift
```diff
@@ -0,0 +1,168 @@
...
+    
+    var downloadsFolderFiles: [URL]? {
+        try? FileManager.default.contentsOfDirectory(at: downloadsFolder, includingPropertiesForKeys: nil)
+    }
+    
+    init(_ notificationCenter: NotificationCenter = NotificationCenter.default) {
+        self.notificationCenter = notificationCenter
+    }
+    
+    func setupDownload(_ navigationResponse: WKNavigationResponse,
```
miasma13: I think we should not omit the first params name and would benefit in followingSwift's factory name convention of prefixing it with 'make'. I think setup is not painting the full picture here as we are really creating (plus setting up) the download object.
Maybe: `makeDownload(navigationResponse: ...)`

DuckDuckGo/DownloadsManager.swift
```diff
@@ -0,0 +1,168 @@
...
+                       temporary: Bool? = nil) -> Download? {
+        
+        guard let metaData = downloadMetaData(for: navigationResponse) else { return nil }
+        
+        let temporaryFile: Bool
+        if let temporary = temporary {
+            temporaryFile = temporary
+        } else {
+            switch metaData.mimeType {
+            case .reality, .usdz, .passbook:
```
miasma13, line 61: Maybe we can extract it to some static var that would be better describing why we do it?

DuckDuckGo/DownloadsManager.swift
```diff
@@ -0,0 +1,168 @@
...
+        
+        if list.contains(newFilename) {
+            let newSuffix = counter + 1
+            return convertToUniqueFilename(filename, counter: newSuffix)
+        } else {
+            return newFilename
+        }
+    }
+    
+    private func fileName(for navigationResponse: WKNavigationResponse) -> String {
```
miasma13: `fileName` -> `filename`

DuckDuckGo/DownloadsManager.swift
miasma13: Or maybe even connect it with `FilePreviewHelper`?

DuckDuckGo/QuickLookPreviewHelper.swift
```diff
@@ -0,0 +1,61 @@
...
+
+import UIKit
+import QuickLook
+
+class QuickLookPreviewHelper: FilePreview {
+    private weak var viewController: UIViewController?
+    private let filePath: URL
+
+    private lazy var qlPreview: QLPreviewController = {
+        let p = QLPreviewController()
```
miasma13: Please avoid single letter variables.

DuckDuckGo/FilePreviewHelper.swift
```diff
@@ -0,0 +1,37 @@
...
+        guard let filePath = download.location else { return nil }
+        switch download.mimeType {
+        case .usdz, .reality:
+            return QuickLookPreviewHelper(filePath, viewController: viewController)
+            
+        case .passbook:
+            return PassKitPreviewHelper(filePath, viewController: viewController)
+            
+        default:
+            return nil
```
miasma13: Maybe we want to also use `QuickLookPreviewHelper` in the `default` scenario?

DuckDuckGo/TabViewController.swift
```diff
@@ -925,11 +967,43 @@ extension TabViewController: WKNavigationDelegate {
...
     }
     
+    /*
+     Some files might be previewed by webkit but in order to share them
+     we need to download them first.
+     This method stores the temporary download or clears it if necessary
+     */
+    private func setupOrClearTemporaryDownload(for navigationResponse: WKNavigationResponse) {
+        let downloadManager = AppDependencyProvider.shared.downloadsManager
+
```
miasma13, line 1007: I suggest putting this function below the function it is first invoked in.

DuckDuckGo/TabViewController.swift
```diff
@@ -925,11 +967,43 @@ extension TabViewController: WKNavigationDelegate {
...
+        }
+    }
+    
     func webView(_ webView: WKWebView,
                  decidePolicyFor navigationResponse: WKNavigationResponse,
                  decisionHandler: @escaping (WKNavigationResponsePolicy) -> Void) {
-        decisionHandler(.allow)
-        url = webView.url
+      
+        guard !navigationResponse.canShowMIMEType else {
```
miasma13: I would see plain `if` to be more descriptive for the branching in logic we do here (and reserve guards more for required condition check). We need to make this altered webview loading logic super clear.

DuckDuckGo/TabViewController.swift
```diff
@@ -925,11 +967,43 @@ extension TabViewController: WKNavigationDelegate {
...
+            url = webView.url
+            decisionHandler(.allow)
+            return
+        }
+       
+        let downloadManager = AppDependencyProvider.shared.downloadsManager
+        let cookieStore = webView.configuration.websiteDataStore.httpCookieStore
+        
+        if let download = downloadManager.setupDownload(navigationResponse, cookieStore: cookieStore) {
+            downloadManager.startDownload(download)
```
miasma13: I think that according to the design we need to prompt the user for the download to start.

DuckDuckGo/TabViewController.swift
```diff
@@ -108,6 +108,8 @@ class TabViewController: UIViewController {
     // In certain conditions we try to present a dax dialog when one is already showing, so check to ensure we don't
     var isShowingFullScreenDaxDialog = false
     
+    var temporaryDownload: Download?
```
miasma13: Thinking about it, maybe a more descriptive name would be better here. Like `currentlyPreviewedFileDownload` or `downloadForCurrentlyPreviewedFile` or `downloadForPreviewedFile` or `temporaryDownloadForPreviewedFile`

DuckDuckGo/TabViewController.swift
```diff
@@ -925,11 +967,43 @@ extension TabViewController: WKNavigationDelegate {
         }
     }
     
+    /*
+     Some files might be previewed by webkit but in order to share them
+     we need to download them first.
+     This method stores the temporary download or clears it if necessary
+     */
+    private func setupOrClearTemporaryDownload(for navigationResponse: WKNavigationResponse) {
```
miasma13, line 1005: I wonder if we can be more clear in the function name about the purpose of it. Maybe something along the lines of `makeTemporaryDownloadForPreviewedFile(with navigationResponse: ...)`?

miasma13 (changes requested):

DuckDuckGo/FilePreviewHelper.swift
Bunn: OK, I'll remove the `case .usdz, .reality:` then to make it easier to read.

DuckDuckGo/TabViewController.swift
Bunn: `makeTemporaryDownloadForPreviewedFile` doesn't make it clear that it's going to `nil` out the var in case the temporary reference shouldn't exist.  `make` gives the impression that there's also something coming out of it.

DuckDuckGo/TabViewController.swift
Bunn: I like `temporaryDownloadForPreviewedFile` thanks

DuckDuckGo/TabViewController.swift
Bunn: Yep, this is left out intentionally since I didn't want to change anything related to the UI on this PR besides the file preview. I'll handle this prompt together on the notification toast PR that will follow

DuckDuckGo/DownloadsManager.swift
Bunn: I don't know if I agree with this, you're then making an implicit dependency on the code layout with what it does, as in, if something changes in the body for any reason the signature would need to change making maintenance more complicated (updating all callsites) without a real benefit.

As for their priority on the argument list, given that 3 out of 4 are optional, I think this requirement is already met, and defining their priorities seems a bit arbitrary since they exist to solve different things.

DuckDuckGo/TabViewController.swift
miasma13: Fair point, I agree.

miasma13 (approved):
Looks great 👍

PR #1003, opened 2021-11-15
Voice search

Bunn:
Task/Issue URL: https://app.asana.com/0/72649045549333/1200934152050177/f
Tech Design URL:
CC:

**Description**:
Adds support for voice-search on iOS 15+ devices (and supported locations)

**Steps to test this PR**:
**Happy Path**
1. On a supported device/locale, check if the microphone button is visible in the address bar
2. Tap on it, grant the privacy permission + mic permission
3. Speak a work or a sentence 
4. Search should automatically be triggered

**Unsupported device**
1. On a device with iOS 14 or lower, check if the microphone is visibel in the address bar
2. It shouldn't be

**Deny Privacy permission**
1. On a supported device/locale, check if the microphone button is visible in the address bar
2. Tap on it, deny the privacy dialog permission
3. Dialog should dissapear 
4. Tap on it again
5. Same dialog should be visible until you approve it

**Deny Mic permission**
1. On a supported device/locale, check if the microphone button is visible in the address bar
2. Tap on it, accept the privacy dialog permission
3. Deny microphone permission 
3. Dialog should dissapear 
4. Tap on it again
5. Alert with a settings button should appear
6. Change settings/allow permission and try again

**Omnibar changes**
* On devices that **do not** support voice search, the should be no changes;
* On devices that do support voice search there were some changes:
     * Empty Tabs:
         *  Magnifying glass and mic should be visible on the non-selected state
         *  Magnifying glass should disappear once the bar is selected, mic icon should be visible
         * Mic should disappear once the user starts typing, cancel button should be visible
    * Non empty tabs
        * If the reload button is visible, the mic shouldn't be
        * Cancel button and mic should be visible when selecting the tab
        *  Mic icon should disappear once the user starts typing, cancel should be visible

**Scripts**
Build scripts were changed in order to support SwiftUI, more specifically Update Localizable.strings and Install Fonts. Make sure these are still working

**Copy Testing**: TBD

* [ ] Use of correct apostrophes in new copy, ie `’` rather than `'`

**Orientation Testing**:

* [x] Portrait
* [x] Landscape

**Device Testing**:

* [x] iPhone SE (1st Gen)
* [x] iPhone 11
* [x] iPhone 6S
* [x] iPad Pro M1

**OS Testing**:

* [ ] iOS 13
* [x] iOS 14
* [x] iOS 15

**Theme Testing**:

* [x] Light theme
* [x] Dark theme

---
###### Internal references:
[...]

PR #1083, opened 2022-03-07
iOS Voice Search - Custom Dialog & Experiment

Bunn:
Task/Issue URL: https://app.asana.com/0/72649045549333/1200934152050177/f

**Description**:
This PR changes 4 things:
* Removes experiment related code. i.e: Feature will be released to everyone
* Removes custom UIAlert dialog that was displayed before the system dialog
* Moves all the SFSpeechRecognizer availability check to a background queue
* Keeps the UI updated depending on the SFSpeechRecognizer availability callbacks by it's delegate

**Steps to test this PR**:
1. On a fresh install, tap on the microphone button
2. You should only see the system dialog asking for mic permission

------
1. Tap on the mic button on the Omnibar and perform a voice search (please use a real device)
2. Check if the voice search was triggered correctly 

------
1. Check that there are no experiment condition being applied for this feature
-----

1. Check that the microphone button is *not* available on iOS 14 or lower
-----
Original PR for context: https://github.com/duckduckgo/iOS/pull/1003

**Device Testing**:

* [x] iPhone SE (1st Gen)
* [x] iPhone 6S
* [x] iPhone 13

**OS Testing**:

* [x] iOS 15

**Theme Testing**:

* [x] Light theme
* [x] Dark theme


DuckDuckGo/LargeOmniBarState.swift
```diff
@@ -54,6 +48,7 @@ struct LargeOmniBarState {
         var onBrowsingStoppedState: OmniBarState { return self }
         var onEnterPadState: OmniBarState { return self }
         var onEnterPhoneState: OmniBarState { return SmallOmniBarState.HomeEmptyEditingState() }
+        var onReloadState: OmniBarState { return BrowsingNonEditingState() }
```
jaceklyp, line 51: return is not needed, even though it matches the convention from lines above I believe we shouldn't use it for 1-liners - actually it would be good to have a rule for that, wdyt?

DuckDuckGo/OmniBar.swift
```diff
@@ -162,6 +172,12 @@ class OmniBar: UIView {
         onQuerySubmitted()
     }
     
+    @objc private func reloadSpeechRecognizerAvailability() {
+        assert(Thread.isMainThread)
```
jaceklyp, line 176: do we need this assert here?

DuckDuckGo/OmniBar.swift
```diff
@@ -162,6 +172,12 @@ class OmniBar: UIView {
...
+        self.state = self.state.onReloadState
+        self.refreshState(self.state)
```
jaceklyp: you can get rid of selfs here

DuckDuckGo/SmallOmniBarState.swift
```diff
@@ -48,6 +48,7 @@ struct SmallOmniBarState {
         var onBrowsingStoppedState: OmniBarState { return self }
         var onEnterPadState: OmniBarState { return LargeOmniBarState.HomeEmptyEditingState() }
         var onEnterPhoneState: OmniBarState { return self }
+        var onReloadState: OmniBarState { return HomeEmptyEditingState() }
```
jaceklyp, line 51: same comment regarding returns

DuckDuckGo/SpeechRecognizer.swift
```diff
@@ -21,18 +21,48 @@ import Foundation
 import Speech
 import Accelerate
 
-class SpeechRecognizer: SpeechRecognizerProtocol {
-    private var audioEngine = AVAudioEngine()
+protocol SpeechRecognizerDelegate: AnyObject {
+    func speechRecognizer(_ speechRecognizer: SpeechRecognizer, availabilityDidChange available: Bool)
+}
+
+class SpeechRecognizer: NSObject, SpeechRecognizerProtocol {
```
jaceklyp: please make it final

DuckDuckGo/SpeechRecognizer.swift
```diff
@@ -21,18 +21,48 @@ import Foundation
...
+            delegate?.speechRecognizer(self, availabilityDidChange: isAvailable)
+        }
+    }
+    
+    override init() {
+        operationQueue = OperationQueue()
+        operationQueue.qualityOfService = .userInteractive
+        
+        speechRecognizer = SFSpeechRecognizer()
+        speechRecognizer?.queue = operationQueue
```
jaceklyp, line 47: please either create two factory methods out of it, i.e. func makeQueue() -> OperationQueue() or instantiate both queue and recognizer inside lazy vars, up to you

jaceklyp (changes requested):

DuckDuckGo/LargeOmniBarState.swift
Bunn: yep, but since the other ones have return I thought it was better to follow the pattern. But I agree we should have a rule for that. I don't think it's part of this PR though.

DuckDuckGo/OmniBar.swift
Bunn: Defensive programming =D
This call changes the UI so it needs to be called on the main thread, I like asserts because:
1 - states the intention. Someone reading it knows it should be done on the main thread
2 - Easier to catch bugs during development

DuckDuckGo/SpeechRecognizer.swift
Bunn: I can't do that because they are let properties so they need to be initialized inside the init, and if I change to lazy they would not be thread safe anymore.

I don't think it's worth to do these changes given that the current initialization of those properties are fairly simple and make contextual sense with init method.

jaceklyp (approved):
LGTM!

PR #1100, opened 2022-03-29
Change the logic for tab preparation for data clearing

miasma13:
Task/Issue URL: https://app.asana.com/0/414709148257752/1202026066957933/f
Tech Design URL:
CC:

**Description**:
During last two weeks we observed increased reports of `Snapshots not cleared error` that started rising since around March 13. Upon investigation it turned out that the increased errors are caused by recent change in data clearing logic. Additionally that change introduced two additional issues: memory leak on data clearing and fire animation stutter. 

**Steps to test this PR**:
_Snapshots not cleared error_
1. Open multiple tabs
2. Tap the Fire button but as the animation starts minimise the app
3. Reopen the app - the `assertTabPreviewCount()` will make a check and trigger the error if snapshot check condition is met

_TabViewController leak_
1. Open multiple tabs
2. Tap the Fire button
3. Check memory graph for TabViewController instances count (should be equal to number of tabs after clearing - 1)

_Fire animation not smooth_
1. Open multiple tabs (e.g. check for 5/10/15/20)
2. Tap the Fire button
3. The fire animation should always be smooth and and without any initial delay

_Check local storage clearing_
Open local storage test page http://privacy-test-pages.glitch.me/features/local-storage.html and run it. The value is stored in browsers local storage and is retrieved on future test page visits. Fire button should clear the local storage so when visiting the test page the initial result is `undefined`. Please check those variants:
1. Open test page and run the test. Tap fire button.
2. Open test page and run the test. Tap start test multiple times. Tap fire button.
3. Open test page and run the test. Tap start test multiple times. Open multiple other tabs. Tap fire button.
4. Open test page and run the test. Tap start test multiple times. Open multiple other tabs including other tabs with test page. Tap fire button.
In all above variants upon the next visit to the test page the initial result should be `undefined`.

### DIFF
--- a/DuckDuckGo/MainViewController.swift
+++ b/DuckDuckGo/MainViewController.swift
@@ -1638,13 +1638,15 @@ extension MainViewController: AutoClearWorker {
         let spid = Instruments.shared.startTimedEvent(.clearingData)
         Pixel.fire(pixel: .forgetAllExecuted)
         
+        tabManager.prepareAllTabsExceptCurrentForDataClearing()
+        
         fireButtonAnimator?.animate {
-            self.tabManager.prepareTabsForDataClearing {
-                self.stopAllOngoingDownloads()
-                self.forgetData()
-                DaxDialogs.shared.resumeRegularFlow()
-                self.forgetTabs()
-            }
+            self.tabManager.prepareCurrentTabForDataClearing()
+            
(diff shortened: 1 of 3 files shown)


jaceklyp (approved):
LGTM! nice work 🥇

PR #1102, opened 2022-03-30
Import / Export Bookmarks

amddg44:
Task/Issue URL: https://app.asana.com/0/72649045549333/1201780742295295/f
Tech Design URL:
CC: @brindy 

**Description**:

New functionality / enhancement that allows the importing and exporting of bookmarks to / from html files

Additionally I've made a few modifications to BookmarksCoreDataStorage + added tests to support this new functionality to allow for:
 * MockBookmarksCoreDataStorage so that existing functionality and new functionality to be unit tested
 * swift concurrency (async await)

**Steps to test this PR**:

**Test Scenario 1**

1. Start with a previous build of the app, with favorites, bookmarks & bookmark folders
2. Install this PR
3. Export bookmarks 
4. Confirm appropriate export success toast message is shown to user 
a. Save to Files = "Your bookmarks have been exported." 
b. Shared to any other app = "Your bookmarks have been shared."
5. Delete + reinstall the app / delete all the bookmarks & favorites in the app
6. Import the bookmarks file from step 3
7. All previous favorites & bookmarks should be re-added
8. Confirm import success toast message is shown to user

**Test Scenario 2**
1. Ensuring the app has a mix of favorites, top level bookmarks & bookmark folders, export bookmarks to file
2. Confirm export success toast message is shown to user
3. Import the bookmarks file
4. Confirm import success toast message is shown to user
5. Top level favorites and bookmarks should not be duplicated, only bookmark folders

**Test Scenario 3**
1. Export bookmarks from another browser e.g. Chrome / Safari
2. Import the bookmarks into the app
3. Confirm import success toast message is shown to user
4. Bookmarks and folders should import correctly

**Test Scenario 4**
1. Export bookmarks from the app
2. Confirm export success toast message is shown to user
3. Import the bookmarks into another browser e.g. Chrome or Safari and confirm bookmarks are imported

**Test Scenario 5**
1. Try to import an invalid html bookmark file (there is one in udner Test -> Mock files)
2. Confirm export failed toast message is shown to user
	
**Test Scenario 6** 
1. Navigate to a bookmark subfolder.
2. Confirm More button is disabled

**Test Scenario 7** 
1. Delete all bookmarks, favorites and folders
2. Confirm the More -> Export button is disabled

**Test Scenario 8** 
1. Add items to Safari reading list and export bookmarks 
2. Import the Safari bookmarks html and confirm the reading list is not imported

**Test Scenario 9** 
1. Import bookmarks into the app
2. Tap on an imported bookmark that does not have a favicon to load that bookmark in browser
3. Go back to the list of bookmarks and confirm the bookmark now has a favicon   

**Test Scenario 10** 
1. Using a device running iOS 13, confirm no import or export functionality is available i.e. More button is not in the toolbar and the import footer button is not visible2. 

**Unit Tests**
- Run Unit Tests for `BookmarkCoreDataStorageTests`, `BookmarksImporterTests` & `BookmarksExporterTests`  

**Pixels**

Test the following pixels fire appropriately

- case bookmarkImportSuccess = "m_bi_s"
- case bookmarkImportFailure = "m_bi_e"
- case bookmarkExportSuccess = "m_be_a"
- case bookmarkExportFailure = "m_be_e"

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone
* [ ] iPad

**OS Testing**:

* [ ] iOS 13
* [ ] iOS 14
* [ ] iOS 15

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:


Core/BookmarkOrFolder.swift
```diff
@@ -0,0 +1,58 @@
...
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+
+public class BookmarkOrFolder {
+    public let name: String
+    public let type: String
```
brindy: Can this be an enum?

Core/Pixel.swift
```diff
@@ -164,6 +164,11 @@ public enum PixelName: String {
     case bookmarkEditedAtTopLevel = "m_bookmark_edited_at_top_level"
     case bookmarkEditedInSubfolder = "m_bookmark_edited_in_subfolder"
     
+    case bookmarkImportSuccess = "m_bi_s"
+    case bookmarkImportFailure = "m_bi_e"
+    case bookmarkExportSuccess = "m_be_a"
+    case bookmarkExportFailure = "m_be_e"
```
brindy, line 170: We decided to be more transparent about pixels, so these values should be more explicit (like the ones above), unless there was a good reason?

brindy (commented):
I still need to test this, but this looks really good, nice job, just a couple of minor comments in the mean time.

Core/BookmarkOrFolder.swift
amddg44: No problem!  I've done it as an enum, hope that's what you had in mind?  Just FYI this class is the Mac app version with the only difference being on Mac it was a struct (on iOS it is a class since I'm mutating / continually adding to the children)

Do you like the enum approach better or more consistency with MacOS (or maybe this doesn't matter til later or not at all?)

Thanks!

Core/Pixel.swift
amddg44: These were the pixels defined in Android so I was matching platforms. I didn't know about the change re pixel transparency but am happy to change values to something more explicit

brindy: Alright, happy for them to stay the same then.  They were probably defined before the transparency guideline. :)

Core/BookmarksCoreDataStorage.swift
```diff
@@ -714,6 +904,21 @@ extension BookmarksCoreDataStorage {
     }
 }
 
+// MARK: Constants
+extension BookmarksCoreDataStorage {
+    struct Constants {
```
brindy: worth changing this to enum while you're here?

Core/BookmarksCoreDataStorage.swift
```diff
@@ -135,6 +124,28 @@ public class BookmarksCoreDataStorage {
             }            
         }
     }
+
+    static internal func rootFolderManagedObject(_ context: NSManagedObjectContext) -> BookmarkFolderManagedObject {
+        guard let bookmarksFolder = NSEntityDescription.insertNewObject(forEntityName: "BookmarkFolderManagedObject", into: context)
+                as? BookmarkFolderManagedObject else {
+            // TODO - this should not be fatal
```
brindy: TODO

Core/BookmarksCoreDataStorage.swift
```diff
@@ -135,6 +124,28 @@ public class BookmarksCoreDataStorage {
...
+        }
+
+        bookmarksFolder.isFavorite = false
+        return bookmarksFolder
+    }
+
+    static internal func rootFavoritesFolderManagedObject(_ context: NSManagedObjectContext) -> BookmarkFolderManagedObject {
+        guard let bookmarksFolder = NSEntityDescription.insertNewObject(forEntityName: "BookmarkFolderManagedObject", into: context)
+                as? BookmarkFolderManagedObject else {
+            // TODO - this should not be fatal
```
brindy: TODO

brindy (commented):
Really close, and looks great.  Just the performance thing we discussed, and a few minor comments.

Core/BookmarksCoreDataStorage.swift
jaceklyp: please see: https://app.asana.com/0/0/1201927187914110/f

Core/BookmarksCoreDataStorage.swift
jaceklyp: same as above ^

Core/BookmarksCoreDataStorage.swift
amddg44: 👍 have made that change 🙂

Core/BookmarksCoreDataStorage.swift
amddg44: I’ve removed the TODO; I had left it as a reminder / question for @brindy about fatalError in this particular scenario

@jaceklyp re: the CoreData insert that’s legacy code that I encapsulated into a method only for reusability and to make it unit testable. I don’t think I should make that change in this PR as I don’t know what else in the app it can impact / break and don’t know the test scenarios. The core data change should be done in a different PR I think.

Here is a screenshot of what code was before and after 
image
image

jaceklyp: It's really up to you :) Personally I prefer to fix the broken code when I see one and this is really a syntactic sugar change from:
`guard let bookmarksFolder = NSEntityDescription.insertNewObject(forEntityName: "BookmarkFolderManagedObject", into: context)
                as? BookmarkFolderManagedObject else {`
to:
`let bookmarksFolder: BookmarkFolderManagedObject = context.insertObject()`

And it helps with your issue because then you don't have to use fatalErrors anymore, because the code is safely unwrapped for you and it doesn't have any negative effect on anything else - meaning - just checking if this code works is sufficient enough.

catehstn:
Why has this PR been awaiting review for 14 days?

samsymons:
@catehstn This one missed getting the correct person assigned, so it was unclear that it was ready for review again (GitHub doesn't surface review requests when you're viewing the PR list). Anya pinged Brindy and I for review earlier and cleared it up, we're on the case now.

samsymons (approved):
This looks good to me, awesome work 👏

I had one note from testing, which isn't a blocker: when testing import on my phone, I got an error with the message `The operation couldn’t be completed. Invalid argument` any time I tried to export to the Files app. I came back to this a little bit later and it has since started working - not worth following up on, but I wanted to note it here in case it sounded familiar.

amddg44:
> This looks good to me, awesome work 👏
> 
> I had one note from testing, which isn't a blocker: when testing import on my phone, I got an error with the message `The operation couldn’t be completed. Invalid argument` any time I tried to export to the Files app. I came back to this a little bit later and it has since started working - not worth following up on, but I wanted to note it here in case it sounded familiar.

Hi Sam, thanks for reviewing!  I've never run into the error message -- thanks for mentioning.  If you get a chance and can send over your import file I'd be happy to take a look.  Thanks again!

PR #1107, opened 2022-04-01
Fix for not sending errors inside ContentBlockerRulesManager

jaceklyp:
Task/Issue URL: https://app.asana.com/0/414235014887631/1202062200831946/f
CC: @bwaresiak 

**Description**:
Fix for not sending errors which happen inside ContentBlockerRulesManager.

**Steps to test this PR**:
1. Mess with the content rules to be sure compilation fails.
2. Notice if we send any pixels related to failing compilation.


miasma13 (approved):
Good catch! 🥇

PR #1115, opened 2022-04-08
Voice search settings toggle

Bunn:
Task/Issue URL: https://app.asana.com/0/414709148257752/1202070298493931/f

**Description**:
Adds a setting option to disable/enable voice-search

**Steps to test this PR**:
(on iOS 15)
1. Open the app, check if the voice search is enabled (it should)
2. Open the settings panel, check if the "Hide Voice Search" is disabled (it should)
3. Enable the "Hide Voice Search" option and check that the mic button is not visible on the Omnibar 

(on iOS 14 or lower)
1. Open the settings panel, check if the "Hide Voice Search" cell is hidden. (It should be hidden, since voice-search only works iOS 15+ we should not display this option on older iOS versions)

**OS Testing**:

* [x] iOS 13
* [x] iOS 14
* [x] iOS 15

**Theme Testing**:

* [x] Light theme
* [x] Dark theme


Core/UserDefaultsPropertyWrapper.swift
```diff
@@ -73,6 +73,7 @@ public struct UserDefaultsWrapper<T> {
         
         case emailWaitlistShouldReceiveNotifications = "com.duckduckgo.ios.showWaitlistNotification"
         case unseenDownloadsAvailable = "com.duckduckgo.app.unseenDownloadsAvailable"
+        case voiceSearch = "com.duckduckgo.app.voiceSearch"
```
samsymons: One tiny request here is that we rename this to `voiceSearchEnabled` to better match its intent (and also implies that this represents a boolean). Tbh as it is right now is fine, but if we add more voice search settings later it would help to avoid confusion.

samsymons (approved):
Looks great! Tested on various iOS versions to confirm that the setting is hidden depending on feature availability. Great work 🙂

Bunn:
I'm closing this PR, more info [here](https://app.asana.com/0/1142021229838617/1202140434638867/f)
 
If we changes our mind in the future I'll re-open and reference this as code-reviewed.

PR #1145, opened 2022-05-05
Start downloads only when response has 2xx status code

miasma13:
Task/Issue URL: https://app.asana.com/0/414235014887631/1202179200624745/f

**Description**:
Some of the download responses are not from the 2xx HTTP status code ranges. For those downloads the experience for the users will be confusing - most probably it will result in not saving of the requested file but instead a HTML error page often under unknown file name.


brindy (approved):
Downloaded a couple of example files from the web and checked PDFs. LGTM!

PR #1153, opened 2022-05-13
Remove favicon on bookmark delete

jaceklyp:
Task/Issue URL: https://app.asana.com/0/414709148257752/1202224701840372/f
CC: @miasma13 

**Description**:
Fix an issue when favicon is not deleted on bookmark removal.

**Steps to test this PR**:
1. Add page to bookmark.
2. Remove the bookmark.
3. Make sure there is no favicon in the cache directory.

### DIFF
--- a/DuckDuckGo/BookmarksManager.swift
+++ b/DuckDuckGo/BookmarksManager.swift
@@ -165,22 +165,25 @@ class BookmarksManager {
     
     func delete(_ bookmarkItem: BookmarkItem, completion: BookmarkItemDeletedBackgroundThreadCompletion? = nil) {
         var reloadWidget = false
+        var bookmarkURL: URL?
         if let bookmark = bookmarkItem as? Bookmark {
-            removeFavicon(forBookmark: bookmark)
             if bookmark.isFavorite {
                 reloadWidget = true
             }
+            bookmarkURL = bookmark.url
         }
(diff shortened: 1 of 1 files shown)


DuckDuckGo/BookmarksManager.swift
```diff
@@ -165,22 +165,25 @@ class BookmarksManager {
...
             completion?(success, error)
         }
     }
     
-    private func removeFavicon(forBookmark bookmark: Bookmark?) {
-        guard let domain = bookmark?.url?.host else { return }
+    private func removeFavicon(forBookmarkURL url: URL?) {
+        guard let domain = url?.host else { return }
         
         coreDataStorage.allBookmarksAndFavoritesFlat { bookmarks in
```
THISISDINOSAUR, line 188: Is it ever possible coreDataStorage.allBookmarksAndFavoritesFlat is fetched before the deletion is properly committed to core data? (I don't remember how core data handles that kind of thing, I think the answers no but want to double check)

THISISDINOSAUR (approved):
Tested and looks good to me, I just have the one (probably dumb) question

DuckDuckGo/BookmarksManager.swift
jaceklyp: we call the completion block after a successful save, so the core data has already processed the changes and only then we call the method removeFavicon which refetches the model state, we should be safe here!

PR #1154, opened 2022-05-13
Do not populate bookmarks cache with favicons for all visited sites

jaceklyp:
Task/Issue URL: https://app.asana.com/0/414709148257752/1202223699061973/f
CC: @miasma13

**Description**:
When we import bookmarks from HTML file we want to download favicons for them not before by messing with caches.

**Steps to test this PR**:
1. Export bookmarks to HTML file.
2. Remove all bookmarks.
3. Import bookmark from HTML file.
4. Make sure that imported websites get the favicon.

### DIFF
--- a/DuckDuckGo/BookmarksViewController.swift
+++ b/DuckDuckGo/BookmarksViewController.swift
@@ -365,7 +365,15 @@ class BookmarksViewController: UITableViewController {
             case .success:
                 dataSource.bookmarksManager.reloadWidgets()
 
-                let bookmarkCountAfterImport = await dataSource.bookmarksManager.allBookmarksAndFavoritesFlat().count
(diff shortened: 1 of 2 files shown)


amddg44:
Hi 👋 

The fix you made is exactly what the Privacy team has ruled out [here](https://app.asana.com/0/0/1202061102875312/1202071586859221/f). 

Also, I've made a quick video and it doesn't resolve this issue - the favicon is still not visible in the bookmarks screen at the end of video. 

The test scenario I was using is importing a large number bookmarks into the app that were not previously bookmarked, as importing bookmarks is primarily for users who have changed device, re-installed the app or are new users and are bringing in bookmarks for the first time. So it's great that you have  another use case that you are testing however you also need to test the use case I mentioned, thanks 🙂 

https://user-images.githubusercontent.com/91189922/168596245-a1ffbea8-285e-4d4b-9c51-7c4b462afc98.mov

PR #1159, opened 2022-05-18
Fix for bookmarks favicon caching

amddg44:
Task/Issue URL: https://app.asana.com/0/414709148257752/1202223699061973/f
CC: @miasma13 

**Description**: When visiting sites ensure that only favicons for bookmarked sites are saved to the bookmarks cache

**Steps to test this PR**:
1. Visit a website that is not bookmarked.
2. Confirm that it's favicon is saved only to the .tabs cache and not to the .bookmarks cache
3. Import bookmarks from HTML file containing some sites that you have not already bookmarked
4. The imported bookmarks will have a default placeholder favicon5. 
5. Tap one of the imported bookmarks to visit that site6. 
6. Return to the bookmarks screen and confirm this bookmark now has a favicon (the same site as visible in tabs view)

---
###### Internal references:


Core/BookmarksCoreDataStorage.swift
```diff
@@ -706,6 +706,32 @@ extension BookmarksCoreDataStorage {
         })
     }
 
+    internal func containsDomain(_ domain: String,
+                                 completion: @escaping BookmarkExistsMainThreadCompletion) {
+        viewContext.perform {
+            let fetchRequest = NSFetchRequest<BookmarkManagedObject>(entityName: Constants.bookmarkClassName)
+            fetchRequest.fetchLimit = 1
+
+            fetchRequest.predicate = NSPredicate(format: "%K CONTAINS %@",
```
bwaresiak: Won't this match "tech.com" when searching for "ech.com" ?

Core/BookmarksCoreDataStorage.swift
```diff
@@ -706,6 +706,32 @@ extension BookmarksCoreDataStorage {
...
+                    #keyPath(BookmarkManagedObject.url.host),
```
bwaresiak: Also, I don't think this key path will work with SQLite store, as that store does not treat URLs as "objects", but uses `NSCoreDataStringSearch` function to match strings in DB columns. This probably will work in tests, where we are using in memory store - which is confusing, so we should better move to use SQLite-backed store in tests as well.

Nonetheless, what I mean is that we don't really narrow this to host only, so trying to match "test.com" in "https://example.com/index.html?q=atest.com" will most likely return true.

amddg44: Good point @bwaresiak. I took a look for an alternative and I see @THISISDINOSAUR  already wrote the logic for checking domains [here](https://github.com/duckduckgo/iOS/blob/65995ec09fab1a14321375431d0fc225c3161a99/DuckDuckGo/BookmarksManager.swift#L185)

I've taken the heart of this logic and added to / encapsulated within `BookmarksCoreDataStorage` and replaced the CoreData call

@THISISDINOSAUR I'm happy to refactor this in a separate PR?  I didn't want to include in this one + wanted to keep this PR on point.

Just one notable for all; I wrapped this new function with `MainActor` - when `Favicons` was calling this method I was consistently getting crashes while testing.  Just mentioning as perhaps its related to the crashes that were mentioned during NativeApps call

THISISDINOSAUR: What kind of refactoring, is there anything besides the tests?

DuckDuckGoTests/FaviconsTests.swift
```diff
@@ -98,11 +98,34 @@ class FaviconsTests: XCTestCase {
...
         
     }
+
+    func testWhenDomainIsBookmarkThenIsFaviconCached() {
+        let expectation = self.expectation(description: "isFaviconCachedForBookmarks")
+
+        let fileName = "MockFiles/DDG-icon_256x256.png"
+
+        guard let imageData = try? FileLoader().load(fileName: fileName, fromBundle: Bundle(for: type(of: self))),
+              let image = UIImage(data: imageData),
```
bwaresiak: Can we use something that is already embedded in the app e.g. `UIImage(named: "Logo")` ?

DuckDuckGo/Favicons.swift
```diff
@@ -35,7 +35,8 @@ public class Favicons {
         static let tabsCache = CacheType.tabs.create()
         static let appUrls = AppUrls()
         static let targetImageSizePoints: CGFloat = 64
-        
+        static let bookmarksCoreDataStorage = BookmarksCoreDataStorage.shared
```
bwaresiak: Can we treat this as a dependency passed through the init?

DuckDuckGo/Favicons.swift
```diff
@@ -138,14 +139,15 @@ public class Favicons {
...
-            (Constants.bookmarksCache.isCached(forKey: resource.cacheKey) || bookmarksStore.contains(domain: domain)),
-            let options = kfOptions(forDomain: domain, usingCache: .bookmarks) else {
-                loadFavicon(forDomain: domain, intoCache: .bookmarks, fromCache: .tabs)
-                return
-            }
+              let resource = defaultResource(forDomain: domain),
+              let options = kfOptions(forDomain: domain, usingCache: .bookmarks) else { return }
+
+        if !isFaviconCachedForBookmarks(forDomain: domain, resource: resource) {
+            loadFaviconForBookmarks(forDomain: domain)
```
bwaresiak, line 155: I don't think this approach is optimal, as we query all bookmarks on each navigation. I'd try to look for a ways to optimize that, like having some caching layer for bookmarks or leveraging existence of tabs cache (so call above code only in case if we insert into tabs cache - as that won't repeat the calls for subsequent navigations on that domain). There are definitely other ideas to consider as well.

DuckDuckGoTests/FaviconsTests.swift
amddg44: Ah, I didn't know about Logo - thanks for letting me know.  I'm happy to do this if you prefer; however if the `Logo` filename is every changed then the test will break.  The logo I added to mockdata is for the test only.  If we have the unit test use production `Logo`, it would be similar to creating a UnitTest that tests production data (although I can appreciate that in this case the `Logo` name is unlikely to change as often compared to something like prod JSON data).  So it's a small code smell but not too smelly so I don't have too strong an opinion

DuckDuckGo/Favicons.swift
amddg44: I spent a couple of hours on this and it cannot be done without changing the singleton nature of Favicons which I think greatly enhances the scope of this bug fix PR.  Favicons in its current state is not designed for DI.  I would recommend we create this as a new issue (Refactor FavIcons for DI + testability?) since this could cause other regressions in the app (and I don't know all the use cases / edge cases to test)

DuckDuckGo/Favicons.swift
amddg44: I thought about this and tested and I didn't notice any impact to the user experience.  Also I found `BookmarksSearchStore` which unfortunately is not in Core but when looking at this I saw this constantly makes the same call to `BookmarksCoreDataStorage` when using the app so I presumed that this trade off was acceptable.  Every time the address bar is tapped in the app it makes this CoreData call for all bookmarks.

Assuming this trade off is not acceptable is there an existing data cache structure within the app already (data cache) I could potentially leverage for this or (not)?  I can create an enhancement to the original bookmarks import export so that the domain was separated from the bookmark URL on import and cached to disk in some manner + hopefully leverage something the app already has.

Also if we go this route is this worth discussing if MacOS has this and if that can be shared in BSK?

DuckDuckGo/Favicons.swift
amddg44: I deleted my last comment and I've made this change. It would be nice in a future issue to remove the singleton so this could be DI'd as I can't use the MockBookmarksCoreDataStore for unit tests

DuckDuckGoTests/FaviconsTests.swift
bwaresiak: Logo image is already used in `loadFaviconSync` so it is related to the problem domain we are touching here - I think any image form that domain would do, logo is just the first one I've spotted. :)

DuckDuckGo/Favicons.swift
bwaresiak: Why - do you have to use `Favicons.shared` in the tests? Why not create new instance of that just for the tests?

Edit: to be more clear :) Singletons we create have no limitation on init - so in practice it is possible to instantiate these multiple times (which obviously should not happen when executing the app), but this gives us flexibility in tests. In case of unit tests, we can instantiate such class and test functionality in isolation. Obviously this won't work (or would be more difficult) for Integration or UI testing.

DuckDuckGo/Favicons.swift
bwaresiak: I think the use case for Search and for this is completely different:
 - In case of search: We cache bookmarks and the result of the query is (nearly) always used when user is typing something in autocomplete to present results. There isn't much happening other than that - so it is not affecting user flow but is part of it.
 - In case of favicon we want to apply it when needed - so when we navigate to a domain and obtain info about favicon related to that domain, then we query bookmarks to see if we need to add this favicon to the cache. User does not need that for his current flow and there is plenty of other things we are doing when loading pages, so I'd rather optimize for using less resources and making less actions. 

Much better would be to do that once per domain (host). Best solution would be to Cache bookmarks (actually - just hosts) in memory and invalidate that cache when there are changes to the store (e.g. on bookmarks did save notification).

I wouldn't think about macOS here too deeply as:
- Desktop has much more resources than mobile.
- Memory separation model between apps on the desktop is different than on mobile, so we may not want to keep bookmark URLs in memory all the time there.

DuckDuckGoTests/FaviconsTests.swift
amddg44: No problem, I've made that change :)

DuckDuckGo/Favicons.swift
amddg44: I updated my test and all the pre-existing tests to no longer use the singleton

DuckDuckGo/Favicons.swift
amddg44: I've taken a look / spent some time on this and what do you think about moving BookmarksCachingSearch to Core and utilising it's cache for favicons as well?  It seems like this could be shared but you know the app better so happy to defer to you 🙂

bwaresiak: I think it makes sense. To be on the same page (as I'm out next week and someone will need to pick this up from here):
 - We should be able to instantiate `BookmarksCachingSearch` lazily when needed, then discard it once we receive `dataDidChange` notification from `BookmarksCoreDataStorage`. It should be fine to just keep that in memory as we already do every time user taps URL bar.
 - It probably is also a good idea to change Autosuggest to use the same instance.

amddg44:
I've updated this PR with earlier discussed changes

amddg44:
I've done another push to fix a failing test.  Favicons and NotFoundCachingDownloader had a circular dependency on each other.  Unless I'm mistaken, I don't think there is a need for NotFoundCachingDownloader to have a property / knowledge of Favicons and so I've encapsulated the circular dependency issue into a Helper class and static function that is stateless

THISISDINOSAUR:
When testing what's in the caches seems like it's always correct, although when I import bookmarks they get favicons straight away:

https://user-images.githubusercontent.com/1093508/170255255-2f9d02bd-c8cb-40de-8978-eafecdf7d0b3.mp4

amddg44:
@THISISDINOSAUR I believe that is related to this PR [Remove favicon on bookmark delete](https://github.com/duckduckgo/iOS/pull/1153) so when you delete the bookmarks, the favicons are not deleting from cache and so are still there when you re-import. If you try deleting and re-installing the app you should see the correct behaviour

DuckDuckGo/BookmarksViewController.swift
```diff
@@ -357,6 +357,10 @@ class BookmarksViewController: UITableViewController {
     }
 
     func importBookmarks(fromHtml html: String) {
+        // pre-emptively deregisterForNotifications so that bookmarksCachingSearch is not saturated with notification events
+        // and constantly rebuilding while bookmarks are being imported (bookmark files could be very large)
+        bookmarksCachingSearch.deregisterForNotifications()
```
THISISDINOSAUR: Very very minor, but I'd prefer if this was moved from the view controller since it feels like something the view controller shouldn't have to know about. E.g. managed by BookmarksImporter or even just behind a method on bookmarksCachingSearch (like `prepareForImport()` or something).

THISISDINOSAUR: Also wondering does the import trigger loads of notifications in any other bookmarksCachingSearch instances that are around? I mean it won't right now since I think the same shared object is used everywhere? But in future it would be easy for someone to not release this and trip up 

How about importing firing notifications before starting and on completion that bookmarksCachingSearch can listen to and then manage what it does with that information itself?

Core/FaviconsHelper.swift
```diff
@@ -0,0 +1,34 @@
...
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+import Kingfisher
+
+struct FaviconsHelper {
```
THISISDINOSAUR, line 23: Heads up @Bunn in case I forget. I can't remember what you called your Favicon thing, but we might want to unify it with this when merging

Core/CoreDependencyProvider.swift
```diff
@@ -0,0 +1,32 @@
...
+
+import Foundation
+
+public protocol CoreDependencyProvided {
+    var bookmarksCachingSearch: BookmarksCachingSearch { get }
+}
+
+/// Provides dependencies for objects that are not directly instantiated
+/// through `init` call
+public class CoreDependencyProvider: CoreDependencyProvided {
```
THISISDINOSAUR, line 28: This is interesting, how do you plan on using it in future?

DuckDuckGo/BookmarksViewController.swift
amddg44: > Very very minor, but I'd prefer if this was moved from the view controller since it feels like something the view controller shouldn't have to know about. E.g. managed by BookmarksImporter or even just behind a method on bookmarksCachingSearch (like `prepareForImport()` or something).

Sounds good + thanks for the suggestion!

amddg44: > Also wondering does the import trigger loads of notifications in any other bookmarksCachingSearch instances that are around? I mean it won't right now since I think the same shared object is used everywhere? But in future it would be easy for someone to not release this and trip up
> 
> How about importing firing notifications before starting and on completion that bookmarksCachingSearch can listen to and then manage what it does with that information itself?

Sounds great + I've done it this way :slightly_smiling_face:

Core/CoreDependencyProvider.swift
amddg44: I saw AppDependencyProvider already in place and thought Core needed something similar in this case; ideally we would move our singletons such as Favicons to this DependencyProvider for DI in the future

THISISDINOSAUR (approved):
LGTM!

PR #1179, opened 2022-06-03
Lazy load fire button animation and decouple preloading from the app startup

miasma13:
Task/Issue URL: https://app.asana.com/0/414709148257752/1202389859384867/f

**Description**:
The fire button animation is preloaded by FireButtonAnimator and it is triggered from the MainViewController's viewDidLoad method. Due to the Lottie version change (its implementation changed from Obj-C to Swift) it impacted the processing time for loading the animation from the JSON file. This combined impacted the app's launch time.

Proposal
- Make the FireButtonAnimator lazy loaded
- Wake it on FireButton press so it is preloaded on first use (it will not impact UX as we first show action sheet and then fire the animation upon confirmation)

**Steps to test this PR**:
- After fresh app start try as quickly as possible try triggering the fire animation
- It should play without delay and there should not be any lag while it is preloaded for the first time when showing the action sheet
- There should be no lag and the animation should run when triggering the fire button from the keyboard on first use after restart

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone
* [ ] iPad

---
###### Internal references:

### DIFF
--- a/DuckDuckGo/MainViewController.swift
+++ b/DuckDuckGo/MainViewController.swift
@@ -103,7 +103,7 @@ class MainViewController: UIViewController {
     
     let gestureBookmarksButton = GestureToolbarButton()
     
-    private var fireButtonAnimator: FireButtonAnimator?
+    private lazy var fireButtonAnimator: FireButtonAnimator =  = FireButtonAnimator(appSettings: appSettings)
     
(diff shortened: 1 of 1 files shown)


bwaresiak (approved):
LGTM!

PR #1184, opened 2022-06-09
Autofill logins

THISISDINOSAUR:
Task/Issue URL: https://app.asana.com/0/72649045549333/1201600439227198/f
Tech Design URL:
CC:

**Description**:
Adds V1 of autofilling logins to iOS. This includes filling and saving (including partial credentials), and some of the settings interface. There's a comprehensive list of what should be in this here: https://app.asana.com/0/0/1202346026470299/f
And a list of things deferred until V2 here: https://app.asana.com/0/0/1202345981744059/f

Also some design review items deferred if you spot any of those: https://app.asana.com/0/0/1202374896817644/f

Also worth noting that there are strings that deliberately aren't in UserText as they shouldn't be translated yet given the internal nature of this release

**Steps to test this PR**:
----
Test email autofill (this shouldn't have changed, so you can just briefly check it's still there)
1. Sign out of email if signed in
1. Test that email autofill does not appear 
1. Sign in to email. 
1. Test email autofill still works (both duck address and generated aliases, test the generated ones are different every time).
----
Test saving logins
Test using both a login and a sign up form. I use https://www.katesclothing.co.uk/account/login and https://www.katesclothing.co.uk/account/register
1. On the first save, you should see a different screen introducing you to the idea of autofill.
1. Test saving one with a long first part of the email (e.g. loooooooooooooooooooooooooooongemail@example.com), when displaying this email it should truncate just the first part (e.g.. looooooo...@example.com)
1. Test only typing a password and saving. It should save only the password
1. Now type the same password, but with a username, it should offer to update (and end up saving as one set of credentials).
1. Now type the same username, but change the password. It should prompt to update the password
----
1. Test Autofilling logins
I use https://www.katesclothing.co.uk/account/login to test
Test saving greater than 3 on the same site, then the autofilling logins modal should have a show more button

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPad

**OS Testing**:

* [ ] iOS 13
* [ ] iOS 14
* [ ] iOS 15

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:


DuckDuckGo/AutofillInterfaceEmailTruncator.swift
```diff
@@ -0,0 +1,43 @@
...
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+
+struct AutofillInterfaceEmailTruncator {
```
brindy, line 22: This looks quick to write a unit test for?

DuckDuckGo/AutofillLoginDetailsView.swift
```diff
@@ -0,0 +1,290 @@
...
+            Section {
+                editableCell("Username", subtitle: $viewModel.username)
+                editableCell("Password", subtitle: $viewModel.password, secure: true)
+            }
+            
+            Section {
+                editableCell("Address", subtitle: $viewModel.address)
+            }
+            
+//            Section {
```
brindy: Needed?

DuckDuckGo/AutofillLoginDetailsView.swift
```diff
@@ -0,0 +1,290 @@
...
+            }
+            
+            Section {
+                CopyableCell(title: "Address", subtitle: viewModel.address, selectedCell: $viewModel.selectedCell) {
+                    viewModel.copyToPasteboard(.address)
+                    viewModel.selectedCell = nil
+                }
+            }
+            
+//            Section {
```
brindy: Needed?

DuckDuckGo/AutofillLoginDetailsView.swift
```diff
@@ -0,0 +1,290 @@
...
+    let subtitle: String
+    @Binding var selectedCell: UUID?
+    let action: () -> Void
+        
+    var body: some View {
+        HStack {
+            VStack(alignment: .leading, spacing: 6) {
+                Text(title)
+                    .label3AltStyle()
+                    .disabled(true)
```
brindy: Why does Text need to be disabled?

DuckDuckGo/AutofillLoginDetailsView.swift
```diff
@@ -0,0 +1,290 @@
...
+                        Spacer()
+                        Image(isPasswordHidden ? "ShowPasswordEye": "HidePasswordEye")
+                            .foregroundColor(.primary)
+                    }
+                    .contentShape(Rectangle())
+                    .frame(width: 50, height: 50)
+                }
+                .buttonStyle(.plain)
+                .background(BackgroundColor(isSelected: selectedCell == id).color)
+                // Prevent taps from being forwarded
```
brindy: Is it the line above that this applies to?

DuckDuckGo/AutofillLoginDetailsViewController.swift
```diff
@@ -0,0 +1,179 @@
...
+        installSubviews()
+        setupNavigationBar()
+        setupCancellables()
+        setupTableViewAppearance()
+        applyTheme(ThemeManager.shared.currentTheme)
+        installConstraints()
+    }
+    
+    deinit {
+        print("DEINIT DETAILS")
```
brindy: 🎏

DuckDuckGo/AutofillLoginDetailsViewController.swift
```diff
@@ -0,0 +1,179 @@
...
+    }
+    
+    override func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
+        authenticate()
+    }
+    
+    private func authenticate() {
+        authenticator.authenticate { error in
+            print("ERROR \(String(describing: error))")
```
brindy: Logging instead?

What happens if an error does occur? Does this function need to do more?

DuckDuckGo/AutofillLoginDetailsViewModel.swift
```diff
@@ -0,0 +1,182 @@
...
+            if let accountID = account.id {
+                let vault = try SecureVaultFactory.default.makeVault(errorReporter: SecureVaultErrorReporter.shared)
+                
+                if let credential = try
+                    vault.websiteCredentialsFor(accountId: accountID) {
+                    self.password = String(data: credential.password, encoding: .utf8) ?? ""
+                }
+            }
+        } catch {
+            print("Can't retrieve password")
```
brindy: More to do here?

DuckDuckGo/AutofillLoginDetailsViewModel.swift
```diff
@@ -0,0 +1,182 @@
...
+                    delegate?.autofillLoginDetailsViewModelDidSave()
+                    
+                    // Refetch after save to get updated properties like "lastUpdated"
+                    if let newCredential = try vault.websiteCredentialsFor(accountId: accountID) {
+                        self.updateData(with: newCredential.account)
+                    }
+                }
+            }
+        } catch {
+
```
brindy: More to do here?

DuckDuckGo/AutofillLoginListAuthenticator.swift
```diff
@@ -0,0 +1,72 @@
...
+        if context.canEvaluatePolicy(.deviceOwnerAuthentication, error: &error) {
+            let reason = reason
+            context.evaluatePolicy(.deviceOwnerAuthentication, localizedReason: reason ) { success, error in
+            
+                DispatchQueue.main.async {
+                    if success {
+                        self.state = .loggedIn
+                        completion(nil)
+                    } else {
+                        print(error?.localizedDescription ?? "Failed to authenticate but error nil")
```
brindy: logging?

DuckDuckGo/AutofillLoginPromptView.swift
```diff
@@ -0,0 +1,214 @@
...
+            VStack {
+            VStack(spacing: 0) {
+                titleHeaderView
+                Spacer()
+                accountButtonsContainer
+                Spacer()
+            }
+            .padding(.top, 43)
+                
+            footer
```
brindy: indentation is confusing here

DuckDuckGo/AutofillLoginPromptView.swift
```diff
@@ -0,0 +1,214 @@
...
+                .foregroundColor(.primary)
+        }
+        .frame(width: Const.Size.closeButtonTappableArea, height: Const.Size.closeButtonTappableArea)
+        .contentShape(Rectangle())
+    }
+    
+    var titleHeaderView: some View {
+        VStack(spacing: 12) {
+            HStack {
+                Image(uiImage: viewModel.faviconImage)
```
brindy: If we have a couple of these favicon images, why not create a `FaviconImageView` which takes the domain/name URL rather than having the view model returning UIImage?

DuckDuckGo/AutofillLoginPromptView.swift
```diff
@@ -0,0 +1,214 @@
...
+            .frame(minWidth: 0, maxWidth: .infinity, maxHeight: Const.Size.CTAButtonMaxHeight)
+            .foregroundColor(.primary)
+            .cornerRadius(Const.Size.CTAButtonCornerRadius)
+    }
+    
+    var accountButtonsContainer: some View {
+        GeometryReader { geometry in
+            ScrollView(showsIndicators: viewModel.shouldUseScrollView) {
+                VStack {
+                accountButtons
```
brindy: identation

DuckDuckGo/AutofillLoginPromptView.swift
```diff
@@ -0,0 +1,214 @@
...
+            
+            Spacer()
+        }
+    }
+    
+    var moreOptionsButton: some View {
+        Button {
+            viewModel.didExpand()
+        } label: {
+            Text(viewModel.moreOptionsButtonString) // TODO email formatting
```
brindy: TODO??

DuckDuckGo/AutofillLoginPromptViewModel.swift
```diff
@@ -0,0 +1,140 @@
...
+            return AutofillInterfaceEmailTruncator.truncateEmail(account.username, maxLength: 36)
+        } else {
+            let urlString = URL(string: account.domain)?.absoluteString.dropPrefix(prefix: "www.")
+            let string = urlString ?? account.domain
+            return "Login for \(string)"
+        }
+    }
+    
+    static func == (lhs: AccountViewModel, rhs: AccountViewModel) -> Bool {
+        return lhs.account.id == rhs.account.id // TODO should make SecureVaultModels.WebsiteAccount actually do this conforming
```
brindy: TODO??

DuckDuckGo/AutofillLoginPromptViewModel.swift
```diff
@@ -0,0 +1,140 @@
...
+    private var accounts: [SecureVaultModels.WebsiteAccount]
+    
+    private var expanded = false {
+        didSet {
+            setUpAccountsViewModels(accounts: accounts)
+        }
+    }
+    
+    var message: String {
+        return "Use Saved Login?" // TODO string
```
brindy: TODO but we talked about strings on MM

DuckDuckGo/AutofillLoginPromptViewModel.swift
```diff
@@ -0,0 +1,140 @@
...
+    }
+    
+    var moreOptionsButtonString: String {
+        return "More Options" // TODO string
```
brindy: TODO but we talked about strings on MM

DuckDuckGo/SaveAutofillLoginManager.swift
```diff
@@ -0,0 +1,133 @@
...
+        self.vaultManager = vaultManager
+        self.autofillScript = autofillScript
+        
+        useStoredCredentialIfNecessary()
+    }
+    
+    // If we have a stored credential with the same password on an empty username account
+    // we want to update it instead of creating a new one
+    private mutating func useStoredCredentialIfNecessary() {
+        if savedMatchingPasswordWithoutUsername != nil {
```
brindy: This is probably safe, but why not use an if let to avoid the force unwrap?

DuckDuckGo/SaveAutofillLoginManager.swift
```diff
@@ -0,0 +1,133 @@
...
+                        if let credentials = credentials {
+                            result.append(credentials)
+                        }
+                    }
+                }
+            }
+            semaphore.signal()
+        }
+        
+        semaphore.wait()
```
brindy: something worries me about this computed property, but I think at least it should have a timeout here and return nothing after a few seconds or else this is going to hang the UI seeing how it is being used.

I think the ideal thing would be to update in the background and the call back then updates models which in turns updates the views... but this seems to be deep a couple of stacks now

DuckDuckGo/SaveLoginView.swift
```diff
@@ -0,0 +1,286 @@
...
+    }
+    
+    var body: some View {
+        GeometryReader { geometry in
+            makeBodyView(geometry)
+        }
+    }
+    
+    private func makeBodyView(_ geometry: GeometryProxy) -> some View {
+        DispatchQueue.main.async { self.frame = geometry.size }
```
brindy, line 77: This does not seem ideal - can we get a comment about why you had to do it like this, please?

DuckDuckGo/SettingsViewController.swift
```diff
@@ -85,7 +85,7 @@ class SettingsViewController: UITableViewController {
     }()
     
     private lazy var shouldShowAutofillCell: Bool = {
-        return false // Hardcoded false until feature is done
+        return featureFlagger.isFeatureOn(.autofill)
```
brindy, line 88: 🥳

DuckDuckGo/SaveLoginViewModel.swift
```diff
@@ -0,0 +1,129 @@
...
+    private var attributedLayoutType: SaveLoginView.LayoutType?
+    
+    internal init(credentialManager: SaveAutofillLoginManagerProtocol, layoutType: SaveLoginView.LayoutType? = nil) {
+        self.credentialManager = credentialManager
+        self.attributedLayoutType = layoutType
+        loadFavicon()
+    }
+    
+    private func loadFavicon() {
+        FaviconsHelper.loadFaviconSync(forDomain: credentialManager.accountDomain,
```
brindy: Yeah, can we try and yeet this into a "FaviconIconView" struct which takes the domain and has its own model, etc?

DuckDuckGo/UIImageViewExtension.swift
```diff
@@ -36,110 +36,18 @@ extension UIImageView {
                      completion: ((UIImage?, Bool) -> Void)? = nil) {
 
         func load() {
-            loadFaviconSync(forDomain: domain, usingCache: cacheType, useFakeFavicon: useFakeFavicon, completion: completion)
+            FaviconsHelper.loadFaviconSync(forDomain: domain,
```
brindy, line 39: Nice one :)

DuckDuckGo/UserText.swift
```diff
@@ -486,6 +486,26 @@ public struct UserText {
...
+    public static let autofillAdditionalLoginInfoMessage = NSLocalizedString("autofill.save-login.additional-login.message", value: "This will save an additional Login for this site.", comment: "Message displayed on modal explaining that an additional login will be saved.")
+    public static let autofillLoginSavedToastMessage = NSLocalizedString("autofill.login-saved.toast", value: "Login saved", comment: "Message displayed after saving an autofill login")
+    public static let autofillLoginUpdatedToastMessage = NSLocalizedString("autofill.login-updated.toast", value: "Login updated", comment: "Message displayed after updating an autofill login")
+    public static let autofillLoginSaveToastActionButton = NSLocalizedString("autofill.login-save-action-button.toast", value: "View", comment: "Button displayed after saving/updating an autofill login that takes the user to the saved login")
+
+    public static let actionAutofillLogins = NSLocalizedString("action.title.autofill.logins", value: "Autofill Logins", comment: "Autofill Logins menu item opening the login list")
+
```
brindy, line 508: Discussed with @THISISDINOSAUR on MM, but if we don't want these translated just now, let's just use String and likewise pull all the other strings from the codebase into here so we're as close as possible to production quality.

brindy:
Not sure what happened to my top level comment with those review comments but here it is ...

1. I have only looked at the Swift code and have left some suggestions / comments.
2. I haven't checked DuckUI against the actual design scheme - I'm going to assume a designer will pick up on any discrepancies. 
3. re Strings - we have some in UserText and some not, and the ones in UserText will get flagged for translation.  Discussed with MM and I think the best approach here is to put all the strings in UserText as regular Strings (ie not NSLocalizedString) and then update them to localised strings when we're ready for public release.
4. Starting manual testing now.

cc @THISISDINOSAUR @Bunn

brindy:
When it works it works, but it doesn't seem reliable at detecting when to prompt for the autofill:

* Clean install
* Launch app
* Visit netflix.com and click sign in
* Enter some bs credentials (e.g. test@example.com and "stupidpassword")
* Save the credentials 
* Fire button
* Visit netflix.com and click sign in

Expected: 

* Offer to use credentials / some UI to show there's credentials / something

Actual:

* Nothing happened until I clicked the fields **a few times**

iOS 15.5 Simulator

Tried with the following sites:

* https://dunoon.scot worked as expected 👍 
* https://www.modiphius.net worked as expected 👍 
* https://katesclothing.co.uk same as Netflix 😢

brindy:
Some UX problems (might be in the list of things to ignore but listing just in case)

1. Can't swipe to delete credentials
2. Can't delete from the edit screen
3. Edit -> Delete on main screen seems wrong
4. No obvious way to close autofill manager (swipe down works, but no 'close' or 'done' button)
5. Close button for initial prompt to save, vs edit after save is different (see below)
   
Screenshot 2022-06-13 at 15 49 24

Screenshot 2022-06-13 at 15 49 06

brindy:
Couple more, also UX

1. Login update <date> text seems off, on macOS we have "Last updated: " and also "Date added: " - I think there should be a good reason if these don't match
2. There's no way to get to the autofill manager without being on a web page

brindy (commented):
See comments above too

DuckDuckGo/AutofillLoginDetailsView.swift
Bunn: Removed

DuckDuckGo/AutofillLoginDetailsView.swift
Bunn: Removed

DuckDuckGo/AutofillLoginDetailsView.swift
Bunn: It was required because without it, the tap recognizer for the MenuItem wasn't being triggered, but this was later fixed some other way and the disable is a left over not needed anymore.
Removed, thanks

DuckDuckGo/SaveLoginView.swift
Bunn: Our [design for this ](https://www.figma.com/file/aM8i2nPES4Edf0EGXKhnwI/Autofill-(iOS)?node-id=0:1) have different margins depending on the phone size (not size class). Without a geometryProxy the frame is not correctly setup and the isSmallFrame property wouldn't work.

I want to remove this before V2, but for V1 seemed like an "ok" workaround.
I've added a comment there. I'll resolve this conversation but feel free to open it again or reach me on MM if you want to discuss it further.

brindy: My question was more about why we were doing the `DispatchQueue.main.async` but that's fine if you plan to improve this later.

Bunn: Oh, without it the runtime complains that we're modifying a state during view update. I've added a comment there as well. But yeah, it should be removed before V2

DuckDuckGo/AutofillLoginPromptView.swift
THISISDINOSAUR: Sorry, old TODO, fixed in the view model

THISISDINOSAUR:
Re the UX stuff:
> Can't swipe to delete credentials

We did have this but Chris didn't like it. Maybe people will mention it in the internal testing so we add it back 😉 


> Can't delete from the edit screen

Yeah, identified as an oversight when I was going through all the things we're supposed to have according to the designs, so we should get it for V2.


> Edit -> Delete on main screen seems wrong

I think like #1 another thing where I think we'll see about feedback during v1 


> No obvious way to close autofill manager (swipe down works, but no 'close' or 'done' button)

This came up in design review, we need some back and forth with Chris about what it should be but decided it was non blocking whilst he was out. We might be able to get it in now he's back.


> Close button for initial prompt to save, vs edit after save is different (see below)

Hmm, not sure what this is supposed to be on the managment interface, I can't find it in the designs @Bunn ?


> Login update text seems off, on macOS we have "Last updated: " and also "Date added: " - I think there should be a good reason if these don't match

There was a lot of discussion about the exact format here: https://app.asana.com/0/38424471409662/1202182448448369/f 
Already more than I think the problem is worth tbh


> There's no way to get to the autofill manager without being on a web page

You should be able to get to it via settings, does it appear in settings for you?

DuckDuckGo/AutofillLoginDetailsViewController.swift
Bunn: The authenticate function updates the callsite wither by closure or my a published value, in this case the published value is being used to handle error/success. I've changed the function to receive an optional closure to better reflect that

DuckDuckGo/AutofillLoginDetailsViewModel.swift
Bunn: I've added a `Pixel.fire(pixel: .secureVaultError)` here CC/ @THISISDINOSAUR

DuckDuckGo/AutofillLoginDetailsViewModel.swift
Bunn: I've added a `Pixel.fire(pixel: .secureVaultError)` here CC/ @THISISDINOSAUR

DuckDuckGo/AutofillLoginDetailsViewModel.swift
THISISDINOSAUR: Do you know what else will happen in these error cases? Do they fail gracefully?

DuckDuckGo/SaveAutofillLoginManager.swift
Bunn: I've added a prepare method on the manager so the data is not prepared in a background queue and the modal only appears after this is done.
As discussed with @THISISDINOSAUR

Bunn:
> > Close button for initial prompt to save, vs edit after save is different (see below)
> 
> Hmm, not sure what this is supposed to be on the managment interface, I can't find it in the designs @Bunn ?
> 

The figma file does not show us what happens after the user taps on "View". I've just added that close button as a suggestion to later be revised by Chris during Design Review, but I did not have a chance to sit down with him and see the options here yet.

DuckDuckGo/AutofillLoginDetailsViewModel.swift
Bunn: @THISISDINOSAUR there was no error handling rather than firing a pixel on V1. Things won't crash, but on this specific case the password on the UI would be empty.

I wonder what could be done here, if securevault doesn't work for some reason the user can't take any action on it, should we display a generic error message? Fail silently (like currently)?

THISISDINOSAUR: I think that's fine for now, I just wanted to check it. doesn't end up locking up the UI or anything like that

THISISDINOSAUR:
Discussed with Brindy on MM and agreed since there's two of us there's a degree to which we can self review, especially since we've now addressed the comments.
So LGTM (although I can't approve since I authored it)

PR #1187, opened 2022-06-10
Additional pixels for bookmark import errors

amddg44:
Task/Issue URL: https://app.asana.com/0/0/1202428268249559/f
Tech Design URL:
CC:

**Description**:
Adding additional pixels to help narrow down the source of the high import error rate 

**Steps to test this PR**:
1. Import an invalid bookmarks html file (there is one [here](https://github.com/duckduckgo/iOS/blob/develop/DuckDuckGoTests/MockFiles/bookmarks_invalid.html)) 
2. Confirm the error pixels `m_bi_e` and `m_bi_e_p` fire
3. Import a valid bookmarks html
4. Confirm the success pixel `m_bi_s` fires

---
###### Internal references:


samsymons (approved):
Works as expected, nice work!

The only worry I have is that we can't differentiate between types of parsing errors, which is where I expect this is most likely to fail.

For instance, `transformSafariDocument` is looking for a `body` tag whereas the core parsing function is looking for a `DL`, and it would be great to know if one of those is more error-prone than the other. Similarly, we can't tell if `transformSafariDocument` is throwing an error or returning nil based on the new pixels alone, since it all gets coalesced into a single pixel.

Core/Pixel.swift
```diff
@@ -168,6 +168,9 @@ public enum PixelName: String {
     
     case bookmarkImportSuccess = "m_bi_s"
     case bookmarkImportFailure = "m_bi_e"
+    case bookmarkImportFailureParsing = "m_bi_e_p"
+    case bookmarkImportFailureSaving = "m_bi_e_s"
+    case bookmarkImportFailureUnknown = "m_bi_e_u"
```
samsymons: One other thing here: [we should move to clear naming conventions where possible](https://app.asana.com/0/69071770703008/1200583483841269/f). Even though it would be inconsistent with the other import pixels, I think it makes sense to go with more verbose names here 🙂

amddg44:
> Works as expected, nice work!
> 
> The only worry I have is that we can't differentiate between types of parsing errors, which is where I expect this is most likely to fail.
> 
> For instance, `transformSafariDocument` is looking for a `body` tag whereas the core parsing function is looking for a `DL`, and it would be great to know if one of those is more error-prone than the other. Similarly, we can't tell if `transformSafariDocument` is throwing an error or returning nil based on the new pixels alone, since it all gets coalesced into a single pixel.

Thanks @samsymons and good point, I've made this change 🙂

Core/Pixel.swift
amddg44: Ah, good to know, thanks! I've made this change too 👍

samsymons (approved):
LGTM 👏

PR #1199, opened 2022-06-16
Add Back/Forward navigation long press menu

Bunn:
Task/Issue URL: https://app.asana.com/0/414235014887631/1202297990501898/f

**Description**:
This PR adds a feature for users to long press the back/forward buttons on the navigation bar and be presented with a menu of options that the user can go directly to.

**Steps to test this PR**:
1. Visit a bunch of different websites
2. Long press the back button and check if the sites you visited are there
3. Select a website in the middle, check if the site loaded correctly
4. Long press the forward button and check if the stack displays the websites you visited after the current one.
____
1. Check if the order of the websites on the back button are newest at the bottom, oldest at the top
2. Check if the order of the websites on the forward button are newest at the top oldest at the bottom
----- 
1. Check if protections are being respected after reloading a website from the back button
----- 
1. This feature is iOS 14+ only, check if this is being respected.
----- 

**Device Testing**:

* [x] iPhone
* [x] iPad

**OS Testing**:

* [x] iOS 13
* [x] iOS 14
* [x] iOS 15

**Theme Testing**:

* [x] Light theme
* [x] Dark theme


DuckDuckGo/MainViewController.swift
```diff
@@ -117,6 +117,7 @@ class MainViewController: UIViewController {
     
     // Skip SERP flow (focusing on autocomplete logic) and prepare for new navigation when selecting search bar
     private var skipSERPFlow = true
+
```
samsymons, line 120: Extra whitespace to be removed here - surprised SwiftLint didn't pick this up actually

DuckDuckGo/BackForwardMenuHistoryItem.swift
```diff
@@ -0,0 +1,62 @@
...
+    }
+    
+    var sanitizedURLForDisplay: String {
+        BackForwardMenuHistoryItemURLSanitizer.sanitizedURLForDisplay(backForwardItem.initialURL)
+    }
+}
+
+struct BackForwardMenuHistoryItemURLSanitizer {
+    
+    static func sanitizedURLForDisplay(_ url: URL) -> String {
```
samsymons, line 40: One small note here, this function isn't Punycode decoding the URL string - but we also don't do that for URLs properly anyway, so it's not a blocker. 🙂

DuckDuckGo/MainViewController.swift
```diff
@@ -1774,4 +1788,49 @@ extension MainViewController: VoiceSearchViewControllerDelegate {
...
+    private func refreshBackForwardMenuItems() {
+        guard let currentTab = currentTab, #available(iOS 14.0, *) else {
+            return
+        }
+        omniBar.backButton.menu = historyMenu(with: currentTab.webView.backForwardList.backList.reversed())
+        backButton.menu = historyMenu(with: currentTab.webView.backForwardList.backList.reversed())
+        
+        omniBar.forwardButton.menu = historyMenu(with: currentTab.webView.backForwardList.forwardList)
+        forwardButton.menu = historyMenu(with: currentTab.webView.backForwardList.forwardList)
```
samsymons: Are we able to build each of these menus once and share them? Something like:

```suggestion
        let backMenu = historyMenu(with: currentTab.webView.backForwardList.backList.reversed())
        omniBar.backButton.menu = backMenu
        backButton.menu = backMenu
        
        let forwardMenu = historyMenu(with: currentTab.webView.backForwardList.forwardList)
        omniBar.forwardButton.menu = forwardMenu
        forwardButton.menu = forwardMenu
```

DuckDuckGo/MainViewController.swift
```diff
@@ -1774,4 +1788,49 @@ extension MainViewController: VoiceSearchViewControllerDelegate {
...
+                }
+            } else {
+                return  UIAction(title: historyItem.title,
+                                 discoverabilityTitle: historyItem.sanitizedURLForDisplay) { [weak self] _ in
+                    self?.loadBackForwardItem(historyItem.backForwardItem)
+                }
+            }
+        }
+        
+        let maxNumberOfItems = 6
```
samsymons: Since we're capping the maximum number of items, what do you think about doing this truncation at the start of the function before we map items to `UIAction`? We can avoid unnecessary allocations that way 🙂

DuckDuckGo/MainViewController.swift
```diff
@@ -1774,4 +1788,49 @@ extension MainViewController: VoiceSearchViewControllerDelegate {
...
+        return Array(menuItems.suffix(maxNumberOfItems))
```
samsymons: One issue here: in the case of the back button, the list is reversed, so taking the suffix means that after a certain point the list won't be accurate.

I just tested this with a back list history of 9 items, and I see 6 in the menu as expected, but it's starting from the oldest item in the list - the last few pages I visited are being truncated.

samsymons (changes requested):

DuckDuckGo/BackForwardMenuHistoryItem.swift
Bunn: I've added a call to the punycodedUrl extension. Should be better now :) Thanks

DuckDuckGo/MainViewController.swift
Bunn: Oh, much better. This happened because of a "I forgot about the iPad" moment, and didn't look it back to make it better. Thank you :)

DuckDuckGo/MainViewController.swift
Bunn: This limitation should not be there. Safari doesn't have it, it just adds a scroll to the Menu. Good catch, thanks

DuckDuckGo/MainViewController.swift
Bunn: Same as above, I'll remove the 6 limitation. Thanks :)

samsymons (approved):
Looks good, great work @Bunn! 👏 

The only things that came to mind at the end of the review were:

1. **State restoration:** If you close the browser, the back/forward list is gone. The problem is that this list is immutable so we'd have to do something custom to handle this correctly, since a lot of the state restoration functionality in `WKWebView` is private. I don't think this should block the initial implementation from going in, so don't worry about this here.
2. **Efficiency:** I worry that re-building the menu on every navigation event is inefficient, since I expect users to only bring it up occasionally during a session. Creating and showing it lazily would be a nice improvement, as long as it's fast to create and display it lazily if you have a large set of back-forward items - but again I don't view this as a blocker.
3. **Instrumentation:** I considered suggesting we add instrumentation to get a read on how frequently this menu is used, but am now against that idea since there's not a lot to be learned with that knowledge. Let's leave pixels out. 🙂

Bunn:
Thank you @samsymons  :) 

> 1. **State restoration:** If you close the browser, the back/forward list is gone. The problem is that this list is immutable so we'd have to do something custom to handle this correctly, since a lot of the state restoration functionality in `WKWebView` is private. I don't think this should block the initial implementation from going in, so don't worry about this here.

Yep, since we currently don't have state restoration for the back/forward menus as it is today I consider this to be something outside of the scope for this feature.


> 2. **Efficiency:** I worry that re-building the menu on every navigation event is inefficient, since I expect users to only bring it up occasionally during a session. Creating and showing it lazily would be a nice improvement, as long as it's fast to create and display it lazily if you have a large set of back-forward items - but again I don't view this as a blocker.

Good point. I did check performance during implementation and didn't find anything that made me worry. The main reason it's not lazy loaded is because I didn't find a good way to do it without fighting UIKit. The menu is a property of UIBarButtonItem, which means we don't actually have a callback for when it's called. One way to fix this would be to add a tap recognizer in both buttons, create the menus in a lazy way and then manually display the UIMenu using the button frame, but it feels convoluted. Am I overthinking this? If you think it's a good approach or if you have a different solution I'm not considering please let me know :) 

> 3. **Instrumentation:** I considered suggesting we add instrumentation to get a read on how frequently this menu is used, but am now against that idea since there's not a lot to be learned with that knowledge. Let's leave pixels out. 🙂

Agreed. If necessary we can add it later.


https://github.com/duckduckgo/iOS/pull/1199#pullrequestreview-1012737379

Issue #1200, opened 2022-06-17
TrackerRadarKit crash when deserialising JSON


miasma13:
**Task/Issue URL:** https://app.asana.com/0/414709148257752/1202407381883565/f

On older devices the JSON deserialisation when initialising TrackerRadarKit objects. Reason of the crash is that this code is done in scene initialisation and with slower performance the device is not doing it under assigned time and the app is killed by watchdog.
The source of the crash is **ContentBlockerRulesManager** initialisation (triggering JSON load) at scene creation time triggered from **MainViewController's** method: **prepare(for:sender:)**

**Proposal**
One of the approaches at fixing this one of high ranking crashes would be to: 
- identify code paths triggering the ContentBlockerRulesManager initialisation from MainViewController's method: prepare(for:sender:)
- decouple the initialisation from the scene loading

bwaresiak:
Are we sure this is the only place where the problem lies? There are more places in the iOS scene initialization flow that reach to TDS, like fallback in `DefaultContentBlockerUserScriptConfig:54` or `DefaultSurrogatesUserScriptConfig:60` to load `trackerDataManager?.trackerData` in case nothing is provided - and it is invoked by `viewDidLoad` logic on Tab initialization:

<img width="430" alt="init_flow" src="https://user-images.githubusercontent.com/2865434/174300590-836a54a6-8e96-48cf-a4ea-7d0370bab326.png">

As it is called in `MainViewController.viewDidLoad()` flow, could that also cause above issue?

mallexxx:
@bwaresiak valid concern, what do you think about refactoring the CBR/UserScripts loading to Publishers sequence and waiting for CBRs to finish compiling in decidePolicyForNavigationAction?

bwaresiak:
Yes - but we need to wait for @jaceklyp's code to land (days I guess?) as he is working on first loading improvements.

mallexxx:
I made an initial effort refactoring the ContentBlockingRules to async waiting and loading as a single asset;
@bwaresiak when you have time please check and we‘ll have a chat to validate the approach

samsymons:
@mallexxx Closing this out as the original PRs have been merged, and the Asana task is complete. Of course, feel free to re-open if there's more work to do that I'm not aware of.

PR #1201, opened 2022-06-17
Load TrackerData lazily

mallexxx:
Task/Issue URL: https://app.asana.com/0/414709148257752/1202407381883565/f
GH Issue URL: https://github.com/duckduckgo/iOS/issues/1200
BSK PR: https://github.com/duckduckgo/BrowserServicesKit/pull/110
macOS PR: https://github.com/more-duckduckgo-org/macos-browser/pull/796
Tech Design URL:
CC: @miasma13 @bwaresiak 

**Description**:
Load TrackerData lazily and asynchronously to avoid JSON Decoding during init leading to app termination at startup

**Steps to test this PR**:
1. Validate Content Blocking Rules load and work without issues

---
###### Internal references:


bwaresiak:
@mallexxx If you can make this up to date, I'll find time to look at it.

DuckDuckGo/PrivacyProtectionController.swift
```diff
@@ -66,6 +65,7 @@ class PrivacyProtectionController: ThemableNavigationController {
             PrivacyProtectionErrorController(coder: coder, configuration: self.privacyConfig)
         }) else { return }
         controller.errorText = errorText
+        controller.delegate = self
```
mallexxx, line 68: Seems like the delegate setting when a page displays error was missing here, was it intentional?

mallexxx:
TODO: make it explicit in DEBUG

DuckDuckGo/TabViewController.swift
```diff
@@ -317,96 +288,24 @@ class TabViewController: UIViewController {
...
-            textSizeUserScript,
-            findInPageScript,
-            navigatorPatchScript,
-            surrogatesScript,
-            contentBlockerRulesScript,
-            fingerprintScript,
-            faviconScript,
-            fullScreenVideoScript,
-            autofillUserScript,
-            printingUserScript
```
mallexxx, line 412: why was documentScript (DocumentUserScript) missing here? was it intentional, should I remove it from UserScripts.swift?

bwaresiak:
I did some tests and it seems that attribution is not handled correctly - state is inconsistent. 

Scenario:
1. Enter attribution flow for some domain.
2. On vendor site see that attribution works.
3. Trigger content rule list re-application by toggling protection on and off.
4. Error - attribution is not active on the vendor site even tho it should be.

DuckDuckGo/TabViewController.swift
bwaresiak: Looking back on the blame, I think this is iOS 12 leftover and could be removed, probably? :)

mallexxx:
@bwaresiak I updated the Attribution logic to match the latest macOS implementation, now this scenario seems working as it should

DuckDuckGo/TabViewController.swift
mallexxx: ok, made extra clean-up of document.js and LongPressGestureRecognizer

bwaresiak:
Overall changes to architecture look great, but there is one thing that worries me with regard to original objective:

Plan was to move loading of the data to background / non-main thread or at least decouple it from app startup flow and scene initialization. Now, since ContentBlocking has became a singleton (instead of class with static fields initialized on demand) first access to that loads all the data (config, TDS) which is in fact in app delegate didFinishLaunchingWithOptions (through `PrivacyFeatures.httpsUpgrade.loadDataAsync()`). So this has been moved from Scene Initialization to App Initialization flow - which is not necessarily a bad thing as we should have a bit more resources available there (I think watchdog timeout is larger there) but at the same time this may be something we may want to improve further.

Note that I don't expect this to negatively impact startup performance - after all we have been already do this exact work, just from the MainVC - so we just moved this to earlier point of time.

I'm going to retest content blocking stack Tomorrow and if everything will be fine, let's go with it, unless @miasma13 or @mallexxx  has something to add to my comment. :)

mallexxx:
I‘ve performed launch measurements and the ContentBlocking initialization at init seems totally irrelevant in launch time, StorageCache initialization has much greater effect (caused mainly by TLD init because of ineffective JSONDecoder memory usage) although not really bad
Screenshot 2022-11-03 at 13 26 28

DuckDuckGo/ContentBlockingUpdating.swift
```diff
@@ -0,0 +1,99 @@
...
+        // 1. Collect updates from ContentBlockerRulesManager and generate UserScripts based on its output
+        cancellable = contentBlockerRulesManager.updatesPublisher
+            // regenerate UserScripts on:
+            // prefs changes notifications with initially published value for combineLatest to work
+            .combineLatest(onNotificationWithInitial(PreserveLogins.Notifications.loginDetectionStateChanged), combine)
+            .combineLatest(onNotificationWithInitial(AppUserDefaults.Notifications.doNotSellStatusChange), combine)
+            .combineLatest(onNotificationWithInitial(AppUserDefaults.Notifications.autofillEnabledChange), combine)
+            .combineLatest(onNotificationWithInitial(AppUserDefaults.Notifications.textSizeChange), combine)
+            .combineLatest(onNotificationWithInitial(AppUserDefaults.Notifications.didVerifyInternalUser), combine)
+            .combineLatest(onNotificationWithInitial(StorageCacheProvider.didUpdateStorageCacheNotification)
```
bwaresiak, line 88: Wasn't that removed from list of notifications that trigger reload?

bwaresiak (approved):
Great job, one comment in line, one more here, but good to go when addressed:

When I create new tab, reload is being called through combine (TabVC:2124) for that new tab - is that needed?

diegoreymendez:
I just restored this branch to investigate a finding page regression.

Issue #1202, opened 2022-06-17
Printing on specific site does not work


miasma13:
Task/Issue URL: https://app.asana.com/0/414709148257752/1202193006784955/f

**Steps to reproduce:**
1. Navigate to https://www.theenglishkitchen.co/2009/09/beer-battered-fish.html
2. Scroll down and tap on the "Print" button
3. A new blank page with "about:blank" URL tab is opened and nothing happens

PR #1205, opened 2022-06-20
Fix popup tab loading interruption

mallexxx:
Task/Issue URL: https://app.asana.com/0/414709148257752/1202193006784955/f
GH Issue URL: https://github.com/duckduckgo/iOS/issues/1202
Tech Design URL:
CC: @miasma13 

**Description**:
- Fixes loading interruption when WebView created in webView:createWebViewWithConfiguration:..
- Loading is interrupted for cleaned urls but managed by a creator tab otherwise
- This fixes an issue of Printing not working on a new tab created for printing

**Steps to test this PR**:
1. Validate print works on https://www.theenglishkitchen.co/2009/09/beer-battered-fish.html?m=1
2. Validate links opened in new tab work (e.g. in yandex search)
3. Validate pop-up windows created with js work

---
###### Internal references:


DuckDuckGo/TabViewController.swift
```diff
@@ -1630,6 +1637,7 @@ extension TabViewController: WKUIDelegate {
             completionHandler(nil)
         }
     }
+
```
miasma13: Unnecessary edit.

DuckDuckGo/TabViewController.swift
```diff
@@ -349,14 +352,18 @@ class TabViewController: UIViewController {
...
-                                           onFinishExtracting: { },
-                                           completion: { [weak self] cleanURL in self?.load(urlRequest: .userInitiated(cleanURL)) })
-            } else {
-                load(urlRequest: request)
+        } else if let url = request?.url {
+            var loadingStopped = false
+            linkProtection.getCleanURL(from: url) { [weak self] in
+                if loadedByParentTab {
+                    loadingStopped = true
+                    self?.webView.stopLoading()
```
miasma13: Can you please provide a bit explanation for this logic?

DuckDuckGo/TabViewController.swift
```diff
@@ -349,14 +352,18 @@ class TabViewController: UIViewController {
...
+                }
+                showProgressIndicator()
+            } onFinishExtracting: {
+            } completion: { [weak self] cleanURL in
+                guard url != cleanURL || !loadedByParentTab || loadingStopped else { return }
```
miasma13: What is purpose for this guard?
Does it even makes sense, as it seems that both `loadedByParentTab` and `loadingStopped` will always have same values.

miasma13 (changes requested):

DuckDuckGo/TabViewController.swift
mallexxx: the check here passes if either loading has stopped or if the tab is NOT loaded by its parent, done so not to interrupt an ongoing popup request loading if possible

mallexxx:
Thanks, @miasma13, I've changed naming and added some comments for clarity
TLDR: the popup window based printing breaks because we‘ve been stopping the popup loading (initiated by its parent js) when URL cleanup happened. This change tries to avoid loading breakage by stopping loading only for non-parent-initiated requests or if the URL was modified. Otherwise (for a popup) we will just keep current request going

miasma13 (approved):
Thanks for the explanation

PR #1207, opened 2022-06-20
Block trackers on start

jaceklyp:
Task/Issue URL: https://app.asana.com/0/72649045549333/1201955257558990/f
Tech Design URL: https://app.asana.com/0/0/1201985695010211/f
CC:

**Description**:

Make sure that trackers are always blocked on start.

**Steps to test this PR**:
1. Compile rules and cache these.
1. Kill the app.
1. Update Privacy config or TDS (along with ETag).
1. See that:
rules are returned quickly (no delay on startup)
trackers are blocked on startup
rules are recompiled to latest correct state within few seconds from browser start

**Device Testing**:

* [ ] iPhone X

**OS Testing**:

* [ ] iOS 15


bwaresiak:
Noticed weird bug:
1. Navigate to SERP.
2. Open SERP settings, enable open in new tab option.
3. Click on any result - two tabs are open instead of one.

brindy (approved):
Xcode is complaining about a couple of swiftlint errors, but otherwise LGTM!

PR #1208, opened 2022-06-20
Drop iOS 13

Bunn:
Task/Issue URL: https://app.asana.com/0/414709148257752/1202397540422086/f

**Description**:
Drop support for iOS 13

**Steps to test this PR**:
General smoke tests to check if everything is OK.

Special attention to these areas:
* Onboarding
* Request review on TabViewController
* Import bookmarks
* Settings screen 

**Device Testing**:

* [x] iPhone 
* [x] iPad

**OS Testing**:

* [x] iOS 13
* [x] iOS 14
* [x] iOS 15


amddg44:
Hi @Bunn, I've taken an initial pass and it's looking really good 🙂  

Just two small comments:
- There are a number of `iOS 13.4` availability checks that could also be removed as part of this PR
- I believe `OnboardingHomeRowViewController.swift` can be deleted since it's been removed from Onboarding.storyboard

Bunn:
@amddg44 good catch, thank you :) 
I've done what you suggested. It would be good to test the two options on onboarding now: Skip and Set as default browser.

I've done it in both iPad and iPhone and it looks good, but a second pair of eyes would be nice.

amddg44:
hi @Bunn, i've done a lot more testing over the weekend and I discovered an issue with bookmark imports. I did a little digging and it seems it is now failing out in the guard let on trying to access the contents of the file in the UIDocumentPickerDelegate [callback] (https://github.com/duckduckgo/iOS/blob/70e33748cf0556605dbb7c6196401acbe81db435/DuckDuckGo/BookmarksViewController.swift#L652)

Otherwise I've thoroughly tested onboarding etc and everything is looking great! 👍 🙂

Bunn:
@amddg44 Good catch, the issue doesn't seem to happen on the simulator because it was a permission thing.
Setting the flag to copy the resource on the picker init makes it work like the previous .import parameter.

Simulator is not the best place to test sandbox things 😅 
It should be working now, tested on both iOS 14 and 15 (on real devices)

amddg44 (approved):
Nice job @Bunn! LGTM 😄

PR #1212, opened 2022-06-22
Make secure vault errors more specific

THISISDINOSAUR:
Task/Issue URL: https://app.asana.com/0/414235014887631/1202489927113794/f
Tech Design URL:
CC:

**Description**:
We have errors coming in, hard to know why, might just be brindy fucking about :D
But would like to know for sure, so this adds more info

**Steps to test this PR**:
This is hard to test since I can't force the errors, and ended up just making sure nothing is wrong with the pixels like this:
Screenshot 2022-06-22 at 18 22 40
Here's some code you can copy and paste:
```
let error = NSError(domain: "domain", code:200, userInfo:nil)
SecureVaultErrorReporter.shared.secureVaultInitFailed(SecureVaultError.failedToOpenDatabase(cause: error))
SecureVaultErrorReporter.shared.secureVaultInitFailed(SecureVaultError.initFailed(cause: error))
SecureVaultErrorReporter.shared.secureVaultInitFailed(SecureVaultError.generalCryptoError)
```
Use something like Charles proxy to check the pixels fire with the error code and domain in there as params

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPad

**OS Testing**:

* [ ] iOS 13
* [ ] iOS 14
* [ ] iOS 15

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:

### DIFF
--- a/DuckDuckGo/SecureVaultErrorReporter.swift
+++ b/DuckDuckGo/SecureVaultErrorReporter.swift
@@ -35,7 +35,7 @@ final class SecureVaultErrorReporter: SecureVaultErrorReporting {
         case .failedToOpenDatabase(let error):
             Pixel.fire(pixel: .secureVaultFailedToOpenDatabaseError, error: error)
         default:
-            Pixel.fire(pixel: .secureVaultError)
+            Pixel.fire(pixel: .secureVaultError, includedParameters: [])
 
         }
     }
--- a/DuckDuckGo/SecureVaultErrorReporter.swift
+++ b/DuckDuckGo/SecureVaultErrorReporter.swift
@@ -30,8 +30,10 @@ final class SecureVaultErrorReporter: SecureVaultErrorReporting {
         guard !ProcessInfo().arguments.contains("testing") else { return }
 #endif
         switch error {
-        case .initFailed, .failedToOpenDatabase:
-            Pixel.fire(pixel: .secureVaultInitError)
+        case .initFailed(let error):
+            Pixel.fire(pixel: .secureVaultInitFailedError, error: error)
+        case .failedToOpenDatabase(let error):
+            Pixel.fire(pixel: .secureVaultFailedToOpenDatabaseError, error: error)
         default:
             Pixel.fire(pixel: .secureVaultError)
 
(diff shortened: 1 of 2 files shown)


DuckDuckGo/SecureVaultErrorReporter.swift
```diff
@@ -30,8 +30,10 @@ final class SecureVaultErrorReporter: SecureVaultErrorReporting {
         guard !ProcessInfo().arguments.contains("testing") else { return }
 #endif
         switch error {
-        case .initFailed, .failedToOpenDatabase:
-            Pixel.fire(pixel: .secureVaultInitError)
+        case .initFailed(let error):
+            Pixel.fire(pixel: .secureVaultInitFailedError, error: error)
+        case .failedToOpenDatabase(let error):
+            Pixel.fire(pixel: .secureVaultFailedToOpenDatabaseError, error: error)
```
samsymons, line 36: One issue here: the `Pixel.fire(pixel:)` and `Pixel.fire(pixel:, error:)` calls have different default behaviour when it comes to the additional parameters they send. The call that includes an error doesn't include ATB by default, but the version without the error _does_.

I don't think we should include ATB with these pixels, unless you have a counterargument for how it would be used. If not, let's change the call below this comment to leave out the additional parameters.

DuckDuckGo/SecureVaultErrorReporter.swift
```diff
@@ -30,8 +30,10 @@ final class SecureVaultErrorReporter: SecureVaultErrorReporting {
...
         default:
             Pixel.fire(pixel: .secureVaultError)
```
samsymons:
```suggestion
            Pixel.fire(pixel: .secureVaultError, includedParameters: [])
```

samsymons (changes requested):

DuckDuckGo/SecureVaultErrorReporter.swift
THISISDINOSAUR: That's a good point, I didn't notice that, will change

THISISDINOSAUR: Tbh, we should probably change what this does by default and by default not send extra info

samsymons (approved):
Looks good!

Issue #1203, opened 2022-06-17
Sharing BLOB URL file does not work


miasma13:
BLOB URLs work when opened directly in the tab. But downloading from such URL is not possible hence sharing a such file does not work (since first it is download to a temp location and then shared to other app).

PR #1213, opened 2022-06-23
Add blob:// download support

mallexxx:
Task/Issue URL: https://github.com/duckduckgo/iOS/issues/1203
Tech Design URL:
CC: @miasma13 

**Description**:
- Adds blob:// download support
- #available(iOS 14) to be dropped after dropping iOS 13 support

**Steps to test this PR**:
1. Validate blob:// preview and sharing works on previewable (e.g. pdf) files (use mega.nz for example)
2. Validate blob:// downloads works for larger files ~15-20MB of unknown type
3. Validate regular downloads work

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 14.5


DuckDuckGoTests/ExternalUrlSchemeTests.swift
```diff
@@ -60,7 +60,7 @@ class ExternalUrlSchemeTests: XCTestCase {
     
     func testThatAboutSchemesAreAllowed() {
         let url = URL(string: "about:blank")!
-        XCTAssertNotEqual(SchemeHandler.schemeType(for: url),
-                          SchemeHandler.SchemeType.navigational)
+        XCTAssertEqual(SchemeHandler.schemeType(for: url),
+                       SchemeHandler.SchemeType.navigational)
```
mallexxx, line 64: a bug?

Bunn:
@mallexxx I didn't finish looking at the code but I found 2 small issues that maybe you can take a look
1 - Suggested filename is not working, I haven't investigated why, but I confirmed the macOS version seems to work, so I believe there is probably something we can do.
Screen Shot 2022-06-28 at 19 34 32
Screen Shot 2022-06-28 at 19 46 19

2 - After the blob download, if you cancel the save to files, then tap on download again, the filename suffix is incremented by one (see gif) I don't think this is major, but it may be an indicator of some bigger problem.
Simulator Screen Recording - iPhone 11 - 2022-06-28 at 19 39 58

cc/ @miasma13

mallexxx:
Thank you @Bunn ! Fixed the issues, should be ready for review now

DuckDuckGo/TabViewController.swift
```diff
@@ -1650,12 +1546,324 @@ extension TabViewController: WKNavigationDelegate {
...
+              !downloadMetaData.mimeType.isHTML
+        else {
+            temporaryDownloadForPreviewedFile?.cancel()
+            temporaryDownloadForPreviewedFile = nil
+            completion(.allow)
+            return
+        }
+        guard SchemeHandler.schemeType(for: url) != .blob else {
+            // suggestedFilename is empty for blob: downloads unless handled via completion(.download)
+            // WKNavigationResponse._downloadAttribute private API could be used instead of it :(
```
Bunn, line 1620: @mallexxx @miasma13 do you think we should file a ~~radar~~ feedback with Apple?

DuckDuckGoTests/ExternalUrlSchemeTests.swift
Bunn: @mallexxx I think so, [this Equatable check](https://github.com/duckduckgo/iOS/blob/24735a3315ddfabd45c91b203b50fd4c0a0017ce/Core/SchemeHandler.swift#L90-L101) seems wrong to me. Thanks for the fix :) 

What scares me is that this test was built specifically expecting the .navigational == .navigation comparison to be false. I wonder why.

Bunn (approved):
✅ I've tested binary, blob and special cases like wallet items, everything it working well.
✅ Fire button still takes the download into consideration
✅ Didn't notice any leaks related to it when running on instruments

✅LGTM
Thanks @mallexxx

DuckDuckGo/TabViewController.swift
mallexxx: I don‘t think so, I'd rather switch to WKDownload usage instead of custom downloads implementation when we drop iOS <14.5

PR #1257, opened 2022-07-28
Add assert and pixel for catching fail scenario when downloaded file is not found

miasma13:
Task/Issue URL: https://app.asana.com/0/414235014887631/1202675417049719/f
Tech Design URL:
CC:

**Description**:
Add assert and pixel for catching fail scenario when downloaded file is not found

### DIFF
--- a/Core/PixelEvent.swift
+++ b/Core/PixelEvent.swift
@@ -292,6 +292,8 @@ extension Pixel {
         case cachedTabPreviewsExceedsTabCount
         case cachedTabPreviewRemovalError
         
+        case missingDownloadedFile
+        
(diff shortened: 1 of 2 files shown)


Core/PixelEvent.swift
```diff
@@ -573,6 +575,8 @@ extension Pixel.Event {
         case .cachedTabPreviewsExceedsTabCount: return "m_d_tpetc"
         case .cachedTabPreviewRemovalError: return "m_d_tpre"
             
+        case .missingDownloadedFile: return "m_missing_downloaded_file"
```
bwaresiak: For debug pixels we are typically using "m_d_" as a prefix.

bwaresiak (approved):
LGTM

PR #1259, opened 2022-07-29
Fix for decision handler in decidePolicyForNavigationResponse not being always called

miasma13:
Task/Issue URL: https://app.asana.com/0/414709148257752/1202688062019256/f
Tech Design URL:
CC:

**Description**:
With release of 7.68.6 a crash was introduced related to download related refactoring. The reason for the crash is decision handler from the webView:decidePolicyForNavigationResponse:decisionHandler: method not being called in all logic flows which is mandated by WebKit.

### DIFF
--- a/Core/PixelEvent.swift
+++ b/Core/PixelEvent.swift
@@ -293,6 +293,7 @@ extension Pixel {
         case cachedTabPreviewRemovalError
         
         case missingDownloadedFile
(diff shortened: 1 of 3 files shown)


Bunn (approved):
LGTM.
Binary, PDF, images and wallet items correctly downloaded

PR #1269, opened 2022-08-11
Add delegated autofill isEnabled check and pixel params

THISISDINOSAUR:
Task/Issue URL:
Tech Design URL:
CC:

**Description**:
This was originally merged along with the app group change, but the app group change had an issue so was reverted. This just adds back the none problematic things:

- Adds two additional pixels to see if the check for secure vault being available is being spammed
- Adds param to see if the app is backgrounded, this will help us get more certainty on the secure vault errors
- Adds a param if the user is internal, so in future if an error pixel is only firing for internal users we can calm down

**Steps to test this PR**:
1. The most important thing to test is that autofill still works. See here for autofill testing steps: https://github.com/duckduckgo/iOS/pull/1184 . Also test that the app doesn't crash when in the background for >10 seconds, since that's the issue the original PR introduced
2. Check that all pixels have the is internal user flag. 
3. In FeatureFlagger, hard code internal user to false, pixels should not have the param at all (not even set to false) 
4. When you load a page check that "m_secure-vault_is-enabled-checked_when-enabled" and "m_secure-vault_is-enabled-checked_when-disabled" are being sent appropriately (hard code it off in FeatureFlagger to check the latter). Check they also have the isBackgrounded flag
5. See https://github.com/duckduckgo/iOS/pull/1212 to test secure vault error pixels, check they have the isBackgrounded param

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPad

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:


Bunn (approved):
✅ Autofill is working (tested saving/editing/deleting/using entries)
✅ No crashes on background
✅ Pixels have internal flag
✅ SecureVault pixels have the isBackgrounded flag and secureVault pixels are being sent
⚠️ It's not from this PR but likely leftover when it was reverted so maybe a good moment to take a look?:

Do we need to set the flag here?
https://github.com/duckduckgo/iOS/blob/c9875bf09ac054c8118660eaaa5fe5fb27d0f051/Core/Pixel.swift#L179-L181

Since this function calls another that also sets the flag
https://github.com/duckduckgo/iOS/blob/c9875bf09ac054c8118660eaaa5fe5fb27d0f051/Core/Pixel.swift#L138-L140

THISISDINOSAUR:
> ✅ Autofill is working (tested saving/editing/deleting/using entries) ✅ No crashes on background ✅ Pixels have internal flag ✅ SecureVault pixels have the isBackgrounded flag and secureVault pixels are being sent ⚠️ It's not from this PR but likely leftover when it was reverted so maybe a good moment to take a look?:
> 
> Do we need to set the flag here?
> 
> https://github.com/duckduckgo/iOS/blob/c9875bf09ac054c8118660eaaa5fe5fb27d0f051/Core/Pixel.swift#L179-L181
> 
> Since this function calls another that also sets the flag
> 
> https://github.com/duckduckgo/iOS/blob/c9875bf09ac054c8118660eaaa5fe5fb27d0f051/Core/Pixel.swift#L138-L140

Good spot, have fixed.

PR #1270, opened 2022-08-11
Revert PR#1214 to fix ActionMessageView crash when adding constraint

miasma13:
Task/Issue URL: https://app.asana.com/0/414709148257752/1202783782685842/f
Tech Design URL:
CC:

**Description**:
Earlier change introduced a timing issue prone to causing crash during adding constraint to a view controller's view that was being removed from the hierarchy.


bwaresiak (approved):
LGTM

PR #1276, opened 2022-08-16
Try to fix Bookmarks CoreData structure 

Bunn:
Task/Issue URL: https://app.asana.com/0/414709148257752/1202779945035904/f

**Description**:

This change tries to restore the bookmarks core data structure when it is not conforming to pre-defined data, like having a single top level folder.

**Steps to test this PR**:
I wasn't able to reproduce the situation that leads to a crash by just using the app so I recommend manually changing the database to force the faulty scenario (either by opening the database manually or with some app like [Browser for SQLite](https://sqlitebrowser.org/)

To find where the sqlite file is, place a breakpoint and inspect the storeURL here:
https://github.com/duckduckgo/iOS/blob/6c2c3cc7aa23b86fb64bdbb559ac58b55371db33/Core/BookmarksCoreDataStorage.swift#L80-L83

***No Orphaned Folders - Fresh install***
1. Open the app
2. Close the app
3. Open the database and remove both top level folders (favorite and bookmark)
4. Open the app and check if the folders were created correctly

***No Orphaned Folders + Data - Fresh install***
1. Open the app
2. Add some favorites and bookmarks by using the app
3. Close the app
4. Open the database and remove both top level folders (favorite and bookmark)
5. Open the app and check if the folders were created correctly and bookmarks/favorites were not lost

***Multiple Orphaned Folders - Fresh install***
1. Open the app
2. Close the app
3. Open the database and add other root level folders 
4 Open the app and check if the extra folders were removed correctly

***Multiple Orphaned Folders + Data - Fresh install***
1. Open the app
2. Add some favorites and bookmarks by using the app
3. Close the app
3. Open the database and add other root level folders 
4. Manually move some random favorites to other root folders created this way by changing their parent column ID
4 Open the app and check if the extra folders were removed correctly and bookmarks/favorites were not list

***Run the tests***
1. Run unit tests which should simulate conditions with multiple or no orphaned folders

**Check Pixels**
1. Check if the new pixels are being correctly sent

**OS Testing**:

* [x] iOS 14
* [x] iOS 15

---
###### Internal references:


THISISDINOSAUR:
I don't love this band aid anyway, but I really don't want us to be doing it twice. Do we have any pixel data on what the actual issue is? If so, I'd rather only cover that case (e.g. get rid of createMissingTopLevelFolder if the problem is only ever there being multiple)

Core/PixelEvent.swift
```diff
@@ -618,11 +621,14 @@ extension Pixel.Event {
...
         
-        case .debugBookmarkOrphanFolder: return "m_d_bookmark_orphan_folder"
-        case .debugBookmarkTopLevelMissing: return "m_d_bookmark_top_level_missing"
+        case .debugBookmarkOrphanFolder2: return "m_d_bookmark_orphan_folder2"
+        case .debugBookmarkTopLevelMissing2: return "m_d_bookmark_top_level_missing2"
         
-        case .debugFavoriteOrphanFolder: return "m_d_favorite_orphan_folder"
-        case .debugFavoriteTopLevelMissing: return "m_d_bookmark_top_level_missing"
+        case .debugFavoriteOrphanFolder2: return "m_d_favorite_orphan_folder2"
+        case .debugFavoriteTopLevelMissing2: return "m_d_bookmark_top_level_missing2"
```
THISISDINOSAUR: Since this has changed, make sure this doesn't overwrite your PR fixing it

Core/PixelEvent.swift
```diff
@@ -320,10 +320,13 @@ extension Pixel {
...
-        case debugBookmarkTopLevelMissing
-        case debugFavoriteOrphanFolder
-        case debugFavoriteTopLevelMissing
+        // 2 suffix added to differenciate from previous pixels when the fix wasn't available
+        case debugBookmarkOrphanFolder2
+        case debugBookmarkTopLevelMissing2
+        case debugFavoriteOrphanFolder2
+        case debugFavoriteTopLevelMissing2
+        case debugCouldNotFixBookmarkFolder
+        case debugCouldNotFixFavoriteFolder
```
THISISDINOSAUR, line 330: I don't love the 2 suffix, I'd rather have something more meaningful (esp since when looking at pixels themselves we won't have the context of the comment)

Core/BookmarksCoreDataStorage.swift
```diff
@@ -1004,4 +1013,99 @@ public class BookmarksCoreDataStorageMigration {
     }
 }
 
+
+// MARK: - CoreData structure fixer
+// https://app.asana.com/0/414709148257752/1202779945035904/f
+
+extension BookmarksCoreDataStorage {
+    
+    private func deleteExtraOrphanedFolders(_ orphanedFolders: [BookmarkFolderManagedObject],
```
THISISDINOSAUR, line 1023: I'm still on the fence about including this at all since it doesn't seem to ever actually occur at the moment. I'm okay with it on the understanding that it's temporary, but, if we're pretty sure the issue is the other thing, including this might just muddy the waters on what actually fixed it

THISISDINOSAUR: I assume also this will only ever be called with one type of folder?

Core/BookmarksCoreDataStorage.swift
```diff
@@ -1004,4 +1013,99 @@ public class BookmarksCoreDataStorageMigration {
...
+        
+        // Create root folder
+        let bookmarksFolder: BookmarkFolderManagedObject
+        if folderType == .favorite {
+            bookmarksFolder = BookmarksCoreDataStorage.rootFavoritesFolderManagedObject(context)
+        } else {
+            bookmarksFolder = BookmarksCoreDataStorage.rootFolderManagedObject(context)
+        }
+        
+        // Assign all bookmarks to the parent folder
```
THISISDINOSAUR, line 1091: We talked about how this will clobber the existing folder structure, but there were issues when not doing that. 
I'm okay with it since if this situation even happens, something unknowable has gone wrong anyway, so it's not surprising surprising things go wrong

Core/BookmarksCoreDataStorage.swift
```diff
@@ -1004,4 +1013,99 @@ public class BookmarksCoreDataStorageMigration {
...
+            let fetchRequest = NSFetchRequest<BookmarkFolderManagedObject>(entityName: Constants.folderClassName)
+            fetchRequest.predicate = NSPredicate(format: "%K == nil AND %K == %@",
+                                                 #keyPath(BookmarkManagedObject.parent),
+                                                 #keyPath(BookmarkManagedObject.isFavorite),
+                                                 NSNumber(value: folderType == .favorite))
+            
+            let results = try? privateContext.fetch(fetchRequest)
+            
+            if let orphanedFolders = results, orphanedFolders.count > 1 {
+               deleteExtraOrphanedFolders(orphanedFolders, onContext: privateContext, withFolderType: folderType)
```
THISISDINOSAUR: Funky indentation

THISISDINOSAUR (commented):
I still need to test

Core/BookmarksCoreDataStorage.swift
THISISDINOSAUR: Oh also, have you updated the asana task with info about which scenario actually occurs?

THISISDINOSAUR (approved):

Core/PixelEvent.swift
Bunn: @THISISDINOSAUR  any suggestion?
maybe debugFavoriteFixOprhanFolder / debugBookmarkFixOprhanFolder?

THISISDINOSAUR: I'd maybe just replace the 2 with new, which should give a reasonable hint that the one without any further context is old

Core/BookmarksCoreDataStorage.swift
Bunn: @THISISDINOSAUR as for keeping both fixes and muddying the water, I don't think it's a problem since we have pixels for all the fixes as well, so we'll know if just one or both are being called.

THISISDINOSAUR (approved):

Bunn:
@THISISDINOSAUR I've added extra pixels as discussed on asana, could you take a quick look just to make sure the names are good? here's the commit: https://github.com/duckduckgo/iOS/pull/1276/commits/8fda00f49c06e6e1bd769ab7e81bf1a982f7f95f

thanks

THISISDINOSAUR (approved):

PR #1277, opened 2022-08-16
Add manual login adding to autofill settings

THISISDINOSAUR:
Task/Issue URL: https://app.asana.com/0/0/1202801587977636/f
Tech Design URL:
CC:

**Description**:
Also does some other things relating to the editing interface (like the cancel button)

**Steps to test this PR**:
1. Test adding a login from settings. Test editing the login straight after
2. Test editing logins and that nothing broke there. 3. 
3. There's still a lack of polish in general, so just testing for regressions

**Copy Testing**:

* [ ] Use of correct apostrophes in new copy, ie `’` rather than `'`

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPad

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:
[Software Engineering Expectations](https://app.asana.com/0/59792373528535/199064865822552)
[Technical Design Template](https://app.asana.com/0/59792373528535/184709971311943)
**When ready for review, remember to post the PR in MM**

PR #1278, opened 2022-08-16
Autofill in browser changes

THISISDINOSAUR:
Task/Issue URL: https://app.asana.com/0/0/1202806483350357/f
Tech Design URL:
CC:

**Description**:
- Removes blur backgrounds
- Introduces a third button type
- Reenables the key icon
- Stops the saving interface and key icons showing if auth isn't available

**Steps to test this PR**:
1. Test saving and autofill backgrounds look correct. Particularly test both manually and automatically expanding the autofill prompt to see if the background does anything weird (Since it used to have screwy logic to change colour, and now should just remain the same)
1. The autofill prompt should have a new button type for other login items than the primary (should now have a border)
1. Check the key icon appears whenever there is a saved login for a site. Check clicking it works, and just generally nothing is weird with the key
1. Disable auth and check that the key doesn't appear, and neither autofill interface appears (both saving and selecting)

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPad

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:


Bunn (approved):
LGTM, I found some issues that are probably unrelated to this PR but I wanted to point out
ps. I've also tested the steps of this PR: https://github.com/duckduckgo/iOS/pull/1277

1: Website needs to be reloaded in order for the autofill settings to be respected. This is probably expected, but it would be nice to send the flag to the autofill script so it could handle itself without reloading.
Simulator Screen Recording - iPhone 13 - 2022-08-16 at 19 14 48

2:  I got into a loop when trying to use the password on hacker news
Simulator Screen Recording - iPhone 13 - 2022-08-16 at 19 08 44

3: After tapping on "More Options" if you manually swipe the modal back to its original size, we still display all options instead of "More Options". I think this is fine since the user can swipe it back up, but I thought it was worth mentioning.
Simulator Screen Recording - iPhone 13 - 2022-08-16 at 19 04 01

Bunn:
ps. I've added some changes that you might want to merge into this branch here: https://github.com/duckduckgo/iOS/pull/1279
Friendly reminder to not merge this into dev if possible because of [this](https://app.asana.com/0/414235014887631/1202144584700632/f)

THISISDINOSAUR:
Great, I merged your PR into this one, thanks!
RE 1. that is by design since reloading the page can potentially be destructive. Tbh I just don't think there's a good solution to this, but think it's fine so people are unlikely to be turning this on and off a lot
2. What happened if you just dismissed the popup the second time?
3. Yeah, this is deliberate, I was both surprised and pleased about it when asking Chris what we should do in this scenario 😅

Bunn:
@THISISDINOSAUR 

1 - Makes sense.
2 - If I dismiss it, it goes away. Interesting that the key icon doesn't appear on hacker news, but if I tap on the right side of the text field it opens the login modal (I don't think it's related to this PR)
3 - Cool, just checking, thanks for the explanation :)

THISISDINOSAUR:
2. Oh interesting, I'm gonna blame CSPs

PR #1279, opened 2022-08-16
Improve keyboard handling on autofill edit screen

Bunn:
@THISISDINOSAUR 
While testing https://github.com/duckduckgo/iOS/pull/1278 I noticed that using the software keyboard was quite annoying 
I wonder who forgot to add keyboard config into the `ClearTextField` struct 
(spoiler alert, it was me 😅🙃)

This wasn't a big deal when just editing data but when adding fresh new data it became an issue for me.

This PR is pointing to your PR (https://github.com/duckduckgo/iOS/pull/1278) branch, feel free to close it if you don't want this fix :)


THISISDINOSAUR:
Oh awesome, this was on my list of things to do, thanks so much! For now I was just doing the strictly new stuff, rather than any of the tweaks to the existing stuff, but it's a fair point this hurts the new flow more

PR #1283, opened 2022-08-19
Autofill settings list deletion undo

THISISDINOSAUR:
Task/Issue URL: https://app.asana.com/0/0/1202808746542215/f
Tech Design URL:
CC:

**Description**:
Originally I was going to bundle all the settings list changes into one PR, but I think this is potentially worth some separate discussion. 

This PR adds a toast when ever you delete a login item from the settings list, with an undo feature that allows you to undelete them.

I'll comment in line, but even though I wrote it, I'm not sure I love where this lives and think it's at least worth a conversation.

**Steps to test this PR**:
1. Ensure existing toasts haven't been broken. Things this PR could have accidentally affect are both the positioning relative to the bottom of the screen, and the actions when the button on them is pressed.
2. Delete some logins, check you get the toast and it's not too high up the screen. Try undoing some of them, they should immediately reappear in the list. Try editing those and check they work correctly

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPad

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:


DuckDuckGo/AutofillLoginListViewModel.swift
```diff
@@ -195,12 +209,34 @@ final class AutofillLoginListViewModel: ObservableObject {
...
+        }
+    }
+    
+    private func undelete(_ account: SecureVaultModels.WebsiteCredentials) {
+        guard let secureVault = try? SecureVaultFactory.default.makeVault(errorReporter: SecureVaultErrorReporter.shared),
+              var cachedDeletedCredentials = cachedDeletedCredentials else {
+            return
+        }
+        do {
+            // We need to make a new account object. If we try to use the old one, secure vault will try to process it as an update, which will fail
```
THISISDINOSAUR, line 233: There are two parts to this PR that I don't love:
1. that it needs to have knowledge of this. Secure vault might change at some point and break assumptions like this
2. That the caching mechanism doesn't live in secure vault. Given how simple it is, I'm not sure it's worth living anywhere else, but would like a second opinion

DuckDuckGo/AutofillLoginListViewModel.swift
```diff
@@ -195,12 +209,34 @@ final class AutofillLoginListViewModel: ObservableObject {
...
+            cachedDeletedCredentials = try secureVault.websiteCredentialsFor(accountId: accountID)
             try secureVault.deleteWebsiteCredentialsFor(accountId: accountID)
+            return true
+        } catch {
+            Pixel.fire(pixel: .secureVaultError)
+            return false
+        }
+    }
+    
+    private func undelete(_ account: SecureVaultModels.WebsiteCredentials) {
```
samsymons, line 227: How do you feel about integrating this with [`UndoManager`](https://developer.apple.com/documentation/foundation/undomanager)? That might be a bit of a stretch goal since we have our own custom undo UI though, so I don't consider it a blocker to leave this for now.

DuckDuckGo/AutofillLoginListViewModel.swift
samsymons: For now, I think this is okay. I'm interesting in adding `UndoManager` support to this UI on macOS, at which point I would take on this work in a project and handle it in a more cross-platform way. For this simple feature, having a bit of logic here in iOS seems acceptable – it's easy to change once we do make some cross-platform friendly changes.

The first concern is definitely a real one... how about adding a unit test for this?

THISISDINOSAUR: I'm happy with that as discussed on MM. Great shout on the unit test (unit tests or lack thereof are why this has the do not merge label atm), that would probably make me happy enough with it

DuckDuckGo/AutofillLoginListViewModel.swift
THISISDINOSAUR: I'm struggling to think of a benefit that makes it worth it tbh (i.e. if this was going to be the totality of our undoing support on iOS or possibly even macOS, I think not using undo manager would be the way to go long term.). I could be convinced if we expect to have much more undoing and/or redoing in future

Bunn: I don't think UndoManager is worth for this case. IMO one of the main benefits of UndoManager is controlling the stack and here we don't have a stack, it's only a single operation. I wouldn't be opposed of using it (specially for semantic reasons), but I don't see a real need here.

DuckDuckGo/AutofillLoginListViewModel.swift
```diff
@@ -195,12 +209,34 @@ final class AutofillLoginListViewModel: ObservableObject {
         }
     }
     
-    private func delete(_ account: SecureVaultModels.WebsiteAccount) {
+    @discardableResult
+    private func delete(_ account: SecureVaultModels.WebsiteAccount) -> Bool {
         guard let secureVault = try? SecureVaultFactory.default.makeVault(errorReporter: SecureVaultErrorReporter.shared),
-              let accountID = account.id else { return }
+              let accountID = account.id else { return false }
```
Bunn, line 215: Do we care about firing a pixel in the case it returns false here?

Bunn (approved):
✅ Download toasts layout and action still working 
✅ Undo working and not leaving trace if the app is force closed in-between
⚠️ There's a missing animation when you undo a deletion, not a big concern ATM but would be nice to have it once the item is added back to the list
⚠️ Just re-enforcing what you already said, would be good to have a unit test for the undo operation

✅ LGTM. I'll let you decide if you want me to take a second look after you add the unit test. Feel free to merge it after you do.

DuckDuckGo/AutofillLoginListViewModel.swift
THISISDINOSAUR: Yeah, quite possibly, I'll take a look

THISISDINOSAUR: The secure vault will automatically fire a pixel if that fails, so I think it's fine

THISISDINOSAUR:
We talked on MM about the difficulties of unit testing this and how I don't think there's a meaningful test we can write. Neither securevault or secureVaultFactory are set up to be mocked, and even if they were, the resulting test would be pointless given that I want to test this assumption "// We need to make a new account object. If we try to use the old one, secure vault will try to process it as an update, which will fail".

So far now I think this should proceed without a unit test, but definitely want to have a broader conversation about secure vault mocking before this project is done

PR #1295, opened 2022-09-02
Autofill Batch 2 in browser experience changes

amddg44:
Task/Issue URL: https://app.asana.com/0/0/1202854530787111/f
Tech Design URL:
CC:

**Description**:

- Prevent autofill being accidentally disabled
  - Copy updates
  - Alert box to confirm disable 
  - Note that the new logic around applying new rules for when to show the confirm disable alert is not part of this PR 
- Make sure people know that DDG is saving passwords updates

**Steps to test this PR**:
1. Confirm that copy on first save sheet is correct
2. Confirm first save sheet is shown for all normal save sheets
3. Test that alert prompts to confirm disabling autofill after 3 rejections and that buttons perform as expected

Note that the new logic around applying new rules for when to show the confirm disable alert is not part of this PR 

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPad

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:


THISISDINOSAUR (commented):
I know this isn't a regression and that you've deliberately not touched the logic of when to show the prompt to disable autofill, but I think the prompt makes the current poor logic a bit too conspicuous.
Even if I've saved a login before (or lots of logins), I still see the prompt to disable if I dismiss it three times, when I think having saved any login means we should never prompt to disable.
I also see the prompt on any subsequent time I press don't save.

I think we should probably improve that a bit before merging, wdyt?

amddg44:
> I know this isn't a regression and that you've deliberately not touched the logic of when to show the prompt to disable autofill, but I think the prompt makes the current poor logic a bit too conspicuous. Even if I've saved a login before (or lots of logins), I still see the prompt to disable if I dismiss it three times, when I think having saved any login means we should never prompt to disable. I also see the prompt on any subsequent time I press don't save.
> 
> I think we should probably improve that a bit before merging, wdyt?

Agree + sounds good to me 👍

THISISDINOSAUR (approved):
Approved now the changed prompt logic is in

PR #1302, opened 2022-09-08
Tie toast message to navigation bar

mallexxx:
Task/Issue URL: https://app.asana.com/0/414709148257752/1201884709997737
CC: @miasma13 

**Description**:
Ties toast message to the toolbar when browser chrome is hidden when scrolling

**Steps to test this PR**:
1. When a Toast Message is shown (e.g. when download has started), validate hiding browser toolbar adjusts the toast position; Validate using fast scrolling does the same.
2. Validate the Toast Message position is adjusted when an overlay view controller is presented/dismissed (e.g. when choosing Open Downloads for a completed download notification)
3.  Make sure if presented View Controller also has a toolbar the Toast respects its height

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPad

---
###### Internal references:

### DIFF
--- a/DuckDuckGo/MainViewController.swift
+++ b/DuckDuckGo/MainViewController.swift
@@ -30,6 +30,10 @@ import BrowserServicesKit
 class MainViewController: UIViewController {
 // swiftlint:enable type_body_length
 
+    static var sharedInstance: MainViewController {
+        (UIApplication.shared.delegate?.window??.rootViewController as? MainViewController)!
+    }
+
     override var preferredStatusBarStyle: UIStatusBarStyle {
         return ThemeManager.shared.currentTheme.statusBarStyle
     }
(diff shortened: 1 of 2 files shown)


DuckDuckGo/MainViewController.swift
```diff
@@ -30,6 +30,10 @@ import BrowserServicesKit
 class MainViewController: UIViewController {
 // swiftlint:enable type_body_length
 
+    static var sharedInstance: MainViewController {
+        (UIApplication.shared.delegate?.window??.rootViewController as? MainViewController)!
+    }
```
miasma13, line 35: We shouldn't add helpers that promote accessing MainVC other than that it comes naturally from parent-child relationship.

miasma13 (changes requested):
I do not think it is a good approach to implement it and it does not follow hints proposed in original task. To follow the advised patterns for elements shown within the MainVC please take a look at: MainVC's properties e.g.: `notificationContainer`, `notificationContainerTop`, `notificationContainerHeight` that do similar thing with the notification box shown between OmniBar and WebView. We should follow the same pattern for the ActionMessageView on the bottom.

Another thing to think about would be rethinking API for ActionMessage presentation (since it won't be universal accessible from anywhere but presentation would be in MainVC). Maybe having ActionMessage class exposing API and communicating via NotificationCenter?

PR #1310, opened 2022-09-14
Hung page detection

mallexxx:
Task/Issue URL: https://app.asana.com/0/72649045549333/1203277006424404/f
Tech Design URL: https://app.asana.com/0/0/1202554826082952/f
CC: @Bunn @miasma13 

**Description**:
This PR fixes cyclic alerts/hung JS problem causing the whole Tab‘s navigation to stop responding

**Steps to test this PR**:
1. Create a local page (put it into /Library/WebServer/Documents) with a JS script `while (true) alert("tada!");` or `while (true) ;`
2. Navigate to your machine IP address, dismiss couple of alerts
3. Validate navigation back/forward/to a back/forward list item/refresh/navigation to a URL takes you away from the page and alerts are suppressed until the next visit
4. Validate the hanging detection implementation doesn‘t break other (normal) navigations flows 
5. Hang the page, hit Reload and switch to another tab
6. Hang the page, tap the page several times and hit Reload

**Device Testing**:

* [ ] iPhone
* [ ] iPad

**OS Testing**:

* [x] iOS 14
* [x] iOS 15

---
###### Internal references:

### DIFF
--- a/Core/WKBackForwardListExtension.swift
+++ b/Core/WKBackForwardListExtension.swift
@@ -28,14 +28,14 @@ public extension WKBackForwardList {
     ///   If `backForwardListItem` is not in the list, returns `nil`.
     ///   When the found index is passed to `item(at: index)` will return the searched item.
     func index(of backForwardListItem: WKBackForwardListItem) -> Int? {
+        if currentItem == backForwardListItem {
+            return 0
+        }
         if let idx = backList.firstIndex(of: backForwardListItem) {
             return -idx - 1
         }
         if let idx = forwardList.firstIndex(of: backForwardListItem) {
             return idx + 1
-        }
-        if currentItem == backForwardListItem {
-            return 0
         }
         return nil
     }
--- a/Core/WKBackForwardListExtension.swift
+++ b/Core/WKBackForwardListExtension.swift
@@ -28,6 +28,9 @@ public extension WKBackForwardList {
         if let idx = forwardList.firstIndex(of: backForwardListItem) {
             return idx + 1
         }
+        if currentItem == backForwardListItem {
+            return 0
+        }
         return nil
     }
 
--- a/Core/WKBackForwardListExtension.swift
+++ b/Core/WKBackForwardListExtension.swift
@@ -1,0 +1,34 @@
+//
+//  WKBackForwardListExtension.swift
+//  DuckDuckGo
+//
+//  Copyright © 2022 DuckDuckGo. All rights reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//  http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import WebKit
+
+public extension WKBackForwardList {
+
+    func index(of backForwardListItem: WKBackForwardListItem) -> Int? {
+        if let idx = backList.firstIndex(of: backForwardListItem) {
+            return -idx - 1
+        }
+        if let idx = forwardList.firstIndex(of: backForwardListItem) {
+            return idx + 1
+        }
+        return nil
+    }
+
+}
--- a/Core/WKBackForwardListExtension.swift
+++ b/Core/WKBackForwardListExtension.swift
@@ -21,6 +21,12 @@ import WebKit
 
 public extension WKBackForwardList {
 
+    /// Returns the offset of the searched item from the Current Item
+    /// - Parameter backForwardListItem: An item to search for in the list.
+    /// - Returns: When found, returns 1-based index from the Current Item, positive for the `forwardList`,
+    ///   negative for the `backList`. If `backForwardListItem` matches the current item, returns `0`.
+    ///   If `backForwardListItem` is not in the list, returns `nil`.
+    ///   When the found index is passed to `item(at: index)` will return the searched item.
     func index(of backForwardListItem: WKBackForwardListItem) -> Int? {
         if let idx = backList.firstIndex(of: backForwardListItem) {
             return -idx - 1
--- a/DuckDuckGo/TabViewController.swift
+++ b/DuckDuckGo/TabViewController.swift
@@ -270,7 +270,7 @@ class TabViewController: UIViewController {
                 }
                 os_log("""
                     -----------------------------------------------------------------------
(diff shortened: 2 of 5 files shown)


DuckDuckGo/TabViewController.swift
```diff
@@ -455,6 +472,34 @@ class TabViewController: UIViewController {
...
+    private func recreateWebView(legacyNavigationURL: URL?) -> WebViewRecreationResult {
+        var sessionStateData: Any?
+        if #available(iOS 15.0, *) {
+            sessionStateData = self.webView.interactionState
+        }
+        let configuration = self.webView.configuration
+        let url = self.webView.url
+
+        self.webView.removeFromSuperview()
+        self.webView = nil
```
bwaresiak, line 522: I wonder if we should have some more explicit cleanup (e.g. removing delegates, observers, etc), I know most of that happens in deinint, but still probably wouldn't hurt? WDYT?

DuckDuckGo/TabViewController.swift
```diff
@@ -486,16 +531,25 @@ class TabViewController: UIViewController {
...
 
         load(url: url, didUpgradeURL: false)
     }
-    
+
     public func load(backForwardListItem: WKBackForwardListItem) {
         webView.stopLoading()
+        scheduleNavigationExpectation(legacyNavigationURL: backForwardListItem.url) { [weak self, url=backForwardListItem.url] in
+            guard let self = self,
+                  let item = (self.webView.backForwardList.backList + self.webView.backForwardList.forwardList).first(where: { $0.url == url })
```
bwaresiak: Maybe we can also compare initial url to better match correct entry in case there are multiple matching urls in the collection?

DuckDuckGo/TabViewController.swift
```diff
@@ -245,7 +247,22 @@ class TabViewController: UIViewController {
...
-    
+
+    private var navigationExpectationTimer: Timer? {
+        willSet {
+            navigationExpectationTimer?.invalidate()
+        }
+    }
+
+    private static let navigationExpectationInterval = 3.0
+    private func scheduleNavigationExpectation(legacyNavigationURL: URL?, replay: (() -> Void)? = nil) {
```
bwaresiak: Naming complaints :) This is not in fact `legacyNavigationURL` but rather `(current)destinationURL` as we are navigating to that site.

Also I wonder if we could be more explicit like `onWebViewRestored` instead of `replay`?

bwaresiak (changes requested):
Overall great job, it's a pity WebView cannot restore from that scenario. 

I've left few comment inline, but also it would be worthwhile to create test webpage for this scenario in https://github.com/duckduckgo/privacy-test-pages for easier testing - can you make PR?

mallexxx:
@bwaresiak this should be ready for another pass now;
I‘ve made the changes you requested and added instant recreation on recursive alerts detection and multiple navigation requests
Opened test pages PR: https://github.com/duckduckgo/privacy-test-pages/pull/101

In fact there is _webViewWebProcessDidBecomeUnresponsive: NavigationDelegate method which is called by a private Responsiveness timer (basically the same as the navigationExpectationTimer) but it‘s invalidated on any page tap made by a user so our implementation should be better

mallexxx:
TODO: make it explicit in DEBUG

mallexxx:
@bwaresiak as discussed, I've added DEBUG log instructions on how to disable the termination behaviour and a breakpoint to make sure it gets noticed

DuckDuckGo/TabViewController.swift
```diff
@@ -486,16 +567,29 @@ class TabViewController: UIViewController {
...
         dismissJSAlertIfNeeded()
 
         load(url: url, didUpgradeURL: false)
     }
-    
+
     public func load(backForwardListItem: WKBackForwardListItem) {
         webView.stopLoading()
+        let destinationURL = backForwardListItem.url
+        var itemIndex = webView.backForwardList.index(of: backForwardListItem)
```
bwaresiak: This should be let to silence lint warning

Core/WKBackForwardListExtension.swift
```diff
@@ -0,0 +1,37 @@
...
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import WebKit
+
+public extension WKBackForwardList {
+
+    func index(of backForwardListItem: WKBackForwardListItem) -> Int? {
```
bwaresiak, line 30: Naming nitpick: change `index(of:)` to `relativeOffset(of:)` ?

DuckDuckGo/TabViewController.swift
```diff
@@ -245,7 +247,55 @@ class TabViewController: UIViewController {
...
+    private var navigationExpectationTimer: Timer? {
+        willSet {
+            navigationExpectationTimer?.invalidate()
+        }
+    }
+
+    private static let navigationExpectationInterval = 3.0
+    private func scheduleNavigationExpectation(destinationURL: URL?, onSessionRestored: (() -> Void)? = nil) {
+        // continuous schedule calls may mean user is trying to hit all the buttons while we‘re waiting
+        guard !forceExpectedNavigationIfNeeded() else { return }
```
bwaresiak, line 260: I get a lot of expirations triggered from here when doing regular navigations - I don't think that is expected.

bwaresiak (changes requested):
Debug changes look great, made it easy to see some weird issue (listed inline).

Core/WKBackForwardListExtension.swift
mallexxx: why? it‘s WKBackForwardList.itemAtIndex: method getting an item so it matches the indexOfItem:

bwaresiak: I sort of dislike that documentation states `index - The offset of the desired item from the current item.` that's why I'd prefer to be explicit. Your call, if you leave it like this can you point to related method and explain that index is in fact an offset?

DuckDuckGo/TabViewController.swift
mallexxx: Do you have a particular example for this or just random occurrences?

bwaresiak: Just navigating using URL bar. I'm running this on iOS 15.5 simulator.

Core/WKBackForwardListExtension.swift
mallexxx: added help message for it and renamed `var index` to `itemOffset` in the caller

DuckDuckGo/TabViewController.swift
mallexxx: these expirations must‘ve been caused by recursive requests sent from `webView:decidePolicyForNavigationAction:`, should be fixed now by resetting the timer in this method

bwaresiak (approved):
Great job, tested this again and got one random firing (when doing plenty of reload/back/forward) but it was so smooth and did not disturb the flow that I don't see any problem. :)

PR #1335, opened 2022-10-07
Update BSK reference for script execution guard

jonathanKingston:
Task/Issue URL: https://app.asana.com/0/1200437802575119/1202155258470004/f
Tech Design URL:
CC:

**Description**:

Depends on: https://github.com/duckduckgo/BrowserServicesKit/pull/148

Visiting https://www.washingtonpost.com/ without this patch you'll see "TypeError: Attempting to change value of a readonly property." errors as we're creating a property called firefox.find (in the isolated world) and then trying to redeclare over it.

This is only a side effect of us executing all user scripts multiple times for about:blank frames; most of our code isn't idempotent so will be creating side effects.

The solution is a little hacky here but we create a guard that prevents the script from executing multiple times for the same execution context.

-  Note when debugging this; we run some scripts in isolated and some in default worlds so the variable is created both times. The issue exists in both worlds but the array is created both times so you'll see different sized arrays as you step through etc (this is because isolated worlds have a totally different window global) the code works as expected for the isolated also.

**Steps to test this PR**:
1.
2.

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPhone 14 Pro
* [ ] iPad

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:

### DIFF
--- a/DuckDuckGo.xcodeproj/project.pbxproj
+++ b/DuckDuckGo.xcodeproj/project.pbxproj
@@ -7161,7 +7161,7 @@
 			repositoryURL = "https://github.com/duckduckgo/BrowserServicesKit";
 			requirement = {
 				kind = exactVersion;
-				version = 32.0.2;
+				version = 32.0.3;
 			};
 		};
 		C14882EB27F211A000D59F0C /* XCRemoteSwiftPackageReference "SwiftSoup" */ = {


samsymons (approved):
LGTM

PR #1352, opened 2022-10-17
Fix the 'cannot access the property' warning in Swift 6

jaceklyp:
Task/Issue URL: https://app.asana.com/0/414709148257752/1203137819376967/f

**Description**:
An attempt to fix the 'cannot access the property' warning in Swift 6
Also, use the UUID to uniquely identify the given tab for rules monitor.

**Steps to test this PR**:
1. Make sure there is no warning.
2. Make sure the time being reported in deinit method is still correct (>0).

### DIFF
--- a/DuckDuckGo/RulesCompilationMonitor.swift
+++ b/DuckDuckGo/RulesCompilationMonitor.swift
@@ -26,7 +26,7 @@ final class RulesCompilationMonitor {
     
     private var didReport = false
     private var waitStart: TimeInterval?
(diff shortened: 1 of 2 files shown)


mallexxx (changes requested):
I think this can be made more consistent with macOS implementation, maybe event moved to BSK?
please check https://github.com/more-duckduckgo-org/macos-browser/blob/3626c5149c17a8998b9d8f69017d98ebbb31bf76/DuckDuckGo/Browser%20Tab/Model/Tab.swift#L1384
https://github.com/more-duckduckgo-org/macos-browser/blob/develop/DuckDuckGo/Statistics/RulesCompilationMonitor.swift

PR #1378, opened 2022-11-11
Fix reload triggering process kill on User Scripts Installed

mallexxx:
Task/Issue URL: https://app.asana.com/0/414235014887631/1203338569768096/f
CC: @brindy @samsymons 

**Description**:
- Resets navigation expectation timer on UserScripts Installed delegate call
- Adds threshold before triggering force process kill

**Steps to test this PR**:
1. Validate the following scenario doesn‘t trigger process killing:

- Open in new tab
- Load request
- Scheduler timer
- Content scripts get set up
- Webview reloaded
- Schedule timer
 - Timer already exists
 - Fire timer prematurely

### DIFF
--- a/Core/PixelEvent.swift
+++ b/Core/PixelEvent.swift
@@ -318,6 +318,7 @@ extension Pixel {
         
         case webKitDidTerminate
         case webKitDidBecomeUnresponsive
+        case expectedNavigationForcedWebViewRecreation
         case webKitTerminationDidReloadCurrentTab
         
         case backgroundTaskSubmissionFailed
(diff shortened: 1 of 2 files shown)


brindy:
I haven't looked at this in detail but I traced back the work via #1310 to the original tech design and am concerned we didn't think this through properly.  

There's no release this week so I'm marking this as do not merge and we can decide what to do next week when @bwaresiak is back.

cc @catehstn, @bwaresiak

Core/PixelEvent.swift
```diff
@@ -670,6 +671,7 @@ extension Pixel.Event {
             
         case .webKitDidTerminate: return "m_d_wkt"
         case .webKitDidBecomeUnresponsive: return "m_d_wkunresponsive"
+        case .expectedNavigationForcedWebViewRecreation: return "m_d_wkforceterminate"
```
bwaresiak: We probably can remove this right?

bwaresiak:
Other than this one comment above, this change looks good to me. As I am unable to reproduce the problem, I'd like to defer to @samsymons or @brindy to do final validation.

samsymons (approved):
LGTM once Bartek's comment is addressed, I can no longer reproduce the issue. I spent some time stress testing tab functionality in general, and it seems good. :+1:

PR #1380, opened 2022-11-11
AutoCompleteRequest refactoring

MarlowBrown:
A simple refactoring of AutoCompleteRequest. More clear function calls along with moving parsing logic to a separate class.


samsymons:
Hey @MarlowBrown! I ran these PRs by the rest of the team, but we agreed that we generally don't accept refactoring PRs without an issue to discuss the changes first, as they have a higher-than-average frequency of causing problems with features being developed internally.

I'll close these PRs, but if you feel strongly please feel free to open an issue and we can chat. Thanks! 🙏

PR #1381, opened 2022-11-11
handle crash collection and send pixel

brindy:
Task/Issue URL: https://app.asana.com/0/1200194497630846/1203353338544857/f
Tech Design URL:
CC:

**Description**:

Uses CrashCollection class to receive info about crashes and then send a pixel.

**Testing**:

See https://github.com/duckduckgo/BrowserServicesKit/pull/169

---
###### Internal references:


DuckDuckGo/AppDelegate.swift
```diff
@@ -66,6 +67,10 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         }
         #endif
 
+        CrashCollection.run { params in
+            Pixel.fire(pixel: .dbCrashDetected, withAdditionalParameters: params)
+        }
```
samsymons: Looks like there's a SwiftLint warning here to clean up

```suggestion
        }
```

DuckDuckGo/AppDelegate.swift
```diff
@@ -66,6 +67,10 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         }
         #endif
 
+        CrashCollection.run { params in
+            Pixel.fire(pixel: .dbCrashDetected, withAdditionalParameters: params)
```
samsymons: Do you happen to know how long the OS will keep crashes around locally? I'm wondering how big of a spike we'll see once this goes live to customers, where there are presumably many of these reports cached and ready to deliver via MetricKit.

I don't think it's a blocker as it's a one-time issue that'll stabilize quickly, but would love to know what to expect.

samsymons (commented):
I'm seeing a new build warning after running this branch:

```
Linking against a dylib which is not safe for use in application extensions: /Users/samsymons/Library/Developer/Xcode/DerivedData/DuckDuckGo-crkltdebenwskgemmdqoruykedfz/Build/Products/Debug-iphonesimulator/PackageFrameworks/Crashes.framework/Crashes
```

DuckDuckGo/AppDelegate.swift
brindy: I had a look in the docs, it just says "The diagnostic reports for fatal and nonfatal CPU exceptions for the app during the reporting period."

I've added a flag which gets sent with the first group of pixels, then gets turned off.  I also added some unit tests for that.  Let me know what you think.  

Thanks!

samsymons (approved):
LGTM, I tested first and subsequent runs, both with 1 and multiple crashes to confirm that we're sending a pixel on each crash. Seems good!

PR #1392, opened 2022-11-17
Fix text size setting on page load

miasma13:
Task/Issue URL: https://app.asana.com/0/414709148257752/1203360000128186/f

**Description**:
The text size setting stopped being applied on page load due to recent changes how user scripts are initiated and delayed passing of the stored text size setting value. As a solution the current text size setting value should be passed at TextSizeUserScript initialization.

**Steps to test this PR**:
1. Open Settings -> Text Size
2. Adjust the value with slider
3. Reload or open new page
4. The text size should be adjusted to specified value on page load

**Device Testing**:

* [ ] iPhone
* [ ] iPad

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16

---
###### Internal references:

### DIFF
--- a/Core/TextSizeUserScript.swift
+++ b/Core/TextSizeUserScript.swift
@@ -32,6 +32,10 @@ public class TextSizeUserScript: NSObject, UserScript {
     public var injectionTime: WKUserScriptInjectionTime = .atDocumentStart
     public var forMainFrameOnly: Bool = false
     public var messageNames: [String] = []
+    
+    public init(textSizeAdjustmentInPercents: Int) {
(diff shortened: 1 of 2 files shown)


samsymons (approved):
LGTM, tested this on a few sites and it works well.

I noticed that it doesn't work with the NY Times, but I believe that was the case even from the beginning (correct me if I'm wrong), so I don't think it blocks this PR.

miasma13:
> I noticed that it doesn't work with the NY Times, but I believe that was the case even from the beginning (correct me if I'm wrong), so I don't think it blocks this PR.

You are correct here.

PR #1393, opened 2022-11-18
Fix how we setup temporary download for sharing a file

miasma13:
Task/Issue URL: https://app.asana.com/0/414235014887631/1203216489723509/f

**Description**:
For some sites the assets may be loaded dynamically within non-main frames. Since we lacked that check we could have setup in rare cases the temporary download for one of such assets where none should be present (e.g. we were navigating to a standard HTML page). As a result share action resulted in sharing that asset instead of a page's URL.

**Steps to test this PR**:
1. Navigate to https://www.rockymountaineer.com 
2. Tap share
3. (should share website's url not an image or any other asset from the file)
4. Browse through the website and verify for subpages
5. Smoketest opening non-HTML URLs and sharing

---
###### Internal references:


jaceklyp (approved):
LGTM :)

PR #1414, opened 2022-12-05
Autofill reduced authentication friction for multi login processes

amddg44:
Task/Issue URL: https://app.asana.com/0/0/1203313768964242/f
Tech Design URL:
CC: @THISISDINOSAUR 

**Description**:
Reduction of autofill authentication friction as per rules defined here: https://app.asana.com/0/0/1203091315317194/f

**Steps to test this PR**:

**Fill Login**
1. Go to a site with a 2 step auth flow e.g. amazon.com that you have a saved login for
2. Use Autofill to populate the email address. You should have to authenticate
3. Use Autofill to populate the password. You should have not need to authenticate
4. Logout and wait 15 seconds from your last authentication
5. Repeat steps 2 and 3, but this time wait 15 seconds before trying to populate the password. This time you should have to authenticate at both stages
6. Logout and wait 15 seconds from your last authentication
7. Use Autofill to populate the email address. You should have to authenticate
8. Minimise the app. When you re-open, try to autofill the password; you should have to authenticate

**Autofill screens**
1. Wait 15 secs from last authentication, then open the Autofill logins screen from the browsing menu and authenticate.
2. Exit out of the screen and then go back in. You should not need to authenticate
3. Exit out of the screen, wait 15 seconds and the go back in. You should have to authenticate
4. Tap through into a saved login. Drag down to dismiss the Autofill screens (but don't background the app)
5. From the browsing menu, access Autofill logins and you should be taken straight to the last saved login that you accessed, and not prompted to authenticate
6. From the Settings screen, access Autofill logins and you should be only taken to the main logins list screen, not the saved login

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPhone 14 Pro
* [ ] iPad

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:

### DIFF
--- a/DuckDuckGo/AutofillLoginSession.swift
+++ b/DuckDuckGo/AutofillLoginSession.swift
@@ -22,11 +22,15 @@ import BrowserServicesKit
 
 class AutofillLoginSession {
 
+    private enum Constants {
+        static let timeout: TimeInterval = 15
+    }
+
     private var sessionCreationDate: Date?
     private var sessionAccount: SecureVaultModels.WebsiteAccount?
     private let sessionTimeout: TimeInterval
 
-    init(sessionTimeout: TimeInterval = 15) {
+    init(sessionTimeout: TimeInterval = Constants.timeout) {
         self.sessionTimeout = sessionTimeout
     }
 
--- a/DuckDuckGo/AutofillLoginSession.swift
+++ b/DuckDuckGo/AutofillLoginSession.swift
@@ -1,0 +1,55 @@
+//
+//  AutofillLoginSession.swift
+//  DuckDuckGo
+//
+//  Copyright © 2022 DuckDuckGo. All rights reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//  http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+import BrowserServicesKit
+
+class AutofillLoginSession {
+
+    private var sessionCreationDate: Date?
+    private var sessionAccount: SecureVaultModels.WebsiteAccount?
+    private let sessionTimeout: TimeInterval
+
+    init(sessionTimeout: TimeInterval = 15) {
+        self.sessionTimeout = sessionTimeout
+    }
+
+    var isValidSession: Bool {
+        guard let sessionCreationDate = sessionCreationDate else { return false }
+        return Date().timeIntervalSince(sessionCreationDate) < sessionTimeout
+    }
+
+    var lastAccessedAccount: SecureVaultModels.WebsiteAccount? {
+        get {
+            return isValidSession ? sessionAccount : nil
+        }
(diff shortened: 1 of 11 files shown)


THISISDINOSAUR:
On the testing front, it mostly looks good. I also tried locking the screen rather than explicitly backgrounding the app, and that also invalidated the auth :party-parrot:

There's a couple of cases were I'm not sure if the behaviour is wrong, it's debatable, but thought worth flagging. They are also a bit marginal given the short time window, and hard to articulate, so I've made a recording (was quite please I managed to do both in the 15 second time window).
- Step 5, if I reaccess the autofill page, press back, dismiss the autofill page, then open it again, it opens that previous login, but probably shouldn't since I already went back
- Arguably a similar problem (although could be want we want? Not sure, thought it was worth flagging) if I go to autofill from the browsing menu, access a login, dismiss, go through settings to the list, then go back to the browsing menu autofill, it shows the previous item again


https://user-images.githubusercontent.com/1093508/205703929-1008c339-5ad6-44e4-a0d2-28563b088244.mp4

DuckDuckGo/AutofillLoginSession.swift
```diff
@@ -0,0 +1,55 @@
...
+import Foundation
+import BrowserServicesKit
+
+class AutofillLoginSession {
+
+    private var sessionCreationDate: Date?
+    private var sessionAccount: SecureVaultModels.WebsiteAccount?
+    private let sessionTimeout: TimeInterval
+
+    init(sessionTimeout: TimeInterval = 15) {
```
THISISDINOSAUR: Tiny nit, but might be good to move this default into a constant somewhere

DuckDuckGo/AutofillLoginSession.swift
```diff
@@ -0,0 +1,55 @@
...
+        self.sessionTimeout = sessionTimeout
+    }
+
+    var isValidSession: Bool {
+        guard let sessionCreationDate = sessionCreationDate else { return false }
+        return Date().timeIntervalSince(sessionCreationDate) < sessionTimeout
```
THISISDINOSAUR: Side note, if we ever remove the backgrounding invalidation, would it be possible to circumvent the check by changing the phone's clock? If so, it might be worth commenting that somewhere

THISISDINOSAUR (commented):
Besides those two possible ux changes, it looks great! I really appreciate the positive impact you're having on our overall test coverage too!

amddg44:
> * Step 5, if I reaccess the autofill page, press back, dismiss the autofill page, then open it again, it opens that previous login, but probably shouldn't since I already went back

Agree, I think if you back out it shouldn't reopen to the previous login + have made that change

> * Arguably a similar problem (although could be want we want? Not sure, thought it was worth flagging) if I go to autofill from the browsing menu, access a login, dismiss, go through settings to the list, then go back to the browsing menu autofill, it shows the previous item again

This one, I'm not sure about. I'm leaning towards leaving it as is, especially as it's pretty edge case but can be swayed either way

DuckDuckGo/AutofillLoginSession.swift
amddg44: Good suggestion! I've tested changing my clock time back and have added an extra check to protect against this and I've added a comment in code

THISISDINOSAUR:
That all sounds good to me! Will rereview

THISISDINOSAUR (approved):
LGTM!

PR #1416, opened 2022-12-07
Change new autofill login page to be modal

amddg44:
Task/Issue URL: https://app.asana.com/0/0/1203473309989293/f
Tech Design URL:
CC: @THISISDINOSAUR 

**Description**:
Changes the new login page to be modal, and modifies the save behaviour so that once the login has been saved dismiss the screen and then segue to the newly created account. There is no change to the existing view / edit states 

**Steps to test this PR**:
1. From the browsing menu open the Autofill logins screen. Tap to add a new login and confirm it is presented modally. 
2. Confirm cancel button dismisses the screen correctly
3. Save a new login. Confirm the screen is dismissed and that you are segued to the new account + can view / edit as normal
4. Go back to main logins list screen and tap to add another login
5. Try to create a duplicate login and confirm the duplicate login alert shows as before and that the screen is not dismissed if there is an error
6. Repeat above steps accessing the Autofill logins screen from the Settings screen

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPhone 14 Pro
* [ ] iPad

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:

### DIFF
--- a/DuckDuckGo/AutofillLoginDetailsViewController.swift
+++ b/DuckDuckGo/AutofillLoginDetailsViewController.swift
@@ -271,7 +271,7 @@ class AutofillLoginDetailsViewController: UIViewController {
     
     @objc private func cancel() {
         if viewModel.viewMode == .new {
-            navigationController?.popViewController(animated: true)
+            dismiss(animated: true)
         } else {
             toggleEditMode()
         }
--- a/DuckDuckGo/AutofillLoginDetailsViewController.swift
+++ b/DuckDuckGo/AutofillLoginDetailsViewController.swift
@@ -24,7 +24,7 @@ import Common
 import Combine
 
 protocol AutofillLoginDetailsViewControllerDelegate: AnyObject {
-    func autofillLoginDetailsViewControllerDidSave(_ controller: AutofillLoginDetailsViewController)
+    func autofillLoginDetailsViewControllerDidSave(_ controller: AutofillLoginDetailsViewController, account: SecureVaultModels.WebsiteAccount?)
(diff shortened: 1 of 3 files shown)


DuckDuckGo/AutofillLoginDetailsViewController.swift
```diff
@@ -266,12 +266,11 @@ class AutofillLoginDetailsViewController: UIViewController {
     
     @objc private func save() {
         viewModel.save()
-        delegate?.autofillLoginDetailsViewControllerDidSave(self)
```
amddg44, line 269: One note here - I've moved this to `func autofillLoginDetailsViewModelDidSave()` as otherwise it is called regardless of whether or not the save was actually successful

THISISDINOSAUR: Were there any side effects of this being called when it should be?

THISISDINOSAUR (approved):
LGTM!

DuckDuckGo/AutofillLoginDetailsViewController.swift
amddg44: I don't think so since if a login failed to save it shouldn't have created / modified the account in the secure vault, so the refreshing of the login list from secure vault would have been fine, but I can't guarantee there weren't edge cases

PR #1432, opened 2023-01-06
Check metadata length

miasma13:
Task/Issue URL: https://app.asana.com/0/414709148257752/1203659089823819/f

**Description**:
Since App Store metadata fields have strict length requirements and they are only checked   A script to check if metadata fields are of appropriate length. The script can be run without parameter to check the metadata under default path or with a parameter for an external folder.

**Steps to test this PR**:
1. Run `check_metadata_length.sh` inside repo (run without parameter)
2. Modify some of the files under `/fastlane/metadata` to exceed the limits and rerun the `check_metadata_length.sh` script to see if it detects errors.
3. Open the project in Xcode and start a build - the same errors should be displayed in Xcode and should fail the build.

---
###### Internal references:


Bunn (commented):
@miasma13 this looks pretty cool, but I do have one issue with the current approach.
Since it also runs on Xcode, it has increased my build time for ~3seconds

Before:
Screenshot 2023-01-09 at 15 20 04
After:
Screenshot 2023-01-09 at 15 20 54

I don't see a big advantage of having this check being ran every time we build, it's something we can setup a GHA to run on PRs or release builds, but it should probably be removed from the Xcode project itself.

What do you think?

miasma13:
@Bunn Thanks for looking at that. I think your point is reasonable. I thought about 2 biggest benefit we can get from the script:
- being able to run it by hand when you receive new `metadata` package
- doing a check before building, signing and uploading release build to app store (when it fails if they are too long)
So I agree that it may not make sense to run it with every build. What about running the check only for release builds?

Bunn:
> @Bunn Thanks for looking at that. I think your point is reasonable. I thought about 2 biggest benefit we can get from the script:
> 
> * being able to run it by hand when you receive new `metadata` package
> * doing a check before building, signing and uploading release build to app store (when it fails if they are too long)
>   So I agree that it may not make sense to run it with every build. What about running the check only for release builds?

Oh, for sure the script is very useful, it would have saved me some time in the past. My only concern was the increased build time. That said, I can see two ways to improve this:
* 1/ Only do it on release builds as you mentioned.
* 2/ Add a step on bitrise and/or GHA to run the script before the compilation step (to fail early) on the release workflow.

Both approaches would solve the issue without increasing dev build compilation time, so I'm fine with either.

miasma13:
After a thought I like the 2/ approach if we want the automated run on release flows. I will migrate the build step check to run triggered from fastlane on release flow.

Bunn (approved):
LGTM!

PR #1439, opened 2023-01-13
Reset compilation time on entering background

jaceklyp:
Task/Issue URL: https://app.asana.com/0/1200194497630846/1203569850690831/f

**Steps to test this PR**:
1. Leave the app once the compilation has started
2. Go back to foreground and make sure the pixel won't be sent

### DIFF
--- a/DuckDuckGo/AppDelegate.swift
+++ b/DuckDuckGo/AppDelegate.swift
@@ -326,6 +326,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         autoClear?.applicationDidEnterBackground()
         lastBackgroundDate = Date()
         AppDependencyProvider.shared.autofillLoginSession.endSession()
+        ContentBlocking.shared.contentBlockingManager.resetCompilationTime()
     }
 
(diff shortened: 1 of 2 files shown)


DuckDuckGo/AppDelegate.swift
```diff
@@ -326,6 +326,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         autoClear?.applicationDidEnterBackground()
         lastBackgroundDate = Date()
         AppDependencyProvider.shared.autofillLoginSession.endSession()
+        ContentBlocking.shared.contentBlockingManager.resetCompilationTime()
```
bwaresiak, line 329: This approach is fine for current compilation, but what if state of manager is `recompilingAndScheduled`, so subsequent one will be started shortly after?

Also alternative solution: maybe we should be starting background task during compilation so it completes just fine after entering background? Open question there is how would that affect potential reload triggered by it, while app is in the background.

Bunn:
@bwaresiak @jaceklyp anything left to do on this PR?

PR #1450, opened 2023-01-17
Prepare release workflow

jaceklyp:
Task/Issue URL: https://app.asana.com/0/1200194497630846/1203749968800720/f

**Description**:
The script that:
- prepares release or hotfix branch,
- commits necessary changes (app version update, build version update, embedded files update, etc.)
- creates a PR against release branch

**Steps to test this PR**:
1. Make sure prepare_release.sh, Fastlane/Fastfile and scripts/assets/appstore-release-template are untracked (so when the script changes branches they remain the same). You may need to locally commit Fastfile change on develop for the same reason.
2. Run the script without arguments, with broken arguments, with all the flags and make sure the outcome is correct.


prepare_release.sh
```diff
@@ -0,0 +1,139 @@
...
+
+set -eo pipefail
+
+mute=">/dev/null 2>&1"
+version="$1"
+is_hotfix=false
+
+if ! [[ $common_sh ]]; then
+	cwd="$(dirname "${BASH_SOURCE[0]}")"
+    source "${cwd}/scripts/helpers/common.sh"
```
ayoy: `./prepare_release.sh: line 11: ./scripts/helpers/common.sh: No such file or directory`

Did you forget to add common.sh?

scripts/helpers/common.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+#
+# Output passed arguments to stderr and exit.
+#
+die() {
```
ayoy: Since this file only contains this short function and for now it’s not used anywhere else, I’d suggest that we move it to prepare_release.sh and delete common.sh.

prepare_release.sh
```diff
@@ -0,0 +1,139 @@
...
+
+stash() {
+    printf '%s' "Stashing your changes ... "
+    eval git stash "$mute"
+    echo "✅"
+}
+
+assert_clean_state() {
+    if git show-ref --quiet "refs/heads/${base_branch}"; then
+        echo "💥 Error: Branch ${base_branch} already exists."; exit 1
```
ayoy: `die` function was specifically introduced for such situations. This line can be rewritten as follows:

```bash
die "💥 Error: Branch ${base_branch} already exists.”
```

prepare_release.sh
```diff
@@ -0,0 +1,139 @@
...
+				;;
+			*)
+				print_usage_and_exit "💥 Error: Unknown option '${option}'"
+				;;
+		esac
+	done
+
+	shift $((OPTIND-1))
+
+    [[ $is_hotfix = false ]] && branch_name="release" || branch_name="hotfix"
```
ayoy: Using `true` and `false` in bash is generally frowned upon as it can lead to [weird errors](https://stackoverflow.com/questions/19670061/bash-if-false-returns-true-instead-of-false-why).

I suggest to only set `is_hotfix=1` in case of hotfix (otherwise keep it uninitialized) and then do the following check:
```bash
branch_name="release"
if [[ "$is_hotfix" ]]; then
    branch_name="hotfix"
fi
```

prepare_release.sh
```diff
@@ -0,0 +1,139 @@
...
+    if git show-ref --quiet "refs/heads/${base_branch}"; then
+        echo "💥 Error: Branch ${base_branch} already exists."; exit 1
+    fi
+    if git show-ref --quiet "refs/heads/${changes_branch}"; then
+        echo "💥 Error: Branch ${changes_branch} already exists."; exit 1
+    fi
+}
+
+create_release_branch() {
+    if [ ${is_hotfix} = false ]; then
```
ayoy: `[` is an old POSIX syntax. We should try to only use `[` or `[[` at all times, and since we use `[[` in some places already, we should stick to it.

PS. `[` is actually a built-in function, not a keyword. More reading: https://www.baeldung.com/linux/bash-single-vs-double-brackets

prepare_release.sh
```diff
@@ -0,0 +1,139 @@
...
+    echo "Setting build version ... ✅"
+}
+
+update_embedded_files() {
+    printf '%s' "Updating embedded files ... "
+    eval ./update_embedded.sh "$mute"
+    git add Core/AppTrackerDataSetProvider.swift
+    git add Core/trackerData.json
+    git add Core/AppPrivacyConfigurationDataProvider.swift
+    git add Core/ios-config.json
```
ayoy: The 4 lines above are equivalent to the following:

```swift
git add Core/AppTrackerDataSetProvider.swift \
    Core/trackerData.json \
    Core/AppPrivacyConfigurationDataProvider.swift \
    Core/ios-config.json
```

And they spawn 4 processes, instead of 1. Not a big deal performance wise nowadays, but at least it looks cleaner :)

prepare_release.sh
```diff
@@ -0,0 +1,139 @@
...
+    eval git checkout -b "${changes_branch}" "$mute"
+    echo "✅"
+}
+
+update_marketing_version() {
+    printf '%s' "Setting app version ... "
+    ./set_version.sh "${version}"
+    git add Configuration/Version.xcconfig
+    git add DuckDuckGo/Settings.bundle/Root.plist
+    eval git commit -m \""Update version number\"" "$mute"
```
ayoy: Escaped quotation mark is `\"`, you don’t need `\""` and it only doesn’t break here because the unescaped quotation marks are ignored by eval. It would be great to replace all occurrences with a single `"`.

ayoy (changes requested):
Great to see the release branch automation @jaceklyp! I did some testing and it worked nice! 👏 I’m leaving a bunch of comments that could be addressed in order to tidy up the script, and I also noticed that Gemfile.lock isn’t checked in, while it should be. Happy to chat about any outstanding questions you may have :) Thanks!

ayoy (approved):
Great job @jaceklyp! 🎉

PR #1453, opened 2023-01-18
Autofill fix for crash when deleting a suggested login

amddg44:
Task/Issue URL: https://app.asana.com/0/0/1203705339477104/f
Tech Design URL:
CC: @THISISDINOSAUR 

**Description**:
Fixes the issue where the app crashes when attempting to delete a suggested login

**Steps to test this PR**:
Since this PR allows a combination of rows and sections to be deleted from the table, these are the scenarios to test for this PR:
1. Suggested section only one row; Login is only row in it's alphabetical section -> deleting suggestion deletes both sections
2. Suggested section only one row; Login is one of multiple rows in it's alphabetical section -> deleting suggestion deletes suggested section and single login row from alphabetical section
3. Suggested section has multiple rows; Login is only row in it's alphabetical section -> deleting suggestion deletes single row from suggested section and delete the alphabetical section (to test this scenario I created 2 amazon.com logins, then renamed the Login title for one so that it was grouped under another alphabetical section)
4.  Suggested section has multiple rows; Login is one of multiple rows in it's alphabetical section -> deleting suggestion deletes single row from suggested section and single login row from alphabetical section
5. And for regression testing, try also deleting a login where it is NOT the suggested login, and also a login when there are no suggestions

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPhone 14 Pro
* [ ] iPad

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:


THISISDINOSAUR (commented):
Testing this it looks good to me, but I might be tempted to move as much of the logic as possible to the view model and write unit tests (e.g. with the view model function returning which sections/rows the VC needs to delete)

amddg44:
Sounds good! I'll make that change 👍

amddg44:
@THISISDINOSAUR this is ready for review 🙂

DuckDuckGoTests/MockSecureVault.swift
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    func resetL2Password(oldPassword: Data?, newPassword: Data) throws {}
+
+    func accounts() throws -> [SecureVaultModels.WebsiteAccount] {
+        return storedAccounts
+    }
+
+    func accountsFor(domain: String) throws -> [SecureVaultModels.WebsiteAccount] {
+        return storedAccounts.filter { $0.domain == domain }
```
THISISDINOSAUR, line 42: This might be a pain to update whenever the logic for secure vault changes. Given that it's not currently used in the tests (unless I missed it?), I might be tempted to ever comment something pointing this problem out, or turn it into just a more obvious no op

THISISDINOSAUR: (Although that said I would also love it if we didn't have that logic as tightly coupled to the DB, so it will depend on what you decided to do for the subdomain matching stuff)

THISISDINOSAUR (approved):
LGTM! One minor comment, but I'm thrilled that you've added a mock secure vault and injected it!

PR #1455, opened 2023-01-18
Edit autofill credentials copy buttons and reveal password behaviour changes 

THISISDINOSAUR:
Task/Issue URL: https://app.asana.com/0/0/1203776367328731/f
Tech Design URL:
CC:

**Description**:
Makes a variety of changes to the login details page, including a copy button to the username and password fields and changing the password field reveal behaviour (in both view and edit mode). It also slightly simplifies the code, as there is now only one type of CopyableCell. It possibly also fixes some bugs and incorrect padding I encountered, but I haven't double checked that the components I fixed were previously being used in circumstances where these actually cropped up.

**Steps to test this PR**:
1. Go to the details page for a login and test the copy buttons for both username and password work as expected. Make sure it's not possible to press the button in the same touch as selecting the cell (which brings up the menu at the same time), as that's an issue I encountered whilst doing this.
2. Select the password cell and test the show/hide options behave as expected. Pay particular attention to the text in the menu being correct, since that not updating correctly was an issue at one point.
3. In edit mode, test that the password is revealed when making the field active. It should not hide when it is no longer active
4. Test the x buttons on both the username and password fields work as expected
5. Test that changes are saved correctly when editing
6. In view mode, test that the notes or website url fields padding don't do anything weird, since they originally had too much padding as if they had a button 

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPhone 14 Pro
* [ ] iPad

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:


amddg44:
@THISISDINOSAUR on first pass this is looking good! Quick question, is the font for the password field supposed to be SF Mono or the same as the other font as the other text fields? In Figma it's Mono but I may have missed a conversation on this in Asana 😅 

image

THISISDINOSAUR:
> @THISISDINOSAUR on first pass this is looking good! Quick question, is the font for the password field supposed to be SF Mono or the same as the other font as the other text fields? In Figma it's Mono but I may have missed a conversation on this in Asana 😅
> 
> image

I think it's always been mono, so should still be. I can't remember why though

amddg44:
> I think it's always been mono, so should still be. I can't remember why though

Agree it was always mono before. It's doesn't appear to be mono now in this PR and I wasn't sure if that was intentional or not 🙂   

I think it's a result of deleting `CopyablePasswordCell`, I noticed it is definitely mono when in edit mode, just not in view mode

amddg44 (changes requested):
@THISISDINOSAUR I've tested every scenario listed as well as dark mode, landscape and iPad and everything looks good except for that one small issue with the incorrect font for the password when in view mode

THISISDINOSAUR:
Good spot re the password, should be fixed now!

amddg44 (approved):
LGTM!

PR #1465, opened 2023-01-23
prepare_release workflow improvements

jaceklyp:
Task/Issue URL: https://app.asana.com/0/414235014887631/1203800022093370/f

**Description**:
- support for release notes,
- support for metadata updates,
- validation of metadata

**Steps to test this PR**:
1. Run the flow and see if release notes are properly commited.
2. Run the flow and see if metadata is properly updated (either zipped one or not).

---
###### Internal references:


prepare_release.sh
```diff
@@ -112,8 +116,12 @@ update_build_version() {
 	username="$(git config user.email 2>&1)"
 	bundle exec fastlane increment_build_number_for_version version:"${version}" username:"$username"
 	git add DuckDuckGo.xcodeproj/project.pbxproj
-	eval git commit -m \"Update build number\" "$mute"
-	echo "Setting build version ... ✅"
+	if [[ "$(git diff --cached --exit-code)" ]]; then
```
ayoy: `"$(git diff --cached --exit-code)”` is the standard output of the command, so `--exit-code` is not needed (it’s not taken into account). I’ve also tried with actually using the exit code:

```bash
if git diff —cached --exit-code; then
```

But it didn’t work too well.

prepare_release.sh
```diff
@@ -123,8 +131,49 @@ update_embedded_files() {
...
+	else
+		printf "\nNo changes to embedded files ✅\n"
+	fi
+}
+
+update_metadata() {
+	echo "Updating metadata files ... "
+	local destination="fastlane/metadata/"
+	if [[ "${metadata}" == *.zip ]]; then
+		mkdir temp_metadata
```
ayoy: This will fail if `temp_metadata` exists and is empty. Use `mkdir -p` to not fail in this scenario.

Ideally you would like to `rm -rf temp_metadata` as the first step.

However, since this doesn’t have to happen specifically in `temp_metadata` directory, let me suggest you an alternative approach.

[`mktemp -d`](https://www.tutorialspoint.com/unix_commands/mktemp.htm) creates a temporary directory with a unique name (and also makes sure that it’s only accessible to the current user). Here’s an example usage: https://github.com/duckduckgo/GRDB.swift/blob/SQLCipher/prepare_release.sh#L13-L15
`trap` makes sure that the directory is deleted when script quits, but you can as well delete it manually once you’re done with the metadata.

prepare_release.sh
```diff
@@ -123,8 +131,49 @@ update_embedded_files() {
...
+		unzip "${metadata}" -d "temp_metadata"
+		rsync -a --delete temp_metadata/*/ "${destination}"
```
ayoy: The `*` is problematic here. The archive you’ve given me for testing contains an additional __MACOSX folder that gets synced into fastlane directory.

ayoy (commented):
Nice work @jaceklyp! I’ve had some issues with the test zip file, but perhaps the file itself was broken. Let’s find out tomorrow.

ayoy (approved):
Works great now, thanks a lot @jaceklyp ! 🎉

PR #1472, opened 2023-01-27
Add Maestro

loremattei:
Task/Issue URL: https://app.asana.com/0/42792087274227/1203635285285764/f
Tech Design URL:
CC:

**Description**:
This PR introduces nightly e2e tests using Maestro. The first set of tests are a porting of the e2e set used on Android.

**Steps to test this PR**:
1. Make sure CI is green

To test locally:
1. Install Maestro follow [their instructions](https://maestro.mobile.dev/getting-started/installing-maestro). (make sure to install IDB tool as well)
1. Install Fastlane and run `setup_e2e_tests` lane: `bundle install && bundle exec fastlane setup_e2e_tests`. This will copy the shared modules to subfolders of each test suite and it's required because Maestro YAML parser doesn't seem to understand relative paths. 
2. Build a testable version of the app and install on your simulator. 
3. With the simulator running, use `maestro test` to run the tests. You can run single suites with `maestro test .maestro/<suite folder>` or single tests with `maestro test .maestro/<suite folder>/<test name>` 

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPhone 14 Pro
* [ ] iPad

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:


.maestro/release_tests/bookmarks.yaml
samsymons, line 1: I was starting to test this by running this file locally, but found that as the test starts running at the app gets launched, the simulator is immediately locked and Maestro loses its connection – is there something I'm doing wrong?

loremattei: @samsymons it worked almost out of the box for me, but a few things to check, off the top of my head: 
- How have you installed Maestro?
- How have you installed the app in the Simulator?
- Which build are you using?
- What version of Xcode are you currently using locally?

samsymons: * Through Homebrew
* This was using Xcode to deploy a debug version of the app (which was successfully launched by Maestro, but then encountered the issue above)
* This was the latest develop branch as of Sunday night
* 14.2

I'll try this again with different simulators. FWIW I played with Maestro late last year and didn't have this issue, so I expect there may be some user error on my end this time around.

loremattei: Everything looks good to me. I have installed Maestro using the shell script, but I don't think it makes any difference in this case. 

I figured out I didn't mention the steps to run the tests locally in the description, so I've added them as you'll need to run the Fastlane lane to copy the shared modules. I don't think this is the cause of the problem you're experiencing though, as Maestro returns a parsing error when the files are not in place (so I assume you already figured this out from the GHA configuration).

.github/workflows/end-to-end.yml
```diff
@@ -0,0 +1,51 @@
...
+        submodules: recursive
+
+    - name: Select Xcode
+      run: sudo xcode-select -s /Applications/Xcode_$(<.xcode-version).app/Contents/Developer
+
+    - name: Build for tests
+      run: |
+        set -o pipefail && xcodebuild \
+          -scheme "DuckDuckGo" \
+          -destination "platform=iOS Simulator,name=iPhone 12" \
```
samsymons: What do you think about moving this to `iPhone 14`?

.maestro/ad_click_detection_flow_tests/01_yjs_heuristic_no_ad_domain_param_u3_param_included.yaml
```diff
@@ -0,0 +1,65 @@
...
+appId: com.duckduckgo.mobile.ios
+---
+
+# Set up 
+- clearState
+- launchApp
+- runFlow: 
+    when: 
+      visible: 
+        text: "Let’s Do It!"
```
samsymons, line 12: Are we able to [match elements by accessibility ID](https://maestro.mobile.dev/platform-support/ios-uikit#interacting-with-views-by-accessibility-ids) here, to future-proof ourselves against copy changes?

samsymons (approved):
Besides the XCUITest server issues I've been having locally, these look really nice. I added a couple other comments, and have one last request: can we add some additional documentation somewhere within the README on how to run the tests? (This could possibly be a new Fastlane lane which runs them, and calls `setup_e2e_tests` implicitly.)

Other than that, it feels like we should just get this running nightly on GHA and see how it goes, we'll be able to evaluate the tests over time.

.github/workflows/end-to-end.yml
loremattei: Good point! Updated!

.maestro/ad_click_detection_flow_tests/01_yjs_heuristic_no_ad_domain_param_u3_param_included.yaml
loremattei: Do we have accessibility ID for that element? 
I initially looked for them, but it looks like we don't have a lot of them set in our app. 
But I agree that we should add more of them and use them in UI tests

PR #1478, opened 2023-02-02
Improve iOS build time by replacing file header check with SwiftLint

samsymons:
Task/Issue URL: https://app.asana.com/0/414235014887631/1203880707578250/f
Tech Design URL:
CC: @brindy 

**Description**:

This PR improves the iOS build time by ~13 seconds by replacing the file header check with a SwiftLint rule. Some changes had to happen to make this possible:

1. Some file headers (20 or so) violated the pattern in subtle ways. These have all been changed to make them consistent.
2. There was a directory full of apparently unused source files which was being detected by SwiftLint. These files have all been removed

Three other small cleanup changes happened here:

1. There is an empty file simply named `a` that currently exists in the root of the iOS directory. It has been removed
2. There was a `-copy` variant of an Info.plist file present in the project file, but there was no actual file for it on disk – the Xcode file viewer indicated this in red. It has been removed from the project file
3. SwiftLint had a warning that one of its enabled rules should have been in the `analyzer_rules` section of its config file, so it has been moved

**Before:**
File Header Check Before

**After:**
File Header Check After

**Steps to test this PR**:
1. Build the app locally and verify that you get no file header warnings
1. Check that there is no 10+ second build step related to listing or file headers
1. Create a new file and build the app, verify that the new file doesn't violate the header rule

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16

---
###### Internal references:


.swiftlint.yml
```diff
@@ -49,9 +52,31 @@ type_name:
...
+                    \/\/  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied\.
+                    \/\/  See the License for the specific language governing permissions and
+                    \/\/  limitations under the License\.
+                    \/\/
+
 # General Config
 excluded:
   - Carthage
   - DuckDuckGo/UserText.swift
+  - fastlane/SnapshotHelper.swift
```
samsymons, line 80: Ignoring this as it's a file we didn't write, we want to preserve its original header value.

.swiftlint.yml
```diff
@@ -49,9 +52,31 @@ type_name:
    warning: 80
    error: 100
 
+file_header:
+  required_pattern: |
+                    \/\/
+                    \/\/  .*?\.swift
+                    \/\/  .*?
```
brindy, line 59: No way to make this compare against the actual file name (which is what the script was doing)?

fastlane/SnapshotHelper.swift
```diff
@@ -166,7 +166,6 @@ open class Snapshot: NSObject {
     }
 
     class func pathPrefix() throws -> URL? {
-        let homeDir: URL
```
brindy, line 169: As you said elsewhere we didn't write this file... is it even being built?  I don't think we use it do we?  Maybe we could delete it instead?

brindy (approved):
Nice tidy up! :)  Couple of comments/questions, but it fixes the problem I had anyway.

.swiftlint.yml
samsymons: Whoops, good catch. Fixed now! SwiftLint provides a `SWIFTLINT_CURRENT_FILENAME` value which does this.

fastlane/SnapshotHelper.swift
samsymons: It is, yep - it's used for AtbUITests only though.

.swiftlint.yml
brindy: That fixed it :)

PR #1491, opened 2023-02-09
Sync Setup UI

brindy:
Task/Issue URL: https://app.asana.com/0/72649045549333/1203756800930488/f
Tech Design URL:
CC:

**Description**:

Adds the setup UI for sync behind an internal user flag.

***Notes***:

* The "camera" screens are fixed in dark mode and on iPhone also fixed in portrait mode by design.
* Some of the code is hacky/place holder for future work

***Principles***:

I'm not precious about any of this, but it helped me to maintain some structure.

* Most of the code is in a local Swift Package which is accessed/bootstrapped by the view controller which lives in the app with the required assets. 
* The Swift Package has been carved up into folders by type (Views, ViewModels) with public stuff as near to the top level as possible.  
* The original intention was that the ViewModels be re-usable across platforms, but that will come later.  If it does the ViewModels will move into BSK, then we'll likely drop the "View" folder and move the structure up a level. 
* Each View swift files contains the `public` view plus any required helpers which are marked as private.   
* Any views that are re-usable and not intended to be public have been extracted into their own files and placed in the `Views\internal` folder with default access.
* I've tried to minimise the UI code in the body var and extract things into view build functions where possible/logical.  Please call out anything that looks like it could be clearer.

**Steps to test this PR**:
1. Run as debug or login to the DDG endpoint
2. Ensure Sync item appears in Settings
3. Smoke test the UI
4. If some code looks hacky and it's not obvious it will be fixed later, please call it out (might just need comment to that effect)

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPhone 14 Pro
* [ ] iPad

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:


brindy:
@THISISDINOSAUR - @ayoy is going to pick this up as he's also doing some work on Sync now. :)

DuckDuckGo/Authenticator.swift
```diff
@@ -41,4 +42,12 @@ public class Authenticator {
             }
         }
     }
+
+    public func authenticate(reason: String) async -> Bool {
+        let context = LAContext()
+        let success = try? await context.evaluatePolicy(policy, localizedReason: reason)
+        #warning("Pixel for error?")
```
ayoy: I think we should handle this before merging

LocalPackages/SyncUI/Sources/SyncUI/Views/Internal/QRCodeView.swift
```diff
@@ -0,0 +1,139 @@
...
+            .frame(maxHeight: 200)
+    }
+
+    func generateQRCode(from text: String) -> UIImage {
+        var qrImage = UIImage(systemName: "xmark.circle") ?? UIImage()
+        let data = Data(text.utf8)
+        let filter = CIFilter.qrCodeGenerator()
+        filter.setValue(data, forKey: "inputMessage")
+
+        let transform = CGAffineTransform(scaleX: 20, y: 20)
```
ayoy: Why a scale of 20? :)

Semi-related – I see that designs show the QR code in inverted colors, i.e. white drawing on a black (clear) background. Perhaps this has been a recent update to designs. Anyway, I figured out that you can do it with a [CIFalseColor filter](https://cifilter.io/CIFalseColor/). Check out the implementation for macOS [here](https://github.com/duckduckgo/macos-browser/blob/dominik/sync-setup-ui/DuckDuckGo/Sync/QRCodeView.swift#L53-L57).

LocalPackages/SyncUI/Sources/SyncUI/Views/ScanOrPasteCodeView.swift
```diff
@@ -0,0 +1,251 @@
...
+                Spacer()
+            }
+            .padding(.horizontal, Constants.hPad)
+        }
+    }
+
+    @ViewBuilder
+    func instructions() -> some View {
+
+        Text("Go to Settings > Sync in the **DuckDuckGo App** on a different device and scan supplied code to connect instantly.")
```
ayoy: Just a note that the text emphasis won’t work if we feed it with a localized string from UserText. I don’t have a solution for it and last time I had to do it we agreed with Stephen that emphasis can be skipped :)

LocalPackages/SyncUI/Sources/SyncUI/Views/SyncSettingsScreenView.swift
```diff
@@ -0,0 +1,92 @@
...
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import SwiftUI
+
+// https://www.figma.com/proto/mpFLwzGJFlbsmyD1JkAwt2/Sync-6?page-id=3744%3A50448&node-id=4101%3A70169&viewport=546%2C933%2C0.11&scaling=scale-down
```
ayoy: Pointing this out in case we don’t want to keep Figma links in the code.

ayoy (commented):
@brindy I still need to check a couple of files, but leaving a few comments I’ve had so far. This is looking really nice otherwise! Will also test it on the iPad tomorrow.

brindy:
@ayoy this is ready for another look if you've got time.  If we can merge this then the additional minor changes I need to make for this project will be a much smaller PR :)

LocalPackages/SyncUI/Sources/SyncUI/Views/Internal/QRCodeView.swift
brindy: Was being lazy tbh.  I've basically just copied your code and fixed it for iOS.  Scale isn't so easy to determine so see if you agree with what I've done.

brindy:
@ayoy should be ready for a look now

brindy:
@ayoy I've done what I can for now, I think.  The recovery key screen needs to support landscape, but since it's in flux I don't want to spend any more time on it.  I need to do the PDF generation, but I'm waiting for the design for that too.  I'll just do *something* to get it working, but as a separate PR.

LocalPackages/SyncUI/Sources/SyncUI/Views/ScanOrPasteCodeView.swift
```diff
@@ -0,0 +1,255 @@
...
+
+                VStack(spacing: 0) {
+                    Rectangle() // Also acts as the blur for the camera
+                        .fill(.clear)
+                        .frame(height: g.safeAreaInsets.top)
+                        .regularMaterialBackground()
+
+                    ZStack {
+                        // Background in case fullscreen camera view doesn't work
+                        Rectangle().fill(Color.black)
```
ayoy: This covers the actual camera viewport (also when the camera is fully functional – I can scan a QR code in this state):
IMG_9827

LocalPackages/SyncUI/Sources/SyncUI/Views/Internal/QRCodeScannerView.swift
```diff
@@ -0,0 +1,119 @@
...
+            self.cameraView = cameraView
+            self.session = AVCaptureSession()
+            super.init()
+        }
+
+        func start(_ uiView: UIView) {
+            session.sessionPreset = .high
+
+            guard let backCamera = AVCaptureDevice.default(.builtInWideAngleCamera, for: .video, position: .back),
+                  let input = try? AVCaptureDeviceInput(device: backCamera) else {
```
ayoy, line 61: The camera view doesn’t honor device orientation – I’m testing on an iPad in landscape and am seeing a rotated preview. Feel free to ignore if you’re going to address landscape mode in a separate PR (note: the screenshot is made after I’ve removed the black overlay from the code :) )

IMG_B8C3F4FE6DD0-1

LocalPackages/SyncUI/Sources/SyncUI/Views/DeviceConnectedView.swift
```diff
@@ -0,0 +1,87 @@
...
+                withAnimation {
+                    self.showRecoveryPDF = true
+                }
+            } label: {
+                Text(UserText.nextButtonTitle)
+            }
+            .buttonStyle(PrimaryButtonStyle())
+        }
+        .padding(.top, 56)
+        .padding(.horizontal)
```
ayoy, line 74: Add also `.padding(.bottom)` to fix iPad sheet. It should also work fine for iPhone, where currently the bottom edge of the button touches safe area.

IMG_6FFEE8E8C20B-1

LocalPackages/SyncUI/Sources/SyncUI/Views/Internal/QRCodeView.swift
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    func generateQRCode(from text: String, scale: CGFloat) -> UIImage {
+        var qrImage = UIImage(systemName: "xmark.circle") ?? UIImage()
+        let data = Data(text.utf8)
+        let qrCodeFilter: CIFilter = CIFilter.init(name: "CIQRCodeGenerator")!
+        qrCodeFilter.setValue(data, forKey: "inputMessage")
+        qrCodeFilter.setValue("H", forKey: "inputCorrectionLevel")
+
+        let transform = CGAffineTransform(scaleX: scale, y: scale)
```
ayoy, line 65: Wouldn’t `UIScreen.main.scale` work here?

LocalPackages/SyncUI/Sources/SyncUI/Views/Internal/QRCodeView.swift
```diff
@@ -0,0 +1,86 @@
...
+        if let image = context.createCGImage(coloredImage, from: outputImage.extent) {
+            qrImage = UIImage(cgImage: image)
+        }
+
+        return qrImage
+    }
+
+}
+
+private extension CIFilter {
```
ayoy: This is not needed anymore.

ayoy (changes requested):
Thanks for the updates @brindy! I’m adding a couple more comments after taking a look at the new build. The main blocker seems to be the black background showing at all times on top of camera view. And I’ve added a suggestion to use UIScreen.main.scale for the QR code.

LocalPackages/SyncUI/Sources/SyncUI/Views/Internal/QRCodeView.swift
brindy: It returns a value of like `3`, which is pretty useless, you just end up with a fuzzy qrcode.  The example I saw used 20 so that when it scaled up it remained crisp.

brindy:
Thanks @ayoy, that should be all your points addressed, but let's talk about the qrcode image scale one directly (I don't think it matters too much just now)

ayoy (approved):
LGTM! The QR code looks great now. Thanks for taking time to address my remarks @brindy!

brindy:
thanks for the reviews :)

PR #1500, opened 2023-02-15
Fix for 'Cookies managed' notification to properly fade out

miasma13:
Task/Issue URL: https://app.asana.com/0/414709148257752/1203967747887450/f

**Description**:
Cookies managed notification appears with a fade-in but is not fading-out after completion. The reason to this behaviour was that on completion it was removed from the view hierarchy too early.

**Steps to test this PR**:
1. Enable "Manage Cookie Pop-ups" in app's settings.
2. Clear cookies.
3. Navigate to as website showing a cookie consent prompt
4. When the pop up is managed the notification should properly animate in the address bar.

---
###### Internal references:

### DIFF
--- a/DuckDuckGo/OmniBarNotificationAnimator.swift
+++ b/DuckDuckGo/OmniBarNotificationAnimator.swift
@@ -53,13 +53,17 @@ final class OmniBarNotificationAnimator: NSObject {
                     omniBar.textField.alpha = 1
                     omniBar.privacyInfoContainer.alpha = 1
                 }
+                
+                DispatchQueue.main.asyncAfter(deadline: .now() + 2 * fadeDuration) {
+                    omniBar.notificationContainer.removePreviousNotification()
+                }
             }
         }
     }
     
     func cancelAnimations(in omniBar: OmniBar) {
         omniBar.privacyInfoContainer.alpha = 0
-        omniBar.notificationContainer.stopAnimation()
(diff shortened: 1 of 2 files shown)


DuckDuckGo/OmniBarNotificationAnimator.swift
```diff
@@ -53,13 +53,17 @@ final class OmniBarNotificationAnimator: NSObject {
                     omniBar.textField.alpha = 1
                     omniBar.privacyInfoContainer.alpha = 1
                 }
+                
+                DispatchQueue.main.asyncAfter(deadline: .now() + 2 * fadeDuration) {
```
Bunn, line 58: @miasma13 what do you think about naming that “2” ?
It’s probably just an offset for the animation, but I believe a name could make it clear why it exists?

Bunn (approved):
LGTM, animations are smoooooth 💅 
Just one small comment that I’ll leave to you to decide if it’s worth the effort.

DuckDuckGo/OmniBarNotificationAnimator.swift
miasma13: I think it is a fair idea. Yeah they are sequenced one after another:
- first one fades out with fadeDuration
- then second fades in with fadeDuration (and single fadeDuration delay)
- animation starts with double fadeDuration delay

PR #1521, opened 2023-03-02
Enable username&password passed in URL

mallexxx:
Task/Issue URL: https://app.asana.com/0/414235014887631/1204080079383830/f
Tech Design URL: https://app.asana.com/0/481882893211075/1204065038916791/f
Security Triage URL: https://app.asana.com/0/1199892415909552/1204067721102886/f
BSK PR: https://github.com/duckduckgo/BrowserServicesKit/pull/245
CC:

**Description**:
- Adds support to pass basic auth username/password in URL

**Steps to test this PR**:
URLs for testing: 
- https://authenticationtest.com/HTTPAuth/ (user:pass)
- http://httpbin.org/basic-auth/foo/bar (foo:bar)

1. Validate login/password passed in URL is used for basic authentication and is saved per session
2. Validate login/password passed in URL is not displayed in the UI or saved in browsing history
3. Validate when clicking links containing username/password (user:pass@domain.com), the credentials aren‘t displayed in UI but are used during basic auth
4. Validate if invalid credentials are provided then auth dialog is displayed


DuckDuckGo/TabViewController.swift
```diff
@@ -242,7 +243,6 @@ class TabViewController: UIViewController {
         }
     }
 
-    private var rulesCompiledCondition: RunLoop.ResumeCondition? = RunLoop.ResumeCondition()
```
brindy, line 246: What does this have to do with this PR?

brindy (changes requested):
This doesn't appear to work.  I created the following URL, pasted it in and hit enter and it loaded the SERP.

https://user:pass@authenticationtest.com/HTTPAuth/

But I created a bookmark with this URL, used the fire button, then opened the link with the bookmark and it worked.

Also, what are the changes about the `rulesCompiledCondition` about - just code clean up?

DuckDuckGo/TabViewController.swift
mallexxx: I was adding urlProvidedBasicAuthCredential property and noticed this rulesCompiledCondition is not used anywhere, I was hesitating to add an extra var to the class but having this removed sounds fair

brindy: Fair enough :)

mallexxx:
@brindy I‘m not sure there‘s a double check for URL validity first in OmniBar.swift:401 and then basically the same procedure (BSK URLExtension: `URL.isValid`) in `webUrl(from text: String)`, so I‘ve just removed the check for `user == nil`, seems working now

brindy (commented):
@mallexxx after using the fire button I am still logged in when testing with https://user:pass@authenticationtest.com/HTTPAuth/ - is that expected?

[Edit: I tested this on macOS and it works as expected - after the fire button you are logged out]

mallexxx:
I guess that is expected, most probably credential cash is not cleaned in iOS

brindy (approved):
Looks like we're caching credentials in prod too, I'll create a separate task for that.  LGTM!

PR #1529, opened 2023-03-09
Tidy up AppURLs

jaceklyp:
Task/Issue URL: https://app.asana.com/0/414709148257752/1184981768664663/f
CC: @brindy 

**Description**:
Tidy up AppURLs

**Steps to test this PR**:
1. Run tests.
2. Perform a couple of smoke tests.

---
###### Internal references:


bwaresiak:
@jaceklyp As we agreed, there may be a chance to tidy up the Statistics Store dependency and maybe remove the need to make this a singleton.

jaceklyp:
@bwaresiak doing it as we speak

Core/AppURLs.swift
```diff
@@ -0,0 +1,128 @@
...
+    static let surrogates = URL(string: "\(staticBase)/surrogates.txt")!
+    static let privacyConfig = URL(string: "\(staticBase)/trackerblocking/config/v2/ios-config.json")!
+    static let trackerDataSet = URL(string: "\(staticBase)/trackerblocking/v3/apple-tds.json")!
+    static let bloomFilter = URL(string: "\(staticBase)/https/https-mobile-v2-bloom.bin")!
+    static let bloomFilterSpec = URL(string: "\(staticBase)/https/https-mobile-v2-bloom-spec.json")!
+    static let bloomFilterExcludedDomains = URL(string: "\(staticBase)/https/https-mobile-v2-false-positives.json")!
+
+    private static var devMode: String { isDebugBuild ? "?test=1" : "" }
+    static let atb = URL(string: "\(base)/atb.js\(devMode)")!
+    static let exti = URL(string: "\(base)/exti/\(devMode)")!
```
bwaresiak, line 43: Are these supposed to be used without ATB params? If not, maybe we should hid these?

Also I'm not sure I like all of these new static fields being on URL class, feels like we are "polluting" the Apple namespace with our stuff, but not strong feeling about that.

jaceklyp: Similar implementation can be found on macos, just checked it.
My reasoning is as follows, for me this is the same use case as with "Colors" - so we used to have Color.destructive, here we have URL.privacyConfig. Feels like extension on URL is a natural way to access any URLs we might ever have.

For atb/exti, exti is being used alone, atb not. However, I cannot make it private as it is used inside StatisticsDependentURL. Solution would be to pack it all in the same file and make it fileprivate, but I feel like a future local module will solve the issue in a cleaner way.

brindy:
Still looking at this but something that jumps out is the factory method names.  The [API Design Guidelines](https://www.swift.org/documentation/api-design-guidelines/) says:

> The first argument to initializer and [factory methods](https://en.wikipedia.org/wiki/Factory_method_pattern) calls should not form a phrase starting with the base name 

And since these aren't a value preserving operation the first parameter should not be omitted. 

So for instance:

```static func make(forPixelWithName pixelName: String, formFactor: String? = nil, includeATB: Bool = true) -> URL```

should probably be:

```static func makePixelURL(name: String, formFactor: String? = nil, includeATB: Bool = true) -> URL```

I think the reasoning here is that factory methods are effectively constructors and you wouldn't do that there either. 

@bwaresiak any thoughts on that?  We've had similar conversations in the past :)

File: Core/AppURLs+StatisticsDependent.swift [hunk 2]
```diff
@@ -0,0 +1,148 @@
...
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+import BrowserServicesKit
+
+public extension URL {
```
brindy: Still looking at this but something that jumps out is the factory method names.  The [API Design Guidelines](https://www.swift.org/documentation/api-design-guidelines/) says:

> The first argument to initializer and [factory methods](https://en.wikipedia.org/wiki/Factory_method_pattern) calls should not form a phrase starting with the base name 

And since these aren't a value preserving operation the first parameter should not be omitted. 

So for instance:

```static func make(forPixelWithName pixelName: String, formFactor: String? = nil, includeATB: Bool = true) -> URL```

should probably be:

```static func makePixelURL(name: String, formFactor: String? = nil, includeATB: Bool = true) -> URL```

I think the reasoning here is that factory methods are effectively constructors and you wouldn't do that there either. 

@bwaresiak any thoughts on that?  We've had similar conversations in the past :) 

(Note, moved to a comment from main discussion)

File: DuckDuckGo.xcodeproj/project.pbxproj
Comment (brindy) [inline DuckDuckGo.xcodeproj/project.pbxproj:1]: I noticed that the app wouldn't run until I recursed submodules and it pulled in the ios-js-support submodule, but that has gone from develop now (replaced with a Swift Package) so this probably needs to be updated from develop?

Comment (jaceklyp) [reply DuckDuckGo.xcodeproj/project.pbxproj:1]: Just merged, it should be fine now.

File: Core/AppURLs+StatisticsDependent.swift (hunk 2 above)
Comment (jaceklyp) [reply Core/AppURLs+StatisticsDependent.swift]: changed, looks much better!

File: Core/AppURLs+StatisticsDependent.swift [hunk 3]
```diff
@@ -0,0 +1,148 @@
...
+
+    private static let defaultStatisticsDependentURL = StatisticsDependentURL()
+
+    static func makeSearchURL(text: String) -> URL? { defaultStatisticsDependentURL.makeSearchURL(text: text) }
+    static func makeSearchURL(query: String, queryContext: URL? = nil) -> URL? {
+        defaultStatisticsDependentURL.makeSearchURL(query: query, queryContext: queryContext)
+    }
+    static func applyingStatsParams(for url: URL) -> URL { defaultStatisticsDependentURL.applyingStatsParams(for: url) }
```
brindy: This is slightly confusing, it looks like a factory method here, but I know how it's used.  Maybe this should be an instance func instead?

Core/AppURLs+StatisticsDependent.swift
```diff
@@ -0,0 +1,148 @@
...
+    static var searchAtb: URL? { defaultStatisticsDependentURL.searchAtb }
+    static var appAtb: URL? { defaultStatisticsDependentURL.appAtb }
+    static func hasCorrectMobileStatsParams(url: URL) -> Bool { defaultStatisticsDependentURL.hasCorrectMobileStatsParams(url: url) }
+    static func makePixelURL(pixelName: String, formFactor: String? = nil, includeATB: Bool = true) -> URL {
+        defaultStatisticsDependentURL.makePixelURL(pixelName: pixelName, formFactor: formFactor, includeATB: includeATB)
+    }
+
+}
+
+public final class StatisticsDependentURL {
```
brindy: This isn't a `URL`, really it's a `URLFactory`

Core/AppURLs+StatisticsDependent.swift
```diff
@@ -0,0 +1,148 @@
...
+     */
+    private func makeSearchURL<C: Collection>(text: String, additionalParameters: C) -> URL
+    where C.Element == (key: String, value: String) {
+        let searchURL = URL.ddg
+            .appendingParameter(name: URL.Param.search, value: text)
+            .appendingParameters(additionalParameters)
+        return applyingStatsParams(for: searchURL)
+    }
+
+    func applyingStatsParams(for url: URL) -> URL {
```
brindy: super-nit: I think I would prefer using `to` rather than `for` if trying to be grammatical, but I think you can simply remove the parameter name altogether since this is really internal to this file

brindy (approved):
I ran this over a few scenarios:

* checked ATB
* sent some feedback
* did a search
* Generally used the app for a bit

A few minor comments about naming, but I don't feel strongly, so up to you. :) 

Otherwise, LGTM!

Core/AppURLs+StatisticsDependent.swift
jaceklyp: makes sense! I changed hasCorrectMobileStatsParams too

jaceklyp:
I decided to merge AppURLs and AppURLs+StatisticsDependent into single file to leverage fileprivate accessor (there's no plan to make it a separate local module for now) not to expose Params and ParamsValue to the rest of the app. Now URL should only contain actual URLs that are used across the app and some instance helpers to check if the url matches some requirements. Hopefully it mitigates some of @bwaresiak concerns.

PR #1555, opened 2023-03-21
Feature flagging with autofill subfeatures

graeme:
Task/Issue URL: https://app.asana.com/0/72649045549333/1203676877763836
Tech Design URL: https://app.asana.com/0/481882893211075/1203938569856702
https://app.asana.com/0/481882893211075/1203984214904257
CC:

Description:

This PR and its corresponding [BSK PR](https://github.com/duckduckgo/BrowserServicesKit/pull/270):
- adds subfeatures support to the Privacy Config component
- moves the FeatureFlagger from iOS into BSK and updates it for the requirements gathered during this project
- moves the InternalUserDecider from iOS into BSK

To understand why I ended up at the solution I did, it's worth pointing out some requirements:
- We should be able to remotely toggle features using the privacy config
- Remote toggling shouldn't necessarily be possible while the feature is in development, only when we're close to release. Up until this point, we should stick with having them enabled internally alone by authenticating as an internal user.
- The current state of a given feature (particularly during development) may well be different in iOS vs macOS and we need a way to express that.
- Some features (autofill included) may be developed (and in the future disabled) only partially, introducing the need for nested features ("subfeatures") in the Privacy Configuration.

Additionally, this stops in-development features being enabled by default in dev builds as discussed here: https://app.asana.com/0/1200194497630846/1203752002572023/f

**Steps to test this PR**:
1. Check that any features configured using the privacy-config still work as expected (particularly autofill).
2. Check feature flags (autofill, debug menu, sync)

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [x] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [x] iPhone 14 Pro
* [ ] iPad

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [x] iOS 16

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:


Core/FeatureFlag.swift
```diff
@@ -0,0 +1,45 @@
...
+        switch self {
+        case .debugMenu, .sync, .credentialsAutofill, .credentialsSaving, .inlineIconCredentials, .accessCredentialManagement:
+            return .internalOnly
+        }
+    }
+}
+
+extension FeatureFlagger {
+    public func isFeatureOn(_ featureFlag: FeatureFlag) -> Bool {
+        isFeatureOn(forProvider: featureFlag)
```
samsymons, line 43: One important feature we had here before was a check for `isDebugBuild`, to avoid having to reauthenticate any time you run a clean build. Is it possible to restore that?

Core/FeatureFlag.swift
```diff
@@ -0,0 +1,45 @@
...
+import Foundation
+import BrowserServicesKit
+
+public enum FeatureFlag: String {
+    case debugMenu
+    case sync
+    case credentialsAutofill
+    case credentialsSaving
+    case inlineIconCredentials
+    case accessCredentialManagement
```
samsymons: What do you think about naming these such that it's clear that they're all autofill related? It seems obvious now, but once we have a bunch more of these I could see it becoming messy. Breaking the enum up with newlines and comments indicating each major feature would probably be fine.

Alternatively I could see separate enums for each big feature, like `AutofillFeatureFlag`, though I think that would be overkill right now.

Core/FeatureFlag.swift
```diff
@@ -0,0 +1,45 @@
...
+}
+
+extension FeatureFlag: FeatureFlagSourceProviding {
+    public var source: FeatureFlagSource {
+        switch self {
+        case .debugMenu, .sync, .credentialsAutofill, .credentialsSaving, .inlineIconCredentials, .accessCredentialManagement:
+            return .internalOnly
```
samsymons, line 36: As we now have the Autofill features merged to the Remote Config as internal-only, does it make sense to use `remoteDevelopment` here for the applicable subfeatures?

DuckDuckGo/TabViewController.swift
```diff
@@ -2345,7 +2345,7 @@ extension TabViewController: SecureVaultManagerDelegate {
     }
     
     func secureVaultManagerIsEnabledStatus(_: SecureVaultManager) -> Bool {
-        let isEnabled = featureFlagger.isFeatureOn(.autofill)
+        let isEnabled = featureFlagger.isFeatureOn(.credentialsSaving)
```
samsymons: This function `secureVaultManagerIsEnabledStatus` generally decides when to provide autofillable values to the webview, should the `credentialsAutofill` subfeature be used here instead?

DuckDuckGo/TabViewController.swift
```diff
@@ -2370,7 +2370,7 @@ extension TabViewController: SecureVaultManagerDelegate {
                             withTrigger trigger: AutofillUserScript.GetTriggerType,
                             completionHandler: @escaping (SecureVaultModels.WebsiteAccount?) -> Void) {
   
-        if !isAutofillEnabled {
+        if !isAutofillEnabled, featureFlagger.isFeatureOn(.credentialsAutofill) {
```
samsymons: The naming of `isAutofillEnabled` threw me off here. To me, it sounds like it should be itself checking for a feature flag, but really it's just detecting whether the user has enabled Autofill in the settings. I don't consider it a blocker, but I'd be interested to see if we can make the naming more clear.

Core/FeatureFlag.swift
samsymons: Of course: https://github.com/duckduckgo/iOS/pull/1565 🙂

Core/FeatureFlag.swift
graeme: Actually we bundled this task into this project: https://app.asana.com/0/1200194497630846/1203752002572023/f

Core/FeatureFlag.swift
graeme: Exactly 😄  Until the `internal` state handling is added to the Privacy Config, that would result in autofill being either off or on for all users. So makes sense to promote it in that PR.

samsymons (approved):
Looks good, nice work!

Core/FeatureFlagger.swift
```diff
@@ -1,96 +0,0 @@
...
-            return true
-        }
-        return didVerifyInternalUser
-    }
-    
-    private static let internalUserVerificationURLHost = "use-login.duckduckgo.com"
-    
-    @discardableResult
-    public func markUserAsInternalIfNeeded(forUrl url: URL?, response: HTTPURLResponse?) -> Bool {
-        if isInternalUser { // If we're already an internal user, we don't need to do anything
```
graeme, line 74: This means that authenticating while in debug does not store the internal user state. Therefore updating debug builds will require logging in again (pending decision as to whether we actually just keep debug builds as `isInternalUser = true` by default)

PR #1617, opened 2023-04-04
Autofill fix for edge case crash when tapping on search bar

amddg44:
Task/Issue URL: https://app.asana.com/0/414709148257752/1204144513391838/f
Tech Design URL:
CC:

**Description**:
Fixes an edge case where tapping on the search bar caused a crashed. 

**Steps to test this PR**:
The issue is that when tapping on the search bar, the viewModel changes the number of sections (since the first section `Save and Autofill Logins switch` is removed), but the tableView data was not reloaded. So when table scrolls up to focus on the search bar, `cellForRowAtIndexPath` is called to render a new row and since the tableView data and the viewModel data is different, `items[indexPath.row]` will crash if there is no item at that position 
1. Test on device (can't reproduce on simulator)
2. Import a very large number of logins (1k+)

3. Comment out the [one line code change](https://github.com/duckduckgo/iOS/pull/1617/files#diff-2bb09f3d45ff594d6804e6abbf6cebba30308d4e31b9c82544a732935f175382R638) (line 680), run up the app and hopefully you should experience the crash (it may take a few attempts of scrolling to different sections of the logins screen and tapping search again)
4. Add back in the code change
5. Test again and there should be no crash.
6. Perform some smoke tests (search, tap search results, cancel search etc) confirming there is no change in behaviour

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPhone 14 Pro
* [ ] iPad

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [x] iOS 16

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:


Bunn:
@amddg44 I'm running the steps with the reloadData removed and I can't reproduce the crash =/ 

https://user-images.githubusercontent.com/782316/230135845-f4b5f1fb-26d0-4442-a373-a6664aa0ff90.MP4

Can you verify if I'm missing something on what I'm doing or the data I'm using as example?
Running on iPhone 6S, iOS 15.7

amddg44:
Thanks for testing @Bunn. I was worried it would be hard to reproduce the crash. I'm able to reproduce consistently on my iPhone 14 Pro, iOS 16.3.1 but it could be dependent on exact screen size, font size etc. Leave it with me to see if I can reproduce consistently on other devices and I'll ping you when I have up update. thanks again!

bwaresiak (approved):
I did manage to reproduce this crash on the device (iPh 13 mini) with 1k+ logins and this fixes it, good job!

PR #1638, opened 2023-04-13
align logging with macOS

mallexxx:
Task/Issue URL: https://app.asana.com/0/1177771139624306/1204359032555835/f
BSK PR: https://github.com/duckduckgo/BrowserServicesKit/pull/312
macOS PR: https://github.com/duckduckgo/macos-browser/pull/1120

**Description**:
- not actual for iOS: fixed the navigation flow when fragment Navigation could end-up without a NavigationAction assigned
- introduced os_log wrapper in BSK to accept message arguments as autoclosures i.e. computed only if valid log and log level are passed
- Logging.swift (logs enabling/disabling) aligned with macOS recent update

**Steps to test this PR**:
1. Validate `os_log` calls across the codebase don‘t need changing (only need to replace `replace import` os with `import Common`), arguments passed as %s and %d parameters are printed correctly, to the right log and with correct log level. Validate interpolated string passed to `os_log` works (`os_log(message: "interpolated \(string)")`)
2. Validate logging works and can be enabled/disabled in Logging.swift using `enabledCategories`
3. Validate `import os.log` (import os, OSLog) is discouraged by swiftlint in favour of `import Common`


Core/Logging.swift
```diff
@@ -18,62 +18,48 @@
...
+    @OSLogWrapper(.contentBlockingLog) static var contentBlockingLog
+    @OSLogWrapper(.adAttributionLog) static var adAttributionLog
+    @OSLogWrapper(.lifecycleLog) static var lifecycleLog
+    @OSLogWrapper(.autoconsentLog) static var autoconsentLog
+    @OSLogWrapper(.configurationLog) static var configurationLog
+
+    // !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+    // To activate Logging Categories add categories here:
+    static var enabledCategories: Set<AppCategories> = [
+        .generalLog,
```
tomasstrba, line 46: Categories should be disabled by default.

tomasstrba (changes requested):
Just a small configuration problem. Otherwise everything works as expected 👍

Core/Logging.swift
```diff
@@ -18,62 +18,48 @@
...
+        .lifecycleLog,
+        .configurationLog
+    ]
 
 }
 
-struct Logging {
+public extension OSLog.OSLogWrapper {
 
-    fileprivate static let generalLoggingEnabled = true
```
mallexxx, line 55: @tomasstrba, these logs were originally enabled in the main branch

brindy (approved):
LGTM

Core/Logging.swift
tomasstrba: yeah, makes sense! Thanks! :)

PR #1640, opened 2023-04-13
AppTP UI phase 2

SlayterDev:
Task/Issue URL: https://app.asana.com/0/72649045549333/1204225219992687/f
Tech Design URL: https://app.asana.com/0/481882893211075/1204275929056721/f
CC: @samsymons 

**Description**:
This is phase 2 of the AppTP UI adding the ability to manage trackers, view blocked/allowed trackers, and updates to the breakage form styling.

**Steps to test this PR**:
1. Verify alignment with design linked in Asana

**Allowlist**
1. Go to AppTP > Manage Trackers
2. Turn off Google trackers related to ads (including doubleclick)
3. With the VPN on go to Safari > search on google > click a sponsored link. The link should load correctly
4. Re-enable the allowed trackers from step 2. Repeat step 3. The page should not load.

**Tracker Details**
1. Tap a tracker in the activity view
2. Disable it from the details screen
3. Trigger the tracker again (Ebay is an easy choice for this)
4. The tracker should show in the activity feed as allowed (sort by date)

**Breakage Form**
1. Verify new design
2. All breakage categories should show the comment field
3. Typing in the comment field should make the placeholder disappear.

**Copy Testing**:

**Orientation Testing**:

* [x] Portrait
* [x] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [x] iPhone 14 Pro
* [ ] iPad

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [x] iOS 16

**Theme Testing**:

* [x] Light theme
* [x] Dark theme

---
###### Internal references:

---
  - https://app.asana.com/0/0/1204377010739656


DuckDuckGo/AppTPManageTrackersViewModel.swift
```diff
@@ -0,0 +1,70 @@
...
+import Foundation
+import Core
+
+struct ManagedTrackerRepresentable: Hashable {
+    let domain: String
+    let trackerOwner: String
+    let blocking: Bool
+}
+
+class AppTPManageTrackersViewModel: ObservableObject {
```
SlayterDev, line 29: I want to unit test this view model but since this PR is already huge I will do a follow up and refactor `TrackerDataParser` to be testable as well.

DuckDuckGo/AppTPManageTrackersViewModel.swift
```diff
@@ -0,0 +1,70 @@
...
+        trackerList.sort(by: { ($0.trackerOwner, $0.domain) < ($1.trackerOwner, $1.domain) })
+    }
+    
+    public func changeState(for trackerDomain: String, blocking: Bool) {
+        if !blocking {
+            allowlist.allow(domain: trackerDomain)
+        } else {
+            allowlist.remove(domain: trackerDomain)
+        }
+        buildTrackerList()
```
SlayterDev, line 63: TODO: This may be a heavy operation. I may be able to just update one entry.

DuckDuckGo/AppDelegate.swift
```diff
@@ -319,6 +319,22 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+            
+            // Make sure we don't fire this pixel multiple times a day
+            let dayStart = Calendar.current.startOfDay(for: date)
+            let fireDate = UserDefaults.standard.object(forKey: key) as? Date
+            if fireDate == nil || fireDate! < dayStart,
+               fwm.status() == .connected {
+                Pixel.fire(pixel: .appTPActiveUser)
+                UserDefaults.standard.set(date, forKey: key)
+            }
+        }
```
SlayterDev: I think this block was accidentally re-added in the merge. Will deal with this on Monday.

Core/AppTrackingProtectionListViewModel.swift
```diff
@@ -91,6 +97,19 @@ public class AppTrackingProtectionListViewModel: NSObject, ObservableObject, NSF
...
     }
+    
+    public func format(timestamp: Date) -> String {
+        let date = Date()
+        let timestampInterval = timestamp.timeIntervalSinceReferenceDate
+        let dateInterval = date.timeIntervalSinceReferenceDate
+        if fabs(dateInterval - timestampInterval) < 1 {
+            // Can't access UserText from Core. To prevent handling the localized string "in 0 seconds"
+            // return "0" here and replace it with UserText on the view side.
+            return "0"
```
samsymons: What do you think about having this function return an optional, and in this case returning `nil`? Right now the value of `"0"` feels a bit magic and requires the caller to know about an implementation detail of this function - if it ever changed, we'd end up with a bug on the client side but have no way for the compiler to tell us about it.

DuckDuckGo/AppDelegate.swift
samsymons: Is this duplicated with the `fireAppTPActiveUserPixel` function below?

Core/AppTrackingProtectionListViewModel.swift
SlayterDev: Yeah that makes way more sense 🤦‍♂️

DuckDuckGo/AppDelegate.swift
SlayterDev: Mentioned in another comment that was posted late. This was a side-affect of the merge that I missed. Will remove.

DuckDuckGo/en.lproj/Localizable.strings
```diff
@@ -136,15 +136,84 @@
...
+"appTP.cell.detail" = "Block app trackers on your device";
+
+/* String indicating AppTP is disabled when viewed from the settings screen */
+"appTP.cell.disabled" = "Disabled";
+
 /* Info string informing the user what App Tracking Protection does. */
 "appTP.empty.disabled.info" = "Enable App Tracking Protection so we can block pesky trackers in other apps.";
 
 /* Info string informing the user we're looking for trackers in other apps. */
 "appTP.empty.enabled.info" = "We're looking for trackers in your apps...";
```
samsymons: Just noticed that this isn't using the correct apostrophe type:

```suggestion
/* Info string informing the user we're looking for trackers in other apps. */
"appTP.empty.enabled.info" = "We’re looking for trackers in your apps...";
```

DuckDuckGo/AppTPActivityView.swift
```diff
@@ -25,6 +25,8 @@ struct AppTPActivityView: View {
     @ObservedObject var feedbackModel: AppTrackingProtectionFeedbackModel
     @ObservedObject var toggleViewModel = AppTPToggleViewModel()
     
+    // We only want to show "Manage Trackers" and "Report an issue" if the user has enabled AppTP at least once
+    @AppStorage("appTPUsed") var appTPUsed: Bool = false
```
samsymons: Are we able to move this into the view model and use `@UserDefaultsWrapper`? Generally we prefer to keep views as dumb as possible, and keep logic like this in the models to facilitate unit testing.

DuckDuckGoTests/AppTrackingProtectionAllowlistModelTests.swift
```diff
@@ -0,0 +1,80 @@
...
+final class AppTrackingProtectionAllowlistModelTests: XCTestCase {
+
+    func testAllowlistInitSucceeds() {
+        let allowlist = AppTrackingProtectionAllowlistModel()
+        XCTAssertNotNil(allowlist, "Allowlist not initialized")
+    }
+    
+    func testAllowlistReturnsTrueForAllowedDomains() {
+        var allowlist = AppTrackingProtectionAllowlistModel()
+        allowlist.allow(domain: "example.com")
```
samsymons, line 51: I think this will modify the real allowlist within the simulator, right? It would be great if we could provide a dependency-injected file path to write to, which would also make it easier to clear out at the end of each test run and validate that each test is beginning in a clean state.

DuckDuckGo/TabViewControllerBrowsingMenuExtension.swift
```diff
@@ -23,6 +23,7 @@ import BrowserServicesKit
 import Bookmarks
 import simd
 import WidgetKit
+import Common
```
samsymons, line 26: Do we need this import?

SlayterDev: Yeah I had a compiler error if I didn't change it. Maybe it got mixed up though.

samsymons: Nah that's okay, happy to leave this in – I've had the same thing happen before. 🙂

Core/AppTrackingProtectionAllowlistModel.swift
```diff
@@ -0,0 +1,106 @@
...
+    private let filename: String
+    
+    lazy private var allowlistUrl: URL? = {
+        let groupContainerUrl = FileManager.default.containerURL(forSecurityApplicationGroupIdentifier: Constants.groupID)
+        return groupContainerUrl?.appendingPathComponent(self.filename, conformingTo: .text)
+    }()
+    
+    var allowedDomains: Set<String>
+    
+    public init(filename: String = Constants.fileName) {
```
samsymons, line 37: I'd be inclined to make this take the full URL instead of just the filename, but to have a convenience initializer which uses the default name + group container URL, since it would let us provide any URL in the test suite (such as temp directory). What do you think?

samsymons: Actually, I think this is fine in the end, the test suite is already solving the problem I was concerned about

samsymons (approved):
Looks great!

PR #1676, opened 2023-04-25
call remove device correctly

brindy:


brindy:
Testing local automation

PR #1646, opened 2023-04-14
Update icons across the app to the new design system

THISISDINOSAUR:
Task/Issue URL: https://app.asana.com/0/0/1204325589959392/f
Tech Design URL:
CC:

**Description**:
As titled. This is fairly straightforward but impacts many many pages, so requires testing the entire app for breakage.
Note that this is built off, and merging into elle/deprecate-colors (https://github.com/duckduckgo/iOS/pull/1616), rather than develop.

**Steps to test this PR**:
1. This is fairly straightforward but impacts many many pages, so requires testing the entire app for breakage. Particular areas to call out are:
- Autofill details
- The broken site flow
- The URL bar suggestions
- The main browsing view
- The browsing menu
- The tab switcher

**Copy Testing**:

* [ ] Use of correct apostrophes in new copy, ie `’` rather than `'`

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPhone 14 Pro
* [ ] iPad

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:
[Software Engineering Expectations](https://app.asana.com/0/59792373528535/199064865822552)
[Technical Design Template](https://app.asana.com/0/59792373528535/184709971311943)

PR #1689, opened 2023-05-02
Revert "Update icons across the app to the new design system"

THISISDINOSAUR:
Reverts duckduckgo/iOS#1646


amddg44 (approved):
LGTM + tested on iPhone & iPad in both light and dark mode. 

@afterxleep smoke tested on iPad and iPhone also and didn't notice any issues either (thank you Daniel!)

PR #1725, opened 2023-05-16
Disable the packet tunnel provider

samsymons:
Task/Issue URL: https://app.asana.com/0/414235014887631/1204593255118994/f
Tech Design URL:
CC:

**Description**:

This PR removes the packet tunnel provider and disables related code.

The feature is controlled behind a conditional compilation flag, so to turn everything back on you just need to set that value in the release build and add the packet tunnel provider back to the main app target.

**Steps to test this PR**:
1. Check that the app builds and runs locally - it's okay if AppTP shows up in settings, it's still turned on for debug builds so that we can guarantee that it doesn't break without us knowing about it
2. Check that you can archive a build, to validate that the release process isn't broken
3. Do a search for `import NetworkExtension` and make sure there's no code in the main app that is doing anything outside of those imports. Note that not *everything* is disabled behind this flag, just the logic related to networking.
4. Remove the `APP_TRACKING_PROTECTION` value from the main project file settings, and then make sure AppTP doesn't show up in the debug build in settings

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPhone 14 Pro
* [ ] iPad

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:


DuckDuckGo/HomeCollectionView.swift
```diff
@@ -70,7 +73,7 @@ class HomeCollectionView: UICollectionView {
     
     func configure(withController controller: HomeViewController,
                    favoritesViewModel: FavoritesListInteracting,
-                   appTPHomeViewModel: AppTPHomeViewModel,
+                   appTPHomeViewModel: AnyObject?, // Set to AnyObject so that AppTP can be disabled easily
```
samsymons, line 76: I wanted to put the `AppTPHomeViewModel` class behind the `if APP_TRACKING_PROTECTION` check, but didn't want to completely replace the initializer - the idea is that this should be as simple as possible to undo when we need to, so for now I've just used AnyObject and cast it to a `AppTPHomeViewModel` later in the function. But, feel free to suggest an alternative approach.

SlayterDev (approved):
LGTM! Tested with and without the flag. Worked as expected. I couldn't test archiving because I don't have the proper certificate however.

PR #1733, opened 2023-05-19
Fix multiple dax dialogs

THISISDINOSAUR:
Task/Issue URL: https://app.asana.com/0/414709148257752/1204534039597636/f
Tech Design URL:
CC:

**Description**:
Stops multiple dax dialogs being shown for the same page back to back

**Steps to test this PR**:
Before you start, go to YouTube.com on a fresh install and makes sure you do not see the YouTube tracking/cookies popup, since this stops the original bug from manifesting 
1. On a fresh install, go straight to YouTube.com and ensure only one dax dialog is show. Make sure you do not see the YouTube tracking/cookies popup
2. Continue browsing to a page where trackers load immediately, and ensure you still get the trackers found dax dialog.
3. On a fresh install, search for something, make sure you get the correct dialog. Go to some other page, and make sure you get the correct dialog for if there are or aren't trackers. Then make sure you get the fire button pulse

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPhone 14 Pro
* [ ] iPad

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:

### DIFF
--- a/DuckDuckGo/DaxDialogs.swift
+++ b/DuckDuckGo/DaxDialogs.swift
@@ -163,6 +163,9 @@ final class DaxDialogs {
     private let variantManager: VariantManager
 
     private var nextHomeScreenMessageOverride: HomeScreenSpec?
+    
+    // So we can avoid showing two dialogs for the same page
+    private var lastURLDaxDialogReturnedFor: URL?
 
     /// Use singleton accessor, this is only accessible for tests
     init(settings: DaxDialogsSettings = DefaultDaxDialogsSettings(),
@@ -257,8 +260,19 @@ final class DaxDialogs {
         settings.fireButtonEducationShownOrExpired = true
         return ActionSheetSpec.fireButtonEducation
     }
+    
+    func nextBrowsingMessageIfShouldShow(for privacyInfo: PrivacyInfo) -> BrowsingSpec? {
+        guard privacyInfo.url != lastURLDaxDialogReturnedFor else { return nil }
+        
+        let message = nextBrowsingMessage(privacyInfo: privacyInfo)
+        if message != nil {
+            lastURLDaxDialogReturnedFor = privacyInfo.url
+        }
+        
+        return message
+    }
 
-    func nextBrowsingMessage(privacyInfo: PrivacyInfo) -> BrowsingSpec? {
+    private func nextBrowsingMessage(privacyInfo: PrivacyInfo) -> BrowsingSpec? {
         guard isEnabled, nextHomeScreenMessageOverride == nil else { return nil }
         guard let host = privacyInfo.domain else { return nil }
         
--- a/DuckDuckGo/DaxDialogs.swift
+++ b/DuckDuckGo/DaxDialogs.swift
@@ -24,6 +24,7 @@ import BrowserServicesKit
 import Common
 import PrivacyDashboard
 
+// swiftlint:disable file_length
 // swiftlint:disable type_body_length
 
 protocol EntityProviding {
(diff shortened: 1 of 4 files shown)


DuckDuckGo/DaxDialogs.swift
```diff
@@ -163,6 +163,9 @@ final class DaxDialogs {
     private let variantManager: VariantManager
 
     private var nextHomeScreenMessageOverride: HomeScreenSpec?
+    
+    // So we can avoid showing two dialogs for the same page
+    private var lastURLDaxDialogReturnedFor: URL?
```
THISISDINOSAUR, line 169: I did consider storing this info on the tab, since then we wouldn't need to worry about clearing it with the fire button, but ultimately I think it would be the wrong place for any of this logic to live and compromise testability

DuckDuckGo/DaxDialogs.swift
```diff
@@ -257,8 +265,19 @@ final class DaxDialogs {
         settings.fireButtonEducationShownOrExpired = true
         return ActionSheetSpec.fireButtonEducation
     }
+    
+    func nextBrowsingMessageIfShouldShow(for privacyInfo: PrivacyInfo) -> BrowsingSpec? {
```
miasma13, line 269: I assume this function is supposed to work like the counter part for the home dialogs `peekNextHomeScreenMessage()`?

The only difference I see and maybe it is a problem that all browsing message returning function calls from `nextBrowsingMessage(privacyInfo: PrivacyInfo)` like:
- `searchMessage()`
- `majorTrackerMessage(host)`
- `majorTrackerOwnerMessage(host, owner)`
-  `trackersBlockedMessage(entityNames)` 
-  `noTrackersMessage()`
they all change the state (boolean flags in `settings: DaxDialogsSettings` of what dialogs were already seen). I think it may cause the issue that if we presented one dialog and we attempt to present some other (due to change state) it will be suppressed but not only on this website - the flags will reflect the state as it was already seen by the user hence we will never present it again. I see this as an issue especially as it is the blocked trackers educational Dax Dialog like in the reported YT case.

miasma13 (changes requested):

DuckDuckGo/DaxDialogs.swift
THISISDINOSAUR: I'm not sure I follow. The idea was for it to work slightly differently to peekNextHomeScreenMessage to avoid having to duplicate any logic and make things easier and reduce the chance of introducing new bugs if we ever want to change the order of dax dialogs or anything like that in future. I also wanted to avoid changing the logic of nextBrowsingMessage as much as possible since it's reasonably involved and would be easy to accidentally introduce a new bug.

My intention was for nextBrowsingMessage() to never be called for the new case where we don't intend to show a dialog (i.e. when still on the same page), so that the state isn't modified again, and we should avoid the problem I think you're describing. But I may have either misunderstood what you're saying, or failed to deliver on that intention

miasma13: Sorry, I was me that missed how it was called and guarded. I think you make a good point - no new message will be requested if the has not changed hence the state is coherent.

miasma13 (approved):

PR #1780, opened 2023-06-20
Make SyncQueue serial and suspensible

ayoy:
Task/Issue URL: https://app.asana.com/0/0/1204720532613165/f

**Description**:
When app enters background, sync service will now cancel all sync operations currently in the queue and suspend
sync queue so that it doesn't run any subsequently enqueued operations. A background task is created
to ensure the app has enough time to clean up after interrupted sync.
When app enters foreground, sync service resumes sync queue and runs any previously enqueued operations,
with the caveat that those operations that have been cancelled while going to background are no-ops.

**Steps to test this PR**:
1. Enable sync (1 device is enough).
2. Make sure you have a bookmark.
3. In iOS device Developer settings, enable Network Link Conditioner and set preset to Very Bad Network
4. In the app, edit the bookmark and immediately put the app to background
5. Watch logs marked with `[DDG Sync]` and confirm that sync is cancelled and queue is suspended, and then the background task is ended when sync response is received
6. Bonus points for simulating background task timeout (100% Loss preset didn't work for me, but I have a custom preset with 14s DNS latency that worked) and verifying that the task is forcibly ended.

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPhone 14 Pro
* [ ] iPad

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:


DuckDuckGo/AppDelegate.swift
```diff
@@ -386,13 +388,37 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+    }
+
+    private func suspendSync() {
+        if syncService.isSyncInProgress {
+            os_log(.debug, log: .syncLog, "Sync is in progress. Starting background task to allow it to gracefully complete.")
+
+            var taskID: UIBackgroundTaskIdentifier!
+            taskID = UIApplication.shared.beginBackgroundTask(withName: "Cancelled Sync Completion Task") {
+                os_log(.debug, log: .syncLog, "Forcing background task completion")
+                UIApplication.shared.endBackgroundTask(taskID)
```
ayoy, line 409: There's really nothing we can do here, besides just preventing the OS from killing the app. After this timeout happens, the resumed app has sync in progress, but the sync operation ended for me right after the app was brought to foreground.

bwaresiak (approved):
LGTM!

PR #1807, opened 2023-06-30
Rebuild breakage form with SwiftUI

SlayterDev:
Task/Issue URL: https://app.asana.com/0/inbox/1164476816701138/1203944813756141/1204945810392144
Tech Design URL: https://app.asana.com/0/481882893211075/1204939034484684/f
CC: @brindy 

**Description**:
This PR rebuilds the breakage form with SwiftUI which accomplishes two things:
1. Add a dropdown for breakage category to match other platforms
2. Add a description field so users can add additional information

**Steps to test this PR**:
1. Tapping the report button from the dashboard and browsing menu should open the new form
3. Select a category
4. Entering a description should be optional to submitting the form
5. The bar button and blue submit button should submit the report and dismiss the view
6. Cancel should dismiss the view without submitting

**Copy Testing**:

**Orientation Testing**:

* [x] Portrait
* [x] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [x] iPhone 14 Pro
* [x] iPad

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [x] iOS 16

**Theme Testing**:

* [x] Light theme
* [x] Dark theme

---
###### Internal references:

---
  - https://app.asana.com/0/0/1203944813756141


GuiltyDolphin:
@SlayterDev Couple of points from the guideines (building a checklist at https://app.asana.com/0/0/1204934900441080/f):

1. By default, no category should be selected (copy is "Pick your issue from the list...")
2. User should be able to return to the "no selected" category
3. If user hasn't picked a category, the `category` parameter should be omitted from the report

bwaresiak:
Small UX related thoughts:
- Did you request translations for new/changed strings?
- It's also good to test with "double length pseudolanguage" language setting in scheme to ensure UI scales well.
- Make sure UI handles keyboard well, e.g. form is not obscured by keyboard appearing over it.

brindy (changes requested):
1/ If you tap the submit button top right, it doesn't send the description. 

2/ The CTA style button doesn't appear to have a pressed state *

3/ The button doesn't submit unless you tap the text, it should be the whole button *

4/ Colours and fonts should use the design system (see DesignResourcesKit)

\* You can use the "PrimaryButtonStyle" from `DuckUI` on buttons for now, these will be updated with DesignResourcesKit (or removed entirely) eventually. 

But fonts and stuff should use their dax equivalent, e.g. `daxTitle1()`, `daxBodyRegular()`, etc.  These should be defined in figma. If this UI is too old for that, just use the closest one.

DuckDuckGo/BrokenSiteCategories.swift
```diff
@@ -21,7 +21,7 @@ import Foundation
 
 struct BrokenSite {
     
-    enum Category: String, CaseIterable {
+    enum Category: String, CaseIterable, Identifiable {
```
brindy, line 24: This is fine, but for future reference if we're only adding `Identifiable` for use in SwiftUI you can do this as a private extension or even just use `id: \.self` in the iterator.

brindy (approved):
LGTM, nice work.

PR #1857, opened 2023-07-21
Release script hot fix changes + "coldfix"

THISISDINOSAUR:
Task/Issue URL: https://app.asana.com/0/414235014887631/1204947064806401/f
Tech Design URL:
CC:

**Description**:
Changes the release script to support changes to the hot fixing process, and adds a new "cold fix" process (i.e. a process for doing release fixes before the release has gone out which doesn't increment the release number).

The general idea behind the hotfix changes are that we should only ever merge in the minimum amount to get stable, i.e. only the code that fixes the bug. It also codifies that we always increment the patch number for a hotfix.

For a hotfix:
As input:
- Takes the version number of *the release you want to hotfix* and then handles the versioning of the new release automatically (by incrementing the patch number).
- Takes a branch name with the fix (which should contain only the fix)

It then:
- Checks the tag exists, and then checks out the tag for the release
- Creates a branch for the hotffix, and then a "changes" branch as with the existing process
- Merges the supplied branch with the fix into this changes branch
- Creates the pull request for the changes branch into the hotfix branch as with the existing process

Perhaps more important is what it does not do:
- It does not update embedded files
- It does not pull in develop

A cold fix is the same, except uses the same version number.

**Steps to test this PR**:
1. Test the script still works for a regular release (i.e. by not supplying -h -c or a fix branch). Check that it still handles the version number currently and updates embedded files as expected
2. Create a small test fix of a release tag (e.g. I have been using elle/release-script-hot-fix-changes-test-fix-branch)
3. Test the script by supplying that version number and that branch with the flag -h. (e.g. `scripts/prepare_release.sh 7.80.0 elle/release-script-hot-fix-changes-test-fix-branch -h`). Check the created PR only contains the updated version number (Which should just be an increment of the patch number), updated release notes, the code changes for the fix (and only the fix), and check that it did not update embedded files.
4. Do the same as with 3, except supply -c instead of -h. The outcome should be identical except the version number should be unchanged, and the build number should instead be incremented as usual. 

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPhone 14 Pro
* [ ] iPad

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:

### DIFF
--- a/scripts/prepare_release.sh
+++ b/scripts/prepare_release.sh
@@ -65,6 +65,8 @@ print_usage_and_exit() {
 }
 
 read_command_line_arguments() {
+	number_of_arguments="$#"
+
 	local regexp="^[0-9]+(\.[0-9]+)*$"
 	if [[ ! "$1" =~ $regexp ]]; then
 		print_usage_and_exit "💥 Error: Wrong app version specified"
@@ -106,6 +108,10 @@ read_command_line_arguments() {
 	shift $((OPTIND-1))
 
 	if [[ $is_hotfix ]]; then
+		if [[ $number_of_arguments -ne 3 ]]; then
+			print_usage_and_exit "💥 Error: Wrong number of arguments. Did you specify a fix branch?"
+		fi
+
 		version_to_hotfix=${version}
 		if ! [[ $is_coldfix ]]; then
 			IFS='.' read -ra arrIN <<< "$version"
--- a/scripts/prepare_release.sh
+++ b/scripts/prepare_release.sh
@@ -108,7 +108,7 @@ read_command_line_arguments() {
 	if [[ $is_hotfix ]]; then
 		version_to_hotfix=${version}
 		if ! [[ $is_coldfix ]]; then
-			IFS='.' read -r -a arrIN <<< $version
+			arrIN=(${version//./ })
 			patch_number=$((arrIN[2]+1))
 			version="${arrIN[0]}.${arrIN[1]}.${patch_number}"
 		fi
--- a/scripts/prepare_release.sh
+++ b/scripts/prepare_release.sh
@@ -241,7 +241,9 @@ main() {
 
 	update_marketing_version
 	update_build_version
-	update_embedded_files
+	if ! [ $is_hotfix ]; then
+		update_embedded_files
+	fi
 	update_release_notes
 	merge_fix_branch_if_necessary
 
--- a/scripts/prepare_release.sh
+++ b/scripts/prepare_release.sh
@@ -163,6 +163,9 @@ create_release_branch() {
 }
 
 update_marketing_version() {
+	if [[ $is_coldfix ]]; then
+		return
+	fi
 	printf '%s' "Setting app version ... "
 	"$script_dir/set_version.sh" "${version}"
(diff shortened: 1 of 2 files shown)


scripts/prepare_release.sh
```diff
@@ -50,11 +51,12 @@ print_usage_and_exit() {
 
 	cat <<- EOF
 	Usage:
-	  $ $(basename "$0") <version> [-h] [-v]
+	  $ $(basename "$0") <version> [banch-with-fix] [-h] [-v]
```
jaceklyp: typo :)

scripts/prepare_release.sh
```diff
@@ -86,7 +105,16 @@ read_command_line_arguments() {
...
 	shift $((OPTIND-1))
 
-	[[ $is_hotfix ]] && branch_name="hotfix" || branch_name="release"
+	if [[ $is_hotfix ]]; then
+		version_to_hotfix=${version}
+		if ! [[ $is_coldfix ]]; then
+			# shellcheck disable=SC2206
+			arrIN=(${version//./ })
+			patch_number=$((arrIN[2]+1))
+			version="${arrIN[0]}.${arrIN[1]}.${patch_number}"
```
jaceklyp: perhaps you can use IFS?

    IFS='.' read -ra arrIN <<< "$version"
    patch_number=$((arrIN[2] + 1))
    version="${arrIN[0]}.${arrIN[1]}.$patch_number"

scripts/prepare_release.sh
```diff
@@ -50,11 +51,12 @@ print_usage_and_exit() {
 
 	cat <<- EOF
 	Usage:
-	  $ $(basename "$0") <version> [-h] [-v]
+	  $ $(basename "$0") <version> [banch-with-fix] [-h] [-v]
 	  Current version: $(cut -d' ' -f3 < "${base_dir}/Configuration/Version.xcconfig")
 
 	Options:
-	  -h         Make hotfix release
+	  -h         Make hotfix release. Requires the version to be the one to hotfix, and a branch with the fix as the second parameter
```
jaceklyp, line 58: do we provide any error info if someone forgets to provide a branch name?

jaceklyp (changes requested):

scripts/prepare_release.sh
THISISDINOSAUR: No, it fails silently (although at least in a fairly obvious way:
"Merging fix branch ... %                                                     
ellesullivanwork@Elles-MBP iOS %"

Is there an easy way to do it?

THISISDINOSAUR: I've added a check based on the number of arguments:

"Usage:
 ...
💥 Error: Wrong number of arguments. Did you specify a fix branch?"

jaceklyp (approved):

PR #1864, opened 2023-07-25
Do not process updates after screen has been dismissed

bwaresiak:
Task/Issue URL: https://app.asana.com/0/414709148257752/1205002344719381
Tech Design URL:
CC:

**Description**:

Attempt to solve the bug by making sure no updates are processed once the screen is dismissed.

**Steps to test this PR**:

Smoke tests of tab switcher.

**Device Testing**:

* [ ] iPhone 
* [ ] iPad

---
###### Internal references:

### DIFF
--- a/DuckDuckGo/TabSwitcherViewController.swift
+++ b/DuckDuckGo/TabSwitcherViewController.swift
@@ -71,7 +71,7 @@ class TabSwitcherViewController: UIViewController {
     var currentSelection: Int?
     
     private var tabSwitcherSettings: TabSwitcherSettings = DefaultTabSwitcherSettings()
-    private var isProcessingUpdates = false
+    private var ignoreProcessingUpdates = false
 
     let favicons = Favicons.shared
     
@@ -332,6 +332,7 @@ class TabSwitcherViewController: UIViewController {
     }
 
     override func dismiss(animated: Bool, completion: (() -> Void)? = nil) {
+        ignoreProcessingUpdates = true
         tabsModel.tabs.forEach { $0.removeObserver(self) }
         super.dismiss(animated: animated, completion: completion)
     }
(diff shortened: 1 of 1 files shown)


DuckDuckGo/TabSwitcherViewController.swift
```diff
@@ -71,7 +71,7 @@ class TabSwitcherViewController: UIViewController {
     var currentSelection: Int?
     
     private var tabSwitcherSettings: TabSwitcherSettings = DefaultTabSwitcherSettings()
-    private var isProcessingUpdates = false
+    private var ignoreProcessingUpdates = false
```
brindy: I see Bunn is going to review, but my comment here is that double negatives are harder reason over.  This imo should be called `shouldProcessUpdates` then the guard in `didChange` will read more easily to understand, e.g. 

`guard shouldProcessUpdates else ... `

DuckDuckGo/TabSwitcherViewController.swift
```diff
@@ -349,12 +350,12 @@ extension TabSwitcherViewController: TabViewCellDelegate {
...
         } else {
             collectionView.performBatchUpdates({
-                isProcessingUpdates = true
+                ignoreProcessingUpdates = true
                 delegate.tabSwitcher(self, didRemoveTab: tab)
                 currentSelection = tabsModel.currentIndex
                 collectionView.deleteItems(at: [IndexPath(row: index, section: 0)])
             }, completion: { _ in
-                self.isProcessingUpdates = false
+                self.ignoreProcessingUpdates = false
```
brindy: Can this completion be called after `dismiss` if so, this might get set back and defeats the point of this PR?

DuckDuckGo/TabSwitcherViewController.swift
Bunn: I had a similar suggestion ready to go, but I prefer @brindy suggestion `shouldProcessUpdates` sounds better than what I was going to suggest `canProcessUpdates`

DuckDuckGo/TabSwitcherViewController.swift
bwaresiak: I think this is pretty unlikely, but I can add another boolean to separate delete scenario from dismiss scenario

Bunn: Not sure if I understand. If the completion is called **after** `dismiss` then the user would not see the updates in the switcher when removing a tab for example.
The selection and title wouldn't update.

bwaresiak: This is both about timing and about state changes. 
Timing: Assuming there can be scenario in which completion is called after dismiss, that flag would get set to false allowing (incorrectly) for processing of pending updates.
State: expectation is that once that flag is set in 'dismiss', it should never change.

That's why I'm keen to add one and make flag names as follow:
- isProcessingUpdates (as it was)
- isDismissed

That will still result in negation in guards, but it is explaining what is happening.

Bunn (approved):
Did a quick pair programming with Bartek while he did some changes on this. LGTM!

PR #1899, opened 2023-08-07
NetP Status View Header

graeme:
Task/Issue URL: https://app.asana.com/0/0/1205084446087078/f

**Description**:

This is the first design implementation task in the overarching Network Protection Status View task and covers just the header view (that is, the views above the toggle view). I’m delaying on making this look good on iOS 14 for now as it will take some extra thought. But I am checking in with the Apple team and have created a follow-up task to fix iOS 14 at the end of the project if need be.

**Steps to test this PR**:
1. (If you haven’t followed the steps in [Network Protection iOS Smoke Tests: Invite Code flow](https://app.asana.com/0/0/1205201898139908) already, do so.)
2. **Check that the status view header (the views above, not including, the toggle view) matches [the designs](https://www.figma.com/file/zuCBX2UpJpS8CgtcmpU1hf/Network-Protection-for-iOS?type=design&node-id=252-21943&mode=design&t=Z7gEv95kIs3uMn7e-0) disconnected state** 
3. Switch the toggle to activate Network Protection.
4. **Check that the status view header again matches the designs but for the connected state**

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [x] iPhone 8
* [ ] iPhone X
* [ ] iPhone 14 Pro
* [ ] iPad

**OS Testing**:

* [ ] iOS 14
* [x] iOS 15
* [x] iOS 16

**Theme Testing**:

* [x] Light theme
* [x] Dark theme

—
###### Internal references:


DuckDuckGo/NetworkProtectionStatusView.swift
```diff
@@ -74,6 +98,49 @@ struct NetworkProtectionStatusView: View {
     }
 }
 
+private extension View {
+    @ViewBuilder
+    func hideScrollContentBackground() -> some View {
```
graeme, line 103: This is to override the system list background colour. Quite an ugly solution pre-iOS 16 but not sure how else to achieve it...

DuckDuckGo/NetworkProtectionStatusViewModel.swift
```diff
@@ -21,59 +21,60 @@
...
+            .assign(to: \.isNetPEnabled, onWeaklyHeld: self)
+            .store(in: &cancellables)
+        isConnectedPublisher
+            .map(Self.titleText(connected:))
+            .assign(to: \.headerTitle, onWeaklyHeld: self)
+            .store(in: &cancellables)
+        isConnectedPublisher
+            .map(Self.statusImageID(connected:))
+            .assign(to: \.statusImageID, onWeaklyHeld: self)
+            .store(in: &cancellables)
```
diegoreymendez, line 75: Just a note that these publishers are not doing `.receive(on: DispatchQueue.main)`, whereas the ones below are.

Since the model is mostly used for UI we may want to add that for peace of mind.

DuckDuckGo/NetworkProtectionStatusView.swift
```diff
@@ -48,17 +57,32 @@ struct NetworkProtectionStatusView: View {
...
+                    Image(statusModel.statusImageID)
+                        .resizable()
+                        .scaledToFit()
+                        .frame(height: 96)
+                        .padding(8)
+                    Text(statusModel.headerTitle)
+                        .font(.system(size: 17, weight: .semibold))
+                        .multilineTextAlignment(.center)
+                        .foregroundColor(.titleText)
+                    Text(UserText.netPStatusHeaderMessage)
```
diegoreymendez, line 76: I'm sharing this with you because I didn't know this until yesterday, and because I have the doubt that the font is correct.  I'm not sure if `.system(...)` returns "SF Pro", so apologies if it does.

You can get the SwiftUI styling code directly from Figma, as shown below.

Screenshot 2023-08-10 at 14 06 52

diegoreymendez (commented):
I'm still testing but I'm also adding some comments for you to start looking at.

DuckDuckGo/NetworkProtectionStatusView.swift
graeme: Yeah, system does return that. But this is still useful as I did not know about that. Thanks!

DuckDuckGo/NetworkProtectionStatusViewModel.swift
```diff
@@ -86,20 +87,56 @@ final class NetworkProtectionStatusViewModel: ObservableObject {
...
+    private class func statusImageID(connected isConnected: Bool) -> String {
+        isConnected ? "VPN" : "VPNDisabled"
+    }
+}
+
+private extension ConnectionStatus {
+    var message: String {
+        switch self {
+        case .notConfigured:
+            return "Not Configured"
```
diegoreymendez: Just a note that at least in macOS we don't show this status even though it exists... we show `.disconnected`.

I don't think it needs to block this PR but something worth synching with between you, Robert and myself maybe?

DuckDuckGo/NetworkProtectionStatusViewModel.swift
```diff
@@ -86,20 +87,56 @@ final class NetworkProtectionStatusViewModel: ObservableObject {
...
+        case .disconnected:
+            return "Disconnected"
```
diegoreymendez: Adding to my comment above, in macOS the text for `.disconnected` is "Not connected".  I also checked the designs and it seems it's the same for iOS.

DuckDuckGo/NetworkProtectionStatusViewModel.swift
diegoreymendez: Also these strings should probably be `LocalizedString` (since they're using facing)?

I know SwiftUI handles that on its own, so I'm not sure whether this is needed or optional for iOS.  Please let me know so I can learn about it!

diegoreymendez (approved):
Added several comments to try and help with things that I believe may need to be adjusted (either in this PR or in a follow up).

But none of these things should be binding at this point, so I'm approving.

DuckDuckGo/NetworkProtectionStatusViewModel.swift
graeme: The shared publisher that these all use already dispatches onto main :) Just added it there to reduce the repetition

PR #1920, opened 2023-08-16
Post merge PR check

Bunn:
Task/Issue URL: https://app.asana.com/0/1203301625297703/1205270007589105/f

**Description**:
Creates asana task is post merge PR check fails

**Steps to test this PR**:
1. Create a new PR from this branch
2. Remove develop branch check and add code errors
3. Check if task is created under iOS section on https://app.asana.com/0/1205237866452338/1205239782816625 project

Example: https://app.asana.com/0/inbox/1201011656765575/1205286484229414/1205286484229416

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -105,7 +105,7 @@ jobs:
     name: Create Asana Task
     needs: [swiftlint, unit-tests]
 
-    if: failure()
+    if: failure() && github.ref_name == 'develop'
     
     env:
       WORKFLOW_URL: https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}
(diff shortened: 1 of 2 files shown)


.github/workflows/pr.yml
```diff
@@ -101,3 +101,24 @@ jobs:
       with:
         report_paths: unittests.xml
 
+  asana:
+    name: Create Asana Task
+    needs: [swiftlint, unit-tests]
```
ayoy: Why shellcheck is not included? Now I've noticed that it's also not included in macOS. We should probably include it too.

ayoy (approved):
LGTM, but I'm wondering why shellcheck is ignored. Thanks @Bunn!

.github/workflows/pr.yml
Bunn: @ayoy Good point, it was deliberated because I thought that we couldn't have flaky issues with shellcheck, but I guess the same could be said to swiftlint I'll add it here and add to macOS later on. Thanks :)

PR #1952, opened 2023-08-25
Keep Kingfisher tab cache in sync with open tabs

amddg44:
Task/Issue URL: https://app.asana.com/0/414709148257752/1204375718000558/f
Tech Design URL:
CC:

**Description**:
One off cache clean-up of the tabs favicon cache + functionality added so that when a tab is closed, its favicon is deleted from tabs cache (as long as no other tabs have the same host)

**Steps to test this PR**:
Note: I branched off of updated Image Cache screen work so the Image Cache screen can be used for validation while testing 🙂 
1. To test the one-off clean up, ideally test from a device with orphaned tab favicons and confirm that after launching the app there are no longer any orphaned tab reported in the Debug > Image Cache screen (and confirm see no impact on performance in terms of opening + interacting with tabs)
2. Open a few tabs with different links for the same site
3. As you close each tab, confirm the favicon for the site remains in the Image Cache until the final tab for that site is closed. 

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPhone 14 Pro
* [ ] iPad

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:


DuckDuckGo/TabsModel.swift
```diff
@@ -152,4 +152,14 @@ public class TabsModel: NSObject, NSCoding {
...
     }
+    
+    func tabExists(withHost host: String) -> Bool {
+        return !tabs.filter {
+            if let tabHost = $0.link?.url.host {
+                return tabHost == host
+            }
+
+            return false
+        }.isEmpty
```
samsymons: We could speed this up using `contains`, since the current implementation will filter the entire tab list every time, even when the tab it's looking for is one of the first in the list. `contains` on the other hand should just stop checking once it finds a match:

```suggestion
        return tabs.contains { $0.link?.url.host == host }
```

samsymons (changes requested):

DuckDuckGo/TabsModel.swift
amddg44: Thanks for the suggestion! I've made that change + tested

samsymons (approved):
LGTM, I tested this with a bunch of different cases, and also ran it on my production device to confirm that orphaned favicons were removed.

I do still see a few orphaned fireproof favicons specifically, but the main set of icons I was concerned about are gone. Thanks for this!

PR #1957, opened 2023-08-25
Separate bundle and group ids in iOS Alpha config file

graeme:
Task/Issue URL: https://app.asana.com/0/0/1205101983701300/f

**Description**:

On implementing [PR: Separate app groups across Alpha and other builds](https://app.asana.com/0/414235014887631/1205332058642473/f), the app groups were updated for Alpha builds, but the generic Configuration.xcconfig that was used for all build configs has defined a single app group prefix for all configurations. This resulted in a crash on startup.

To fix this, we’ll need to define a separate app group prefix. This could be done through an override in the Xcode Build Settings, but I feel it’s cleaner to do this through a Configuration-Alpha.xcconfig.

**Steps to test:**

1. Check out this branch.
2. For the Alpha configuration, temporarily change the it to automatic signing through Xcode
3. (If this fails, you may have to do what I did and add a temporary Fastlane lane called sync_signing_dev_alpha in the Fastfile then add a for_lane block for that in the Matchfile which is a copy of the sync_signing_alpha one but with type “development”, then run bundle exec fastlane sync_signing_dev_alpha and use the resulting profiles from that… but hopefully it doesn’t come to that as it only seems to be me having that problem.)
4. Build this (preferably on top of an old Alpha from TestFlight, but it doesn’t really matter).
5. Check that it starts up and check that you can start Network Protection and add a Bookmark.

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPhone 14 Pro
* [ ] iPad

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:

### DIFF
--- a/Configuration/Configuration-Alpha.xcconfig
+++ b/Configuration/Configuration-Alpha.xcconfig
@@ -1,4 +1,8 @@
-//  Copyright © 2019 DuckDuckGo. All rights reserved.
+//
+//  Configuration-Alpha.xcconfig
+//  DuckDuckGo
+//
+//  Copyright © 2023 DuckDuckGo. All rights reserved.
 //
 //  Licensed under the Apache License, Version 2.0 (the "License");
 //  you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


samsymons (approved):
LGTM!

PR #1978, opened 2023-09-05
PacketTunnelProvider crash debug buttons

graeme:
Task/Issue URL: https://app.asana.com/0/0/1205376937712482/f
Tech Design URL:
CC:

**Description**:

To verify that crash reporting works as expected from the PacketTunnelProvider extension, this PR adds debug menu items to trigger two types of crash: a fatal error and a memory overuse crash.

Note that this is an MVP of this functionality to verify that crash reports are sending. MacOS has a more elegant way of doing this realtime through sending messages, but I’ve kept this a bit simpler on iOS to start with.

Also note that iOS now has an implementation where simulation options are mutually exclusive. I’m aware this introduces a divergence between the two platforms, but thought I’d use this as somewhat of a proposal for how we could handle these as it’s a bit more concise code-wise and I believe it doesn’t really make sense to trigger multiple simulation options at the same time.

**Steps to test this PR**:
1. Ensure Settings -> Debug -> Internal User is toggled
2. Ensure NetworkProtection is not running
3. Go to Settings -> Debug -> Network Protection
4. Press one of the two crash simulation menu items (try both)
5. Go to Settings -> Network Protection and toggle the start switch
6. Confirm that it tried to connect then instantly fails (because 💥)

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPhone 14 Pro
* [ ] iPad

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

—
###### Internal references:


diegoreymendez (approved):
Looks good to me, works well.

PR #1997, opened 2023-09-09
fix:Added Async/Await

jairajput:
Task/Issue URL:
Tech Design URL:
CC:

**Description**:

**Steps to test this PR**:
1.
2.

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPhone 14 Pro
* [ ] iPad

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:

### DIFF
--- a/DuckDuckGo/AuthenticationViewController.swift
+++ b/DuckDuckGo/AuthenticationViewController.swift
@@ -58,17 +58,36 @@ class AuthenticationViewController: UIViewController {
             onCouldNotAuthenticate()
         }
     }
+    
+    class Authenticator {
+        func canAuthenticate() -> Bool {
+            return true // Replace with your logic
+        }
+
+        func authenticate(reason: String) async throws -> Bool {
+            
+            return true
+        }
+    }
+
 
     private func authenticate() {
         hideUnlockInstructions()
-        authenticator.authenticate { (success, _) in
-            if success {
-                self.onAuthenticationSucceeded()
-            } else {
-                self.onAuthenticationFailed()
+        Task {
+            do {
+                let success = try await authenticator.authenticate(reason: "Authentication required")
+                if success {
+                    onAuthenticationSucceeded()
+                } else {
+                    onAuthenticationFailed()
+                }
+            } catch {
+                
+                onAuthenticationFailed()
             }
         }
     }
+
 
     @IBAction func onTap(_ sender: Any) {
         authenticate()
--- a/DuckDuckGo/AuthenticationViewController.swift
+++ b/DuckDuckGo/AuthenticationViewController.swift
@@ -61,7 +61,7 @@ class AuthenticationViewController: UIViewController {
     
     class Authenticator {
         func canAuthenticate() -> Bool {
-            return true // Replace with your logic
+            return true
         }
 
         func authenticate(reason: String) async throws -> Bool {
@@ -70,7 +70,7 @@ class AuthenticationViewController: UIViewController {
         }
     }
 
-
+// Added Async/Await
     private func authenticate() {
         hideUnlockInstructions()
         Task {


jairajput:
@THISISDINOSAUR  Can You Please Review My Code

DuckDuckGo/AuthenticationViewController.swift
```diff
@@ -58,18 +58,37 @@ class AuthenticationViewController: UIViewController {
             onCouldNotAuthenticate()
         }
     }
+    
+    class Authenticator {
```
samsymons, line 62: We shouldn't nest a new `Authenticator` class in here - instead, the [original class](https://github.com/duckduckgo/iOS/blob/develop/DuckDuckGo/Authenticator.swift) should be modified.

Furthermore, the original class [already has an `async` variant](https://github.com/duckduckgo/iOS/blob/178157a7896bf976076a89a3e9a7fbc751edfa5b/DuckDuckGo/Authenticator.swift#L46) of the authenticate function, so this class isn't necessary.

DuckDuckGo/AuthenticationViewController.swift
```diff
@@ -58,18 +58,37 @@ class AuthenticationViewController: UIViewController {
...
+        func canAuthenticate() -> Bool {
+            return true
```
samsymons, line 64: This is inaccurate; there are cases where authentication isn't available, so we can't return `true` at all times. It's up to the `LocalAuthentication` framework to tell us when it's available instead.

DuckDuckGo/AuthenticationViewController.swift
```diff
@@ -58,18 +58,37 @@ class AuthenticationViewController: UIViewController {
...
+        }
+
+        func authenticate(reason: String) async throws -> Bool {
+            
+            return true
+        }
```
samsymons, line 70: This is inaccurate in a few ways:

1. The function is `async` but it's not `await`ing anything
2. It returns true no matter what, when instead it should be going through `LocalAuthentication` - in this state, the authentication feature will never present an authentication prompt
3. The `reason` parameter isn't used

DuckDuckGo/AuthenticationViewController.swift
```diff
@@ -58,18 +58,37 @@ class AuthenticationViewController: UIViewController {
...
+    }
 
+// Added Async/Await
```
samsymons, line 73: We shouldn't add a comment here with what has changed - we use source control for that instead. 🙂

DuckDuckGo/AuthenticationViewController.swift
```diff
@@ -58,18 +58,37 @@ class AuthenticationViewController: UIViewController {
...
     private func authenticate() {
         hideUnlockInstructions()
-        authenticator.authenticate { (success, _) in
-            if success {
-                self.onAuthenticationSucceeded()
-            } else {
-                self.onAuthenticationFailed()
+        Task {
+            do {
+                let success = try await authenticator.authenticate(reason: "Authentication required")
```
samsymons, line 78: The reason is hardcoded in English, but we translate our app into other languages, so it should come from the `UserText` file instead.

samsymons:
Hey @jairajput!

I'll have to close this PR unfortunately, as it doesn't have an associated issue proposing to fix the issue. We request that all third party contributors propose their changes first so that we can make sure that we're in a position to accept the change to begin with.

I've taken a look at the code anyway and left some areas for improvement, in case it's useful. Thanks!

jairajput:
Thanks For The Review

PR #2019, opened 2023-09-18
Network Protection Error Pixels

graeme:
Task/Issue URL: https://app.asana.com/0/0/1205365033608088/f

**Description**:

We want to report errors when the user tries to activate Network Protection on macOS, or when Network Protection encounters an unexpected failure. This includes:
- When we fail to read a stored network configuration file from the Keychain
- When we fail to write a server list file to the application support directory
- When the WireGuard framework reports an error back to the macOS client
- When the Network Extension receives an error during activation

We would like to append a suffix (`_d` | `_c`) on each pixel to represent whether or not this was the first time the pixel was fired each day for the user.
This is to let us determine whether a spike in pixels is distributed across many devices, or isolated to a small number of devices that are in a bad state and sending pixels repeatedly.

**Steps to test this PR**:
Testing all the Pixels may be somewhat overkill and, in some cases, not even possible.

1. Select a few of the Pixels that should be reproducible (e.g `m_netp_network_extension_error_activation_request_failed_` which should fire if the user denies VPN permissions)
2. Run the PacketTunnelProvider in the debugger
3. Put a breakpoint in the `fire(pixelNamed` function
4. Check that it fires as expected with all the correct parameters
—
###### Internal references:


graeme:
@samsymons I fixed a couple of issues I noticed. See my two commits from today.

PacketTunnelProvider/NetworkProtectionPacketTunnelProvider.swift
```diff
@@ -20,24 +20,142 @@
...
+            source.setEventHandler {
+                let event: DispatchSource.MemoryPressureEvent  = source.mask
+                print(event)
+                switch event {
+                case DispatchSource.MemoryPressureEvent.normal:
+                    break
+                case DispatchSource.MemoryPressureEvent.warning:
+                    Pixel.fire(pixel: .networkProtectionMemoryWarning)
+                case DispatchSource.MemoryPressureEvent.critical:
+                    Pixel.fire(pixel: .networkProtectionMemoryCritical)
```
samsymons, line 163: I'll be curious to see if these ones ever come through - will the extension have enough time to send the pixel before it gets killed? If we start getting warning pixels but never critical ones then that would be my theory, and we can look at implementing an offline pixel cache to catch situations like this.

samsymons (approved):
As far as I can see, this is working really nicely :+1:

PR #2037, opened 2023-09-25
Implement config changes to UA

jaceklyp:
Task/Issue URL: https://app.asana.com/0/0/1205442309571983/f
CC: @jonathanKingston 

**Description**:
It allows to modify UA through config by either using the old logic, safari UA or safari UA with DDG component.

**Steps to test this PR**: TBD
1.
2.

### DIFF
--- a/Core/UserAgentManager.swift
+++ b/Core/UserAgentManager.swift
@@ -197,11 +197,11 @@ struct UserAgent {
             url?.isPart(ofDomain: domain) ?? false
         }) { return ddgFixedLogic(forUrl: url, isDesktop: isDesktop, privacyConfig: privacyConfig) }
 
-        if closestUserAgentVersions(forConfig: privacyConfig).contains(statistics.atb ?? "") {
+        if closestUserAgentVersions(forConfig: privacyConfig).contains(statistics.atbWeek ?? "") {
             return closestLogic(forUrl: url, isDesktop: isDesktop, privacyConfig: privacyConfig)
         }
 
-        if ddgFixedUserAgentVersions(forConfig: privacyConfig).contains(statistics.atb ?? "") {
+        if ddgFixedUserAgentVersions(forConfig: privacyConfig).contains(statistics.atbWeek ?? "") {
             return ddgFixedLogic(forUrl: url, isDesktop: isDesktop, privacyConfig: privacyConfig)
         }
 
@@ -341,4 +341,19 @@ struct UserAgent {
         return "\(Constants.desktopPrefixComponent) \(suffix) \(versionComponent)"
     }
     
+}
+
+private extension StatisticsStore {
+
+    var atbWeek: String? {
+        guard let atb else { return nil }
+        let trimmed = String(atb.dropFirst())
+
+        if let hyphenIndex = trimmed.firstIndex(of: "-") {
+            return String(trimmed.prefix(upTo: hyphenIndex))
+        } else {
+            return trimmed
+        }
+    }
+
 }
--- a/Core/UserAgentManager.swift
+++ b/Core/UserAgentManager.swift
@@ -85,8 +85,16 @@ public class DefaultUserAgentManager: UserAgentManager {
 }
 
 struct UserAgent {
+
+    private enum DefaultPolicy: String {
+
+        case ddg
+        case ddgFixed
+        case closest
+
+    }
     
-    private struct Constants {
+    private enum Constants {
         // swiftlint:disable line_length
         static let fallbackWekKitVersion = "605.1.15"
         static let fallbackSafariComponent = "Safari/\(fallbackWekKitVersion)"
@@ -96,6 +104,10 @@ struct UserAgent {
         
         static let uaOmitSitesConfigKey = "omitApplicationSites"
         static let uaOmitDomainConfigKey = "domain"
+
+        static let defaultPolicyConfigKey = "defaultPolicy"
+        static let ddgDefaultSitesConfigKey = "ddgDefaultSites"
+        static let ddgFixedSitesConfigKey = "ddgFixedSites"
         // swiftlint:enable line_length
     }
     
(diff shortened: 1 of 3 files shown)


Core/UserAgentManager.swift
```diff
@@ -124,23 +136,101 @@ struct UserAgent {
...
             return concatWithSpaces(baseDesktopAgent, resolvedApplicationComponent, safariComponent)
         } else {
             return concatWithSpaces(baseAgent, resolvedApplicationComponent, safariComponent)
         }
     }
+
+    private func closestLogic(forUrl url: URL?,
+                              isDesktop: Bool,
+                              privacyConfig: PrivacyConfiguration = ContentBlocking.shared.privacyConfigurationManager.privacyConfig) -> String {
+        let customUAEnabled = privacyConfig.isEnabled(featureKey: .customUserAgent)
```
miasma13: From the logic it turns out that `func closestLogic` is only called when the `customUAEnabled` is enabled.

miasma13 (approved):

miasma13:
Overall it looks fine. One small comment and the tests are failing due to user agent changes.

Core/UserAgentManager.swift
```diff
@@ -124,23 +136,101 @@ struct UserAgent {
...
+        // If the UA should have DuckDuckGo append it prior to Safari
+        if let resolvedApplicationComponent {
+            if let index = defaultSafari.range(of: "Safari")?.lowerBound {
+                defaultSafari.insert(contentsOf: resolvedApplicationComponent + " ", at: index)
+            }
+        }
+        return defaultSafari
+    }
+
+    private func ddgFixedLogic(isDesktop: Bool) -> String {
```
jonathanKingston: This method should be the `closest` policy, fixed will contain the DuckDuckGo part.

Maybe the confusion came in that this is now a fixed UA, but the fixed policy is about repairing known issues in our existing UA.

jonathanKingston: This method shouldn't be used when the user is <16.6 also.

miasma13 (approved):
LGTM

Core/UserAgentManager.swift
```diff
@@ -253,32 +253,44 @@ struct UserAgent {
...
+        return nil
     }
-    
+
+    var deviceProfile: String {
+        let regex = try? NSRegularExpression(pattern: "\\((.*?)\\)")
+        if let match = regex?.firstMatch(in: baseAgent, range: NSRange(baseAgent.startIndex..., in: baseAgent)) {
+            let range = Range(match.range(at: 1), in: baseAgent)
+            if let range = range {
+                return String(baseAgent[range])
```
jonathanKingston, line 296: I've not validated yet this but the approach looks correct.

Core/UserAgentManager.swift
```diff
@@ -253,32 +253,44 @@ struct UserAgent {
...
 
-    private var shouldUseUpdateLogic: Bool {
-        // Unfortunately you can't combine #available with any other condition in the same line
-        if isTesting {
-            return true
-        } else if #available(iOS 16.6, *) {
-            return true
+    private var shouldUseUpdatedLogic: Bool {
+        guard let webKitVersion else { return false }
+        return webKitVersion.versionCompare("605.1.15") != .orderedAscending
```
jonathanKingston, line 277: I think this will be true for "605.1.14" and "605.1.16" etc which is incorrect.

I think the following might be what we're wanting:
```suggestion
        return webKitVersion.versionCompare("605.1.14") == .orderedDescending
```

jaceklyp: I believe this is fine - added two tests that should validate that - please could you take a look at them?

DuckDuckGoTests/UserAgentTests.swift
```diff
@@ -28,6 +28,8 @@ final class UserAgentTests: XCTestCase {
     private struct DefaultAgent {
         static let mobile = "Mozilla/5.0 (iPhone; CPU iPhone OS 12_4 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Mobile/15E148"
         static let tablet = "Mozilla/5.0 (iPad; CPU OS 12_4 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Mobile/15E148"
+        static let oldWebkitVersionMobile = "Mozilla/5.0 (iPhone; CPU iPhone OS 12_4 like Mac OS X) AppleWebKit/605.1.14 (KHTML, like Gecko) Mobile/15E148"
+        static let newWebkitVersionMobile = "Mozilla/5.0 (iPhone; CPU iPhone OS 12_4 like Mac OS X) AppleWebKit/605.1.16 (KHTML, like Gecko) Mobile/15E148"
```
jonathanKingston, line 32: Can you also add a test for `AppleWebKit/605.1.15` too? I'm a little nervous based on that comparison

jaceklyp: of course, I didn't add it initially, because pretty much everything else testes this use case

Core/UserAgentManager.swift
```diff
@@ -124,24 +145,160 @@ struct UserAgent {
...
+        if ddgDefaultSites(forConfig: privacyConfig).contains(where: { domain in
+            url?.isPart(ofDomain: domain) ?? false
+        }) { return oldLogic(forUrl: url, isDesktop: isDesktop, privacyConfig: privacyConfig) }
+
+        if ddgFixedSites(forConfig: privacyConfig).contains(where: { domain in
+            url?.isPart(ofDomain: domain) ?? false
+        }) { return ddgFixedLogic(forUrl: url, isDesktop: isDesktop, privacyConfig: privacyConfig) }
+
+        if closestUserAgentVersions(forConfig: privacyConfig).contains(statistics.atbWeek ?? "") {
+            if canUseClosestLogic {
```
jonathanKingston, line 201: You could bury this check within the `closestLogic` as you have for `ddgFixedLogic` and just keep the single other check within `ddgFixedLogic` to keep the fall-back behaviour correct.

But also... we're going to rip this out quickly again hopefully too; so feel free to ignore.

jaceklyp: That was my initial idea, but it caused a side effect that made one test fail - when the closest user agent "could not be used" because of old webkit version it fallbacked to oldUserAgentLogic which returned the old UA string which contained "DuckDuckGo/7" component. So inside ddgFixedLogic we were appending yet another DuckDuckGo/7 component.

So my idea was to make it as specific as it can be, because we want to remove it soon.

jonathanKingston: Yeah you'd still need the check in `ddgFixedLogic` to keep the same structure I think that should work (but also not tried).

PR #2130, opened 2023-11-06
kill variant when receiving atb update

brindy:
Task/Issue URL: https://app.asana.com/0/414709148257752/1205887954648070/f
Tech Design URL:
CC:

**Description**:

Removes variant when a remote update to the installed ATB is retrieved.

**Steps to test this PR**:
1. Perform with a clean simulator `xcrun simctl erase all`
2. Ensure the DefaultManager will allocate a variant
3. Put a break point in StatisticsLoader.swift at line 83
4. Execute debugger command: `po self.statisticsStore.atb = "v230-5"` to set ATB to an old value
5. Allow execution to continue
6. Perform a search
7. Confirm that ATB is recent to the most recent value and that the variant has been removed - you can do this by using a proxy or by breakpointing on the StatisticsLoader in the completion of the `refreshSearchRetentionAtb` function 

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPhone 14 Pro
* [ ] iPad

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:

### DIFF
--- a/Core/StatisticsLoader.swift
+++ b/Core/StatisticsLoader.swift
@@ -134,6 +134,7 @@ public class StatisticsLoader {
     public func storeUpdateVersionIfPresent(_ atb: Atb) {
         if let updateVersion = atb.updateVersion {
             statisticsStore.atb = updateVersion
+            statisticsStore.variant = nil
             returnUserMeasurement.updateStoredATB(atb)
         }
     }


Core/StatisticsLoader.swift
```diff
@@ -134,6 +134,7 @@ public class StatisticsLoader {
     public func storeUpdateVersionIfPresent(_ atb: Atb) {
         if let updateVersion = atb.updateVersion {
             statisticsStore.atb = updateVersion
+            statisticsStore.variant = nil
```
cmonfortep, line 137: do we know for sure that update atb doesn't contain variant? is that documented somewhere so we can trust the change is just doing this?

brindy: Yes, see https://dub.duckduckgo.com/duckduckgo/ddg/blob/bttf/lib/DDG/Util/ATB.pm#L127

If you want to parse out the update version and apply the atb and variant separately to remain future proof then that's fine, but my understanding is this won't change as there's no value to keeping the variant (the whole point is to make small buckets of clients a little bigger in order to remain anonymous).

bwaresiak (approved):
LGTM!

PR #2142, opened 2023-11-09
Add NetP widget

samsymons:
Task/Issue URL: https://app.asana.com/0/414235014887631/1205921206830836/f
Tech Design URL:
CC: @graeme 

**Description**:

This PR adds a NetP widget. It's only available in the debug and alpha builds, and only available on iOS 17.

**Steps to test this PR**:
1. First, update the `#if ALPHA` checks in this PR to include `DEBUG` - the alpha checks are because we would like to get the widget into the alpha builds, but we need CI to support Xcode 15, which we can't do for any workflows that run tests unfortunately
1. Run the app on your device
2. Add the widget from the widget UI
3. Toggle NetP within the app, check that the widget state updates accordingly
4. Try the "Connect" and "Disconnect" buttons inside the widget - note that the widget does not show connecting/disconnecting states, this functionality is pretty hacky at the moment

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPhone 14 Pro
* [ ] iPad

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:


DuckDuckGo/AppDelegate.swift
```diff
@@ -536,7 +536,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         }
 
         NotificationCenter.default.post(name: AutofillLoginListAuthenticator.Notifications.invalidateContext, object: nil)
-        mainViewController?.clearNavigationStack()
+        // mainViewController?.clearNavigationStack()
```
samsymons: Need to fix this, it was getting in the way of the open URL functionality. I'll update it to avoid this check depending on the URL.

DuckDuckGo/NetworkProtectionStatusViewModel.swift
```diff
@@ -57,7 +58,11 @@ final class NetworkProtectionStatusViewModel: ObservableObject {
 
     // MARK: Toggle Item
     @Published public var isNetPEnabled = false
-    @Published public var statusMessage: String
+    @Published public var statusMessage: String {
+        didSet {
+            WidgetCenter.shared.reloadTimelines(ofKind: "VPNStatusWidget")
+        }
+    }
```
samsymons: There is likely a better place for this, this was the quickest approach I hacked together.

DuckDuckGo/AppDelegate+AppDeepLinks.swift
```diff
@@ -22,6 +22,7 @@ import Core
 
 extension AppDelegate {
 
+    // swiftlint:disable:next cyclomatic_complexity
```
samsymons, line 25: I don't usually like to disable this one, but this function is specifically designed to be a giant switch statement that handles a large number of options, and I don't think it suffers because of it since each option just calls through to a single function.

DuckDuckGo.xcodeproj/project.pbxproj
```diff
@@ -8477,7 +8495,7 @@
 				MTL_ENABLE_DEBUG_INFO = YES;
 				ONLY_ACTIVE_ARCH = YES;
 				SDKROOT = iphoneos;
-				SWIFT_ACTIVE_COMPILATION_CONDITIONS = "DEBUG NETWORK_PROTECTION";
+				SWIFT_ACTIVE_COMPILATION_CONDITIONS = "DEBUG NETWORK_PROTECTION ALPHA";
```
samsymons, line 8537: This is so that I can include the widget in the alpha specifically.

DuckDuckGo/VPNIntents.swift
```diff
@@ -0,0 +1,102 @@
...
+
+            while iterations <= 10 {
+                try? await Task.sleep(interval: .seconds(0.5))
+
+                if manager.connection.status == .disconnected {
+                    return .result()
+                }
+
+                iterations += 1
+            }
```
samsymons, line 54: This is a hack to return from the intent when the VPN status has actually changed - this is necessary since `stopVPNTunnel()` only begins the shutdown process, it doesn't offer a callback for when it's complete. I've capped the maximum wait at 5 seconds, just in case.

samsymons:
⚠️ I just remembered that we don't yet support Xcode 15. I'll try and find a workaround for this - perhaps it can be used for the alpha build only.

DuckDuckGo/NetworkProtectionStatusViewModel.swift
graeme: Yeah, I think there should probably be a new type for this which is observing the status and has a lifecycle tied to the AppDelegate.

DuckDuckGo/VPNIntents.swift
graeme: I’m still wondering whether this could be overcome by setting the connection status to an app-group shared UserDefaults store. Could then just use KVO to observe and use the Combine provided timeout. Hit me up if you want to talk about this idea

PacketTunnelProvider/NetworkProtection/NetworkProtectionPacketTunnelProvider.swift
```diff
@@ -209,6 +213,14 @@ final class NetworkProtectionPacketTunnelProvider: PacketTunnelProvider {
             source.resume()
         }
     }
+
+    private func observeServerChanges() {
```
graeme, line 217: I think we should also utilise the observable UserDefaults pattern with this.

Widgets/VPNWidget.swift
```diff
@@ -0,0 +1,291 @@
...
+        case .disconnecting, .disconnected: return false
+        default: return false
+        }
+    }
+}
+
+@available(iOSApplicationExtension 17.0, *)
+struct VPNStatusView: View {
+    @Environment(\.widgetFamily) var family: WidgetFamily
+    var entry: VPNStatusTimelineProvider.Entry
```
graeme, line 115: I see that a lot of this has a very synchronous data flow. I’m not sure where the limitations lie with that but let’s have a quick chat about it now I’ve seen the PR.

graeme (commented):
I think this looks generally good, but I’d like to try and understand a bit more about what the limitations are for us being able to observe changes from the extension. It might be that there’s still something we can do that hasn’t been tried to make it a bit more reactive.

WidgetsExtensionAlpha.entitlements
```diff
@@ -4,8 +4,9 @@
 <dict>
 	<key>com.apple.security.application-groups</key>
 	<array>
-		<string>group.com.duckduckgo.alpha.database</string>
 		<string>group.com.duckduckgo.alpha.bookmarks</string>
+		<string>group.com.duckduckgo.alpha.database</string>
+		<string>group.com.duckduckgo.netp</string>
```
samsymons: I think this whole file needs to be changed to use `GROUP_ID_PREFIX`.

DuckDuckGo/AppDelegate.swift
```diff
@@ -411,6 +420,10 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         syncService.scheduler.notifyAppLifecycleEvent()
         fireFailedCompilationsPixelIfNeeded()
         refreshShortcuts()
+
+#if NETWORK_PROTECTION
+        widgetRefreshModel.refreshVPNWidget()
```
graeme, line 425: Nice :)

graeme (approved):
Look good! Just need to figure out what to do with the commented out code + also will need to come back to do Localization before shipping to external users. But that can be done in a later task.

PR #2169, opened 2023-11-16
Address Bar Spoofing Tests + Remediation

not-a-rootkit:
Task/Issue URL: https://app.asana.com/0/414235014887631/1205886249660032/f

**Description**:
In an attempt to reduce our exposure to address bar spoofing issues I'd like to introduce these 8 new maestro tests that test against privacy-test-pages for specific known address bar spoofing vulnerabilities. The core issue in iOS involves download links that perform 301/302 HTTP redirects - the address bar is incorrectly updated to the download URL even though the target URL has no associated HTML document, therefore we're left with a stale HTML document and a spoofed address bar. See more information here: https://app.asana.com/0/0/1205809497861069/f

In my proposed fix, we update the omnibar text value to "about:blank" when a file download prompt is shown. This is consistent with most other browsers, and the address bar should be correctly updated when there is a HTML document in the renderer, so it shouldn't impact other file downloads.

**Steps to test this PR**:
(tested on iPhone 14 Pro with iOS 17.0.3 and the simulator, no UI changes so this should be sufficient)
1. run `maestro test .maestro/security_tests/0_all.yaml`
2. check all tests are passing
3. open the browser on https://privacy-test-pages.site/security/address-bar-spoofing/spoof-js-download-url.html
4. tap on "Start"
5. tap on "Cancel"
6. ensure the address bar is still "about:blank" and not "staticcdn.duckduckgo.com..."
7. go back
8. tap on "Start"
9. tap "Save to Downloads"
10. check the address bar is "about:blank" and is not spoofed
11. now check normal downloads work as expected, navigate to https://filesamples.com/formats/bin
12. download the first file
13. check "about:blank" is shown in address bar briefly
14. tap either "Download" or "Cancel"
15. check the address bar value reverts to the correct origin

### DIFF
--- a/.maestro/security_tests/0_all.yaml
+++ b/.maestro/security_tests/0_all.yaml
@@ -1,0 +1,39 @@
+# all.yaml
+
+appId: com.duckduckgo.mobile.ios
+---
+
+# Set up 
+- clearState
+- launchApp
+- runFlow: 
+    when: 
+      visible: 
+        text: "Let’s Do It!"
+        index: 0
+    file: ../shared/onboarding.yaml
+
+# Load Site
+- assertVisible:
+    id: "searchEntry"
+- tapOn: 
+    id: "searchEntry"
+- inputText: "https://privacy-test-pages.site/security/address-bar-spoofing/"
+- pressKey: Enter
+
+# Manage onboarding
+- runFlow:
+    when:
+      visible:
+        text: "Got It"
+        index: 0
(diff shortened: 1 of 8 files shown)


not-a-rootkit:
This PR is trying to merge from a fork which is making the build release step fail. I will create a new one from a branch instead. The new PR is here: https://github.com/duckduckgo/iOS/pull/2181

PR #2176, opened 2023-11-20
removed unarchiveTopLevelObjectWithData deprecation

federicocappelli:
Task/Issue URL: https://app.asana.com/0/414709148257752/1205990984763603/f

**Description**:

I replaced the deprecated (iOS 12) API `unarchiveTopLevelObjectWithData` maintaining the use of non-secure coding in order to achieve backward compatibility with the already saved Tabs.

All functions that can throw errors are now correctly handled and in case of error a log is added.

**Steps to test this PR**:
1. Install the current app and save tabs
2. Update the new app version and check if the tabs are still there.

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPhone 14 Pro
* [ ] iPad

**OS Testing**:

* [ ] iOS 15
* [ ] iOS 16
* [ ] iOS 17

### DIFF
--- a/DuckDuckGo/TabsModelPersistenceExtension.swift
+++ b/DuckDuckGo/TabsModelPersistenceExtension.swift
@@ -18,6 +18,8 @@
 //
 
 import Foundation
+import Core
+import Common
 
 extension TabsModel {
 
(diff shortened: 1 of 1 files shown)


DuckDuckGo/TabsModelPersistenceExtension.swift
```diff
@@ -26,17 +28,34 @@ extension TabsModel {
...
     }
 
     public static func clear() {
          UserDefaults.app.removeObject(forKey: Constants.key)
     }
 
     func save() {
-        guard let data = try? NSKeyedArchiver.archivedData(withRootObject: self, requiringSecureCoding: false) else { return }
-        UserDefaults.app.set(data, forKey: Constants.key)
+        var data: Data
```
brindy: This is only used inside the block so might as well be defined there and can also be a `let`

brindy (approved):
Appears to work.  Just one minor comment, then feel free to merge.

DuckDuckGo/TabsModelPersistenceExtension.swift
federicocappelli: agree, fixed!

PR #2180, opened 2023-11-22
update theme to use system colours

brindy:
Task/Issue URL: https://app.asana.com/0/0/1206009605456322/f
Tech Design URL:
CC:

**Description**:

Rationalises some of the colours to use the system design and some minor UI fixes.

**Steps to test this PR**:
1. Launch the app on iPad and iPhone and test in Light and Dark mode
2. Smoke test the app (browsing, tabs, bookmarks, autofill, settings) and check that nothing jumps out as wrong


DuckDuckGo/AutofillLoginSettingsListViewController.swift
```diff
@@ -552,6 +552,21 @@ extension AutofillLoginSettingsListViewController: UITableViewDelegate {
             return 0
         }
     }
+
+    func tableView(_ tableView: UITableView, willDisplayHeaderView view: UIView, forSection: Int) {
+        if let view = view as? UITableViewHeaderFooterView {
+            let theme = ThemeManager.shared.currentTheme
+            view.textLabel?.textColor = theme.tableHeaderTextColor
```
THISISDINOSAUR, line 559: It would be nice to be able to move away from needing the theme thing, but I don't have a good sense of what's holding us back there. Is there anything that necessitates it versus just being a lot of work for us to get rid of it?

THISISDINOSAUR (approved):
LGTM!

Not really related, but when did we stop using yellow?
Screenshot 2023-11-22 at 19 04 34

DuckDuckGo/AutofillLoginSettingsListViewController.swift
brindy: I think once we have migrated all the theme colours to the design system we could discuss dropping the theme entirely.  The only advantage of keeping it is if we want to do custom themes, but even then because we've been using the design system colours directly in places it'd be a lot of work.

brindy:
> when did we stop using yellow?

Oh, I think during the ship review of the sync related bookmark changes, but it's yellow when you swipe right to add/remove favorite.

PR #2184, opened 2023-11-22
bookmarks migration refactor and health monitoring

brindy:
Task/Issue URL: https://app.asana.com/0/0/1205887954649229/f
Tech Design URL:
CC:

**Description**:

Refactor bookmarks migration and fire pixel if migration happens multiple times.

**Steps to test this PR**:
1.  Launch a previous version of the app and add bookmarks and favorites
2. Run this version and ensure that bookmarks and favorites are still there
3. Clean install of app ensure it launches as expected and saving bookmarks/favorites works
4. Check sync related functionality
5. After running the app once hardcode `migrationHappened` to true - should fire a pixel and assert (might not see pixel because of assert though)


DuckDuckGo/AppDelegate.swift
```diff
@@ -166,48 +156,15 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
-                Thread.sleep(forTimeInterval: 1)
-                fatalError("Could not prepare Bookmarks DB structure")
-            }
-
-            WidgetCenter.shared.reloadAllTimelines()
+        if BookmarksDatabaseSetup.loadStoreAndMigrate(bookmarksDatabase: bookmarksDatabase,
+                                                      crashOnLoadStoreError: !shouldPresentInsufficientDiskSpaceAlertAndCrash) {
+            // MARK: post-Bookmarks migration logic
+            assertBookmarksMigrationHappensOnlyOnce()
+            shouldResetBookmarksSyncTimestamp = true
```
brindy, line 164: I noticed that "syncDataProviders" can only be nil here because this immediately after launch before a value has been assigned, so I just reduced this to the logic that would be called anyway.

DuckDuckGo/AppDelegate.swift
```diff
@@ -166,48 +156,15 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
         }
 
+        WidgetCenter.shared.reloadAllTimelines()
```
brindy, line 167: This doesn't need to be called from the loadStore because it's gonna happen anyway.

DuckDuckGo/BookmarksDatabaseSetup.swift
```diff
@@ -0,0 +1,79 @@
...
+import CoreData
+import Core
+import Bookmarks
+import Persistence
+
+struct BookmarksDatabaseSetup {
+
+    static func loadStoreAndMigrate(bookmarksDatabase: CoreDataDatabase, crashOnLoadStoreError: Bool) -> Bool {
+        var migrationHappened = false
+        bookmarksDatabase.loadStore { context, error in
```
brindy, line 30: The idea here was 1 line per action related to migrations.

DuckDuckGo/BookmarksDatabaseSetup.swift
```diff
@@ -0,0 +1,79 @@
...
+            guard let context = assertContext(context, error: error, crashOnError: crashOnLoadStoreError) else { return }
+            self.migrateFromFirstVersionOfBookmarksCoreDataStorage(context)
+            migrationHappened = self.migrateToFormFactorSpecificFavorites(context)
+            // Add future bookmarks database migrations here and set boolean to result of whatever the last migration returns
```
brindy: The idea here is 1 line per action related to migrations.

DuckDuckGo/AppDelegate.swift
```diff
@@ -354,6 +286,18 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         return true
     }
 
+    private func assertBookmarksMigrationHappensOnlyOnce() {
+        let key = UserDefaultsWrapper<Any>.Key.bookmarksLastGoodVersion.rawValue
+        if let lastGoodVersion = UserDefaults.standard.string(forKey: key) {
+            Pixel.fire(pixel: .debugBookmarksLost, withAdditionalParameters: [
+                PixelParameters.bookmarksLastGoodVersion: lastGoodVersion
+            ])
+            assertionFailure("Unexpected bookmarks migration")
```
ayoy, line 295: When migrating from a pre-FFS-favorites version, I'm getting this assertion on every app launch. This is because [this migration code](https://github.com/duckduckgo/BrowserServicesKit/blob/main/Sources/Bookmarks/Migrations/BookmarkFormFactorFavoritesMigration.swift#L65-L78) is apparently being run every time with a non-nil `orderedUUIDs` array. This means we have to fix the migration code in BSK, before we merge this PR, and we probably want to do it quick (e.g. on Monday). I'll create a task to track it.

PR #2209, opened 2023-11-30
NetP waitlist final touches

samsymons:
Task/Issue URL: https://app.asana.com/0/0/1206027185389966/f
Tech Design URL:
CC:

**Description**:

This PR makes some final touches for the NetP waitlist, including:

* Adding RMF support
* Adding pixels
* Updates NetP to only be visible to US users (unless they're internal)
* Recording the date at which the user activated NetP, in order to show messages remotely if necessary

**Steps to test this PR**:

This PR is using a custom RMF URL which matches NetP users in the US who have used the feature for at least 3 days. To test this properly may require overriding some of these values.

To override the matching values, see the `fetchAndProcess` function inside `RemoteMessaging.swift`. The `isNetworkProtectionWaitlistUser` and `daysSinceNetworkProtectionEnabled` variables should be changed to test various conditions.

Note that RMF only processes messages once a day - if you change the values above, this will not necessarily change what you see. To override this, go to `RemoteMessagingConfigProcessor.swift` and change `shouldProcessConfig` to always return true.

1. After following the instructions above, you should begin to get a remote message visible in the app; check that you do, and that the URL is opened with survey parameters included
2. Run through the waitlist flow and check that pixels get fired along the way - note that these are daily pixels, we unfortunately have no way to reset these right now except for uninstalling and reinstalling the app to clear user defaults

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPhone 14 Pro
* [ ] iPad

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:


DuckDuckGo/NetworkProtectionStatusViewModel.swift
```diff
@@ -106,6 +109,13 @@ final class NetworkProtectionStatusViewModel: ObservableObject {
             .map(Self.statusImageID(connected:))
             .assign(to: \.statusImageID, onWeaklyHeld: self)
             .store(in: &cancellables)
+        isConnectedPublisher
```
graeme: Possibly nitpicking, but this is not really the best place to put this. If the user taps connect and then navigates off the screen before it completes, the connection won’t get registered. I know that might be a statistically irrelevant number of cases, but it more makes the point that this is possibly not the best place to put this.

You could alternatively put this into the `NetworkProtectionPacketTunnelProvider` as that’s really the source of truth for when it’s activated.

graeme (commented):
This looks good, but lemme know if/when you fix the survey link issue.

graeme:
Agreed that this can be merged once we figure out the survey link issue. Can also potentially fix in a later PR if you don’t want it to block the rest.

graeme (approved):

PR #2230, opened 2023-12-06
Use new WKWebsiteDataStore API for tabs [needs Xcode 15]

brindy:
Task/Issue URL: https://app.asana.com/0/0/1205887954649226/f
Tech Design URL:
CC: @bwaresiak 

**Description**:
Move to use new WKWesbiteDataStore API for creating a persistent container. 

**Steps to test this PR**:

Using a device:
1. Using `develop` sign into at least one website (e.g. Netflix) and fireproof it 
2. Use https://www.w3schools.com/js/tryit.asp?filename=tryjs_cookie_username to set a cookie
3. Open a new tabs on various websites
4. Check out this branch and run the app
5. Check you're still logged into the website, the cookie is still set and all the tabs are still present
6. Use the fire button 
7.  Check you're still logged into the website, but the cookie has been cleared
8. Use the fire button again
9. Open and close some tabs
10. Using the tab switcher add a new tab and use the fire button
11. Remove the domain from the list of fire proofed sites (in settings) and refresh the webview with that site, you should now be logged out

Using the simulator:
1. Check that the containers are deleted from disk
* see Xcode console for location of app on disk using the simulator
* add a breakpoint to `DataStoreIdManager` in `WKWebViewConfigurationExtension.swift` to get the container id.  
* Note that the containerId on disk is in **lowercase**.  
* Use a command like `find <path to app> | grep <container id lowercased>` before and after the fire button


Core/WebCacheManager.swift
```diff
@@ -139,30 +151,124 @@ public class WebCacheManager {
...
+
+            let uuids = await WKWebsiteDataStore.allDataStoreIdentifiers
+            for uuid in uuids {
+                do {
+                    try await WKWebsiteDataStore.remove(forIdentifier: uuid)
+                } catch {
+                    Pixel.fire(pixel: .debugCouldNotRemoveWebsiteDataStore, error: error)
+
+                    // Not a fatal error because it will get deleted next time
+                    assertionFailure("Failed to remove datastore")
```
miasma13: During testing on simulator I was hit it with the following error:
`Error Domain=WKWebSiteDataStore Code=1 "Data store is in use (by network process)" UserInfo={NSLocalizedDescription=Data store is in use (by network process)}`

miasma13: And: `Error Domain=WKWebSiteDataStore Code=1 "Data store is in use" UserInfo={NSLocalizedDescription=Data store is in use}`

miasma13 (changes requested):
Great job. Here are couple of comments:
1. During testing it was very frequent for me to hit both of the asserts on fire button so maybe replacing them to some less aggressive but explicit log would be better for us developing.
2. As mentioned in the comments multiple stores scenario is not an issue as we create a new one so the old ones will eventually get purge later on but what is more important they won't be used by the app so we minimise the cookie leaking to zero.
3. Found one issue: when triggering fire button clearing twice (quickly one after another) the fireproofed cookies get cleared resulting in user being signed out from these websites. Can you give it a look?

brindy:
1. That's disappointing but yeah it is fairly easy to make it happen. I looked at the memory graph and can see the datastore(s) but they are not connect to anything in our code base directly or indirectly, it's like the WKWebView and other classes just hang around for a bit.  To me that feels a like a bug because there's no way to explicitly dispose of that memory.  Not sure I have the energy to open a feedback item that'll just get ignored though.  WDYT?
2. Yep, am going to change that code slightly so that it does a `try?` instead. There's a subsequent check immediately after it anyway which can fire a pixel if there's any unexpected datastores around. 
3. On it.

brindy:
3. Super annoying. Doesn't matter how long you wait - if you fire button twice in a row the cookies are lost.  Let's have a chat about it next week.

brindy:
@miasma13 - I ended up removing some legacy code and repurposing CookieStorage to store the cookies in rather than putting them directly into the new container. I would rather not have to do that but seems to be the only way. :( Let me know what you think.

Core/WebCacheManager.swift
```diff
@@ -182,7 +182,7 @@ public class WebCacheManager {
 
             storeIdManager.allocateNewContainerId()
             // If cookies is empty it's likely that the webview was not used since the last fire button so
-            //  don't overwrite previously saved cookes
+            //  don't overwrite previously saved cookies
             if let cookies, !cookies.isEmpty {
                 cookeStorage.cookies = cookies
```
miasma13: And the `coke storage` here.

miasma13 (approved):
Great job 👍

PR #2236, opened 2023-12-06
bookmarks migration refactor

brindy:
Task/Issue URL: https://app.asana.com/0/0/1205887954649229/f
Tech Design URL:
CC:

**Description**:

Refactor bookmarks migration and fire pixel if migration happens multiple times.

**Steps to test this PR**:
1.  Launch old version of the app (before form factor migration) and add bookmarks and favorites
2. Run this version and ensure that bookmarks and favorites are still there
3. Clean install of app ensure it launches as expected and saving bookmarks/favorites works
4. Check sync related functionality
5. After running the app once hardcode `migrationHappened` to true - should fire a debug pixel and assert (might not see pixel because of assert though)


ayoy (approved):
LGTM! I also tested migration from 7.96.0 and it worked great!

PR #2263, opened 2023-12-13
Run sync tests in matrix

loremattei:
Task/Issue URL: https://app.asana.com/0/1201493110486074/1206174678008392/f
Tech Design URL:
CC:

**Description**:
This PR refactor Sync e2e tests code and let the tests run in parallel on multiple iOS versions. 

**Steps to test this PR**:
1. Check CI is green

---
###### Internal references:

### DIFF
--- a/.github/workflows/sync-end-to-end.yml
+++ b/.github/workflows/sync-end-to-end.yml
@@ -1,6 +1,7 @@
 name: Sync-End-to-End tests
 
 on:
+  push:
   schedule:
     - cron: '0 5 * * *' # run at 5 AM UTC
 
--- a/.github/workflows/sync-end-to-end.yml
+++ b/.github/workflows/sync-end-to-end.yml
@@ -1,7 +1,6 @@
 name: Sync-End-to-End tests
 
 on:
-  push:
   schedule:
     - cron: '0 5 * * *' # run at 5 AM UTC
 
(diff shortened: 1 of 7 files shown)


.github/workflows/sync-end-to-end.yml
```diff
@@ -43,62 +43,74 @@ jobs:
...
+        path: |
+          xcodebuild.log
+          DerivedData/Logs/Test/*.xcresult
+        retention-days: 7
+    
+  sync-end-to-end-tests:
+    name: Sync End To End Tests
+    needs: build-for-sync-end-to-end-tests
+    runs-on: macos-13
+    strategy:
```
ayoy, line 69: There's one more argument that we can pass to `strategy` which is `fail-fast: false`. By default it's `true` and whenever any of the concurrent jobs fails, all others that are still in progress are cancelled. When we set it to `false`, all concurrent jobs will run to finish (or error) without looking at other jobs. I think it would be useful for us to check whether a failure happens on a specific iOS version. What do you think @loremattei?

```suggestion
    strategy:
      fail-fast: false
```

ayoy (approved):
The matrix works great! I took the liberty to update upload/download actions to the recently released v4 which is much much faster. Thanks for the patch @loremattei!

loremattei:
Thank you @ayoy !

PR #2330, opened 2024-01-11
fix page position on navigation

brindy:
Task/Issue URL: https://app.asana.com/0/715106103902962/1205688349294686/f
Tech Design URL:
CC:

**Description**:
Fixes the following bugs:

* page position would be forgotten on navigation
* scrolling would stop when user scrolls the webview and the bars are being hidden

**Steps to test this PR**:

On both iPhone and iPad, ideally on a device:

* Browse the web and use backwards and forwards navigation.  The page position should not change on navigation.
* Check different orientations
* Check keyboard adjusts the webview correctly as needed (when entering fields on the page)
* Check the browsing menu appears in the expected location 

On iPhone specifically (or iPad when in small mode on split view):

* Check browsing with top and bottom address bar positions (bottom bar position not available on iPad unless in “small” mode / split screen)
* Check scrolling hides and shows the bar as previous, but does not stop the page scrolling
* Quick swipe should hide show bars as previous
* When at bottom of the screen over scrolling should show bars as previous


DuckDuckGo/BarsAnimator.swift
```diff
@@ -84,10 +84,6 @@ class BarsAnimator {
         let ratio = calculateTransitionRatio(for: scrollView.contentOffset.y)
         delegate?.setBarsVisibility(1.0 - ratio, animated: false)
         transitionProgress = ratio
-
-        var offset = scrollView.contentOffset
-        offset.y = transitionStartPosY
-        scrollView.contentOffset = offset
```
brindy, line 90: adjusting the content offset while it was scrolling is what causes the page to stop scrolling while the bars are being hidden, which is not what other browsers do and at least one person internally has complained about

DuckDuckGo/BarsAnimator.swift
```diff
@@ -139,7 +131,12 @@ class BarsAnimator {
         let cumulativeDistance = (barsHeight * transitionProgress) + distance
         let normalizedDistance = max(cumulativeDistance, 0)
 
-        return min(normalizedDistance / barsHeight, 1.0)
+        // We used to fix the scroll position in place as the transition happened
+        //  but now the bars disappear too. This adjusts for that.
+        let transitionSpeed = 0.5
```
brindy, line 136: This is related to the content offset code removal above.

DuckDuckGo/BrowserChromeManager.swift
```diff
@@ -38,7 +38,6 @@ protocol BrowserChromeDelegate: AnyObject {
 class BrowserChromeManager: NSObject, UIScrollViewDelegate {
 
     struct Constants {
-        static let dragThreshold: CGFloat = 30
```
brindy, line 41: unused

DuckDuckGo/BrowsingMenu/BrowsingMenuAnimator.swift
```diff
@@ -61,7 +61,7 @@ final class BrowsingMenuAnimator: NSObject, UIViewControllerAnimatedTransitionin
             transitionContext.containerView.addSubview(fromSnapshot)
         }
 
-        toViewController.bindConstraints(to: fromViewController.currentTab?.webView)
+        toViewController.bindConstraints(to: fromViewController.viewCoordinator.contentContainer)
```
brindy, line 64: the web view can now be as big as the app so is the wrong thing to base the browsing menu on

DuckDuckGo/BrowsingMenu/BrowsingMenuViewController.swift
```diff
@@ -166,12 +166,12 @@ final class BrowsingMenuViewController: UIViewController {
         }
     }
 
-    private weak var webView: UIView?
-    private var webViewFrameObserver: NSKeyValueObservation?
-    func bindConstraints(to webView: UIView?) {
-        self.webView = webView
-        self.webViewFrameObserver = webView?.observe(\.frame, options: [.initial]) { [weak self] webView, _ in
-            self?.recalculateMenuConstraints(with: webView)
+    private weak var guideView: UIView?
```
brindy, line 169: since it's no longer a web view now, I've renamed it to match its purpose

DuckDuckGo/MainViewController.swift
```diff
@@ -137,6 +137,8 @@ class MainViewController: UIViewController {
     
     // Skip SERP flow (focusing on autocomplete logic) and prepare for new navigation when selecting search bar
     private var skipSERPFlow = true
+        
+    private var keyboardHeight: CGFloat = 0.0
```
brindy, line 141: Set when the keyboard frame changes so that it can be used when refreshing the content insets for the web view

DuckDuckGo/MainViewController.swift
```diff
@@ -872,18 +876,29 @@ class MainViewController: UIViewController {
         }
     }
 
-    private func addToView(tab: TabViewController) {
+    private func addToWebViewContainer(tab: TabViewController) {
```
brindy, line 879: renamed to reflect purpose instead of unclear overloaded names

DuckDuckGo/MainViewController.swift
```diff
@@ -872,18 +876,29 @@ class MainViewController: UIViewController {
...
+        viewCoordinator.logoContainer.isHidden = true
+        viewCoordinator.contentContainer.isHidden = true
+        
         tab.progressWorker.progressBar = viewCoordinator.progress
         chromeManager.attach(to: tab.webView.scrollView)
         tab.chromeDelegate = self
     }
 
-    private func addToView(controller: UIViewController) {
+    private func addToContentContainer(controller: UIViewController) {
```
brindy, line 900: renamed to reflect purpose

DuckDuckGo/MainViewController.swift
```diff
@@ -872,18 +876,29 @@ class MainViewController: UIViewController {
...
     }
 
-    private func addToView(tab: TabViewController) {
+    private func addToWebViewContainer(tab: TabViewController) {
         removeHomeScreen()
         updateFindInPage()
         currentTab?.progressWorker.progressBar = nil
         currentTab?.chromeDelegate = nil
-        addToView(controller: tab)
+        currentTab?.webView.scrollView.contentInsetAdjustmentBehavior = .never
```
brindy, line 884: iOS does weird stuff if this not disabled, once you start messing with the content insets

DuckDuckGo/TabViewController.swift
```diff
@@ -1147,7 +1147,12 @@ extension TabViewController: WKNavigationDelegate {
         DispatchQueue.main.async { [weak self] in
             guard let webView = self?.webView,
                   webView.bounds.height > 0 && webView.bounds.width > 0 else { completion(nil); return }
-            UIGraphicsBeginImageContextWithOptions(webView.bounds.size, false, UIScreen.main.scale)
+            
+            let size = CGSize(width: webView.frame.size.width,
+                              height: webView.frame.size.height - webView.scrollView.contentInset.top - webView.scrollView.contentInset.bottom)
+            
+            UIGraphicsBeginImageContextWithOptions(size, false, UIScreen.main.scale)
+            UIGraphicsGetCurrentContext()?.translateBy(x: 0, y: -webView.scrollView.contentInset.top)
```
brindy, line 1155: we're only interested in the bit of the screen currently visible - without this we end up with potentially blank space

brindy:
Just did a bit of a self review and pushed a couple of commits.  I'm definitely done now @THISISDINOSAUR when you're ready.  Thanks!

DuckDuckGo/BarsAnimator.swift
THISISDINOSAUR: Do you know why we were doing this?

brindy: Not really, but it's been flagged as a bug since and does seem inconsistent with other browsers and personally I think it is a bit annoying :)

THISISDINOSAUR (approved):
LGTM! 
The change with scrolling when the URL bar is at the bottom is way better.

I haven't been able to get it to forget the scroll position, but that bug was always really hard to replicate. It at least doesn't seem worse 😅

brindy:
Thank you!

PR #2343, opened 2024-01-15
Add NetP feedback form

quanganhdo:
Task/Issue URL: https://app.asana.com/0/72649045549333/1206151925871430/f

**Description**:

Implements the native feedback form for NetP and collect necessary metadata. 

**Steps to test this PR**:

1. Go to Settings > Network Protection > Share Feedback
2. Test out the form
3. Check the submitted metadata against https://app.asana.com/0/0/1206198549247475/f

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPhone 14 Pro
* [ ] iPad

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

—
###### Internal references:


DuckDuckGo/Feedback/VPNFeedbackSender.swift
```diff
@@ -0,0 +1,47 @@
...
+import Core
+
+protocol VPNFeedbackSender {
+    func send(metadata: VPNMetadata, category: VPNFeedbackCategory, userText: String) async throws
+}
+
+struct DefaultVPNFeedbackSender: VPNFeedbackSender {
+
+    func send(metadata: VPNMetadata, category: VPNFeedbackCategory, userText: String) async throws {
+        let encodedUserText = userText.addingPercentEncoding(withAllowedCharacters: .alphanumerics) ?? userText
```
samsymons: I consider this one a bug in the macOS implementation - I recently fixed it. Instead, we should use the following character set, which avoids overly encoding the feedback:

```suggestion
        let encodedUserText = userText.addingPercentEncoding(withAllowedCharacters: .alphanumerics.union(.init(charactersIn: "-._~"))) ?? userText
```

DuckDuckGo/Feedback/VPNMetadataCollector.swift
```diff
@@ -0,0 +1,240 @@
...
+
+        let networkPathChange = VPNSettings(defaults: .networkProtectionGroupDefaults).networkPathChange
+        let now = Date()
+
+        while true {
+            if !monitor.currentPath.availableInterfaces.isEmpty {
+                path = monitor.currentPath
+                monitor.cancel()
+
+                let changeDate = networkPathChange?.date ?? .distantPast
```
samsymons: It's possible to get `.distantPast` if you submit feedback right after starting NetP for the first time and not yet changing networks - I could see the setup flow being a source of feedback, what do you think about using some other value here?

I was thinking that `secondsSincePathChange` could become be converted from int to string, and in the result of no path change then it can just say `never changed` or something.

DuckDuckGo/NetworkProtectionTunnelController.swift
```diff
@@ -24,6 +24,7 @@ import Combine
 import Core
 import NetworkExtension
 import NetworkProtection
+import Common
```
samsymons: Is this required? (I had to do the same thing with a few macOS classes lately, so this is a genuine question 😄)

DuckDuckGo/Feedback/VPNMetadataCollector.swift
```diff
@@ -0,0 +1,240 @@
...
+        let connectedServer = serverInfoObserver.recentValue.serverLocation ?? "none"
+        let connectedServerIP = serverInfoObserver.recentValue.serverAddress ?? "none"
+
+        return .init(connectionState: connectionState,
+                     lastDisconnectError: await lastDisconnectError(),
+                     connectedServer: connectedServer,
+                     connectedServerIP: connectedServerIP)
+    }
+
+    private func lastDisconnectError() async -> String {
```
samsymons, line 198: Nice :+1:

DuckDuckGo/NetworkProtectionTunnelController.swift
quanganhdo: Good catch, in this instance it’s not necessary at all :D

DuckDuckGo/Feedback/VPNMetadataCollector.swift
quanganhdo: Makes sense. I’d change that to report “unknown” if changeDate is nil 👍

samsymons (approved):
This is working really nicely, thanks so much!

PR #2349, opened 2024-01-17
support multiple passes

brindy:
Task/Issue URL: https://app.asana.com/0/414709148257752/1206364468723840/f
Tech Design URL:
CC:

**Description**:

Adds support for handling multiple PKPASS items at once. 

**Steps to test this PR**:
1. Visit the URL https://chrisbrind.rocks/example.pkpasses - should trigger to passes to be prompted to add to your wallet
2. Visit https://www.passsource.com/pass/register.php?hashedSerialNumber=eNortjIysVKyzLFMNPGyqPByzLZMSSlJDfAJNk9Lt7VVsgZcMJnACXc,&psRegisteredFlag=1&psMessage=& and click on the "add to wallet" button - should trigger a single pass to be added to your wallet

---
###### Internal references:

### DIFF
--- a/DuckDuckGo/PassKitPreviewHelper.swift
+++ b/DuckDuckGo/PassKitPreviewHelper.swift
@@ -21,6 +21,8 @@ import Common
 import UIKit
 import PassKit
 
+import ZIPFoundation
+
 class PassKitPreviewHelper: FilePreview {
     private weak var viewController: UIViewController?
     private let filePath: URL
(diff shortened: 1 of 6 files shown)


DuckDuckGo/PassKitPreviewHelper.swift
```diff
@@ -42,3 +44,45 @@ class PassKitPreviewHelper: FilePreview {
         }
     }
 }
+
+class ZippedPassKitPreviewHelper: FilePreview {
```
federicocappelli: Why not create this in a new file?

DuckDuckGo/PassKitPreviewHelper.swift
```diff
@@ -42,3 +44,45 @@ class PassKitPreviewHelper: FilePreview {
...
+        self.viewController = viewController
+    }
+    
+    func preview() {
+        if let passes: [PKPass] = extractDataFromZip(at: self.filePath)?.compactMap({ try? PKPass(data: $0) }),
+           passes.count > 0,
+           let controller = PKAddPassesViewController(passes: passes) {
+            viewController?.present(controller, animated: true)
+        } else {
+            os_log("Can't present passkit: No passes in passes file", type: .debug)
```
federicocappelli: I would log this as .error instead of .debug

DuckDuckGo/PassKitPreviewHelper.swift
```diff
@@ -42,3 +44,45 @@ class PassKitPreviewHelper: FilePreview {
...
+                _ = try archive.extract(entry, skipCRC32: true) { data in
+                    passData.append(data)
+                }
+                
+                if passData.count > 0 {
+                    dataObjects.append(passData)
+                }
+            }
+        } catch {
+            os_log("Error reading pkpasses file: %s", type: .debug, error.localizedDescription)
```
federicocappelli: Same here, why logging error as .debug and not .error?

DuckDuckGo/PassKitPreviewHelper.swift
```diff
@@ -42,3 +44,45 @@ class PassKitPreviewHelper: FilePreview {
...
+        if let passes: [PKPass] = extractDataFromZip(at: self.filePath)?.compactMap({ try? PKPass(data: $0) }),
+           passes.count > 0,
+           let controller = PKAddPassesViewController(passes: passes) {
+            viewController?.present(controller, animated: true)
+        } else {
+            os_log("Can't present passkit: No passes in passes file", type: .debug)
+        }
+    }
+ 
+    func extractDataFromZip(at zipPath: URL) -> [Data]? {
```
federicocappelli: This function interface is a bit ambiguous:
-  I would avoid using nil returns as errors. An optional result should be used only in cases where it's really a valid result
- To show failures I would allow the `func` to throw errors, and manage the errors in the caller.
- It's not clear if the `zipPath: URL` can be a local file, a remote URL or both, could you add some standard comments and/or change the param name?

federicocappelli (changes requested):
Few comments added, please ping me for any clarification

DuckDuckGo/PassKitPreviewHelper.swift
brindy: following what was already in the file. will update them all

DuckDuckGo/PassKitPreviewHelper.swift
brindy: My thinking is:
* new file isn't really that useful (especially if you use Xcode to navigate the symbols)
* two highly related things are right next to each other

Don't feel too strongly though

federicocappelli (approved):
LGTM

PR #2357, opened 2024-01-19
Display short URL when not editing

dus7:
Task/Issue URL: https://app.asana.com/0/1206226850447395/1205485701057249

**Description**:
Displays a short URL (hostname without `www` component), while browsing. While short URL is displayed, the text alignment is adjusted to the center.

**Steps to test this PR**:
1. Open a new tab
2. Observe leading side address bar placeholder alignment
3. Browse to any URL (except DDG SERP)
4. Observe short URL in the address bar, vertically centered
5. Start editing
6. Observe complete URL, selected, aligned to the leading side

**Theme Testing**:
* [ ] Light theme
* [ ] Dark theme


DuckDuckGo/AddressDisplayHelper.swift
```diff
@@ -0,0 +1,76 @@
...
+            if !isEditing, let shortAddress = shortURLString(url) {
+                return NSAttributedString(
+                    string: shortAddress,
+                    attributes: [.foregroundColor: ThemeManager.shared.currentTheme.searchBarTextColor])
+            } else {
+                return deemphasisePath(forUrl: url)
+            }
+        }
+
+        static func deemphasisePath(forUrl url: URL) -> NSAttributedString {
```
brindy: I'm not sure we need this now do we?  The point of this was to make everything after the first slash slightly lighter but we're not showing that any more.  When the user is editing it should all be the same colour anyway imo (I don't think anyone expects the text to be demphasised while editing).

brindy (changes requested):
Works good, but I don't think we need the de-emphasis anymore.  That's only shown when editing and I don't think anyone expects that while editing.

DuckDuckGo/AddressDisplayHelper.swift
dus7: I had mixed feelings about this part too, but ended up leaving it as it is since the text appearance was not updated while editing - that made me believe it’s intentional behavior.

dus7: Removed: https://github.com/duckduckgo/iOS/pull/2357/commits/84be4eae2ecc6b2c67b8d93acf621f47194e4607

brindy (approved):
LGTM :)

PR #2370, opened 2024-01-23
swipe tabs

brindy:
Task/Issue URL: https://app.asana.com/0/72649045549333/1205842534063579/f
Tech Design URL:
CC:

**Description**:
Add swipe tabs functionality. 

**Steps to test this PR**:

***Internal User***:
1. This feature is behind the internal user flag.  When this is off then swipe tabs should not work on any device.

***Phone***:
1. When on Home Screen with no tabs: no scrolling
2. When on a single tab on a web page: swiping left will launch a new tab.  Note that if there are favourites they will not be shown until a 'preview' of the tab has been captured, so will show Dax the first time you swipe to a new tab
3. Swiping between pages will show a preview if one has been captured.  Web pages not loaded yet will start to load (same as if opening from the tab switcher).
4. Switching orientation will scale the preview, but won't align it with the scroll position on the page in the other orientation.
5. Check the fire button
6. Check opening a new tab from long press the tab button
7. Check opening a new tab from the tab switcher

***iPad***:
1. Swipe tabs will only show when in the "iPhone view" - e.g. the small split screen view.  
2. When in iPad view no swiping should work and the tabs bar should be visible

***Themes***:
1. Check light and dark mode.

***OS Version***:
1. Should work for all supported OS. Ideally check 14, 15, 16 and 17 (I've tested 16 and 17 as don't have device/simulator for 14/15 - ping me if it's a problem and I'll get them).


Core/UIViewExtension.swift
```diff
@@ -73,4 +73,17 @@ extension UIView {
...
+    public func createImageSnapshot(inBounds bounds: CGRect? = nil) -> UIImage? {
+        let bounds = bounds ?? self.frame
+        let size = bounds.size
+        UIGraphicsBeginImageContextWithOptions(size, false, UIScreen.main.scale)
+        UIGraphicsGetCurrentContext()?.translateBy(x: -bounds.origin.x, y: -bounds.origin.y)
+        drawHierarchy(in: frame, afterScreenUpdates: true)
+        let image = UIGraphicsGetImageFromCurrentImageContext()
+        UIGraphicsEndImageContext()
+        return image
+    }
```
samsymons, line 87: Any chance the [`snapshotView` API](https://developer.apple.com/documentation/uikit/uiview/1622531-snapshotview) can be used here instead?

DuckDuckGo/SwipeTabsCoordinator.swift
```diff
@@ -0,0 +1,359 @@
...
+        omniBar?.updateOmniBarPadding(left: left, right: right)
+    }
+    
+}
+
+extension TabsModel {
+    
+    func safeGetTabAt(_ index: Int) -> Tab? {
+        guard tabs.indices.contains(index) else { return nil }
+        return tabs[index]
```
samsymons, line 356: Would it make sense to have an assertionFailure in here so that we can catch when it comes up? Or is this expected to happen in legitimate usage?

Core/UIViewExtension.swift
brindy: I tried that and wasn't getting anywhere with it (because the offset of the view matters and it wasn't clear how to use the snapshot API like that).  But I'll give it another shot as it'll be faster.

DuckDuckGo/SwipeTabsCoordinator.swift
brindy: I'm assuming I can get a nil tab here tbh.

samsymons:
@brindy One issue I'm running into today is a pretty persistent flickering when changing tabs:

https://github.com/duckduckgo/iOS/assets/183774/a8063bd3-0f55-4c78-a300-68408346388c

Core/UIViewExtension.swift
brindy: Would require quite a bit of code to change tbh.  Would rather come back to this as an optimisation.

samsymons: Sounds good :+1:

samsymons (approved):
Besides the flickering issue I mentioned, this is all behaving well. I found a nasty memory leak, but was able to reproduce it on main so I'll raise it elsewhere - it's not introduced by this PR.

Code changes seem sensible and I've not been able to break anything on any device with various iOS versions. Let's merge it and give it a go internally. Nice work!

PR #2381, opened 2024-01-24
fix iPad tabs crash

brindy:
Task/Issue URL: https://app.asana.com/0/414235014887631/1206390999092189/f
Tech Design URL:
CC:

**Description**:

Add some range guard for tab selection.

**Steps to test this PR**:
1. Launch iPad
2. Add a bunch of tabs
3. Open + close, a few times
4. Open enough to scroll
5. Long press a tab - no crash should occur.  If assertion occurs, capture the stack trace.

To force the issue update TabBarViewController.swift and change the following line:

> delegate?.tabsBar(self, didSelectTabAtIndex: path.row + 1)

to 

> delegate?.tabsBar(self, didSelectTabAtIndex: path.row)

Then long press the last tab - an assertion should occur.

### DIFF
--- a/DuckDuckGo/MainViewController.swift
+++ b/DuckDuckGo/MainViewController.swift
@@ -857,8 +857,13 @@ class MainViewController: UIViewController {
     func select(tabAt index: Int) {
         viewCoordinator.navigationBarContainer.alpha = 1
         allowContentUnderflow = false
-        let tab = tabManager.select(tabAt: index)
-        select(tab: tab)
+        
+        if tabManager.model.tabs.indices.contains(index) {
+            let tab = tabManager.select(tabAt: index)
+            select(tab: tab)
+        } else {
+            assertionFailure("Invalid index selected")
+        }
     }
 
     fileprivate func select(tab: TabViewController) {


DuckDuckGo/MainViewController.swift
```diff
@@ -857,8 +857,13 @@ class MainViewController: UIViewController {
...
         viewCoordinator.navigationBarContainer.alpha = 1
         allowContentUnderflow = false
-        let tab = tabManager.select(tabAt: index)
-        select(tab: tab)
+        
+        if tabManager.model.tabs.indices.contains(index) {
+            let tab = tabManager.select(tabAt: index)
+            select(tab: tab)
+        } else {
+            assertionFailure("Invalid index selected")
```
samsymons, line 865: Do you think we could a debug pixel here to tell us how often this case is actually being triggered? (I'm not convinced that we do, but just asking anyway)

I was also wondering if we should default to selecting the last tab in the tab manager, in the case that the index somehow ends up being higher than the total count.

samsymons (approved):
Testing steps worked as expected :+1:

DuckDuckGo/MainViewController.swift
brindy: You got me thinking - I'm wondering if somehow the array is empty. It shouldn't be, but TabManager isn't really thread safe. 

    func clearAll() {
        tabs.removeAll() // 1
        tabs.append(Tab()) // 2
        currentIndex = 0 // 3
    }

1: at this point currentIndex will always be >= count and cause a problem if the array is accessed
2: if currentIndex >= 1 this could cause a problem
3:  finally ok

Without adding a bunch of locking, could change it to this to tighten it up a bit:

    func clearAll() {
        currentIndex = 0
        tabs = [Tab()]
    }

Worth changing?

samsymons: That sounds like a reasonable change. Tbh I would expect the tab manager to always be accessed from the main thread, but if that's not the case then yeah tidying this up sounds good.

brindy: Nah that's a good point. I don't want to change too much so I'll leave it.

Am also going to not make the other changes you suggest: 

* pixel: we'll see if this works as the crashes will go away, so doesn't seem useful (especially with triage overhead)
* set to last: if it *is* empty then that will likely cause some other problem. We can monitor feedback to see if folk complain about tab weirdness

PR #2388, opened 2024-01-25
Don't crash when AppRatingPromptEntity fetch errors

graeme:
Task/Issue URL: https://app.asana.com/0/0/1206441184793818/f

**Description**:

Removes a fatal error when we try to fetch the `AppRatingPromptEntity` and it errors. We might want to add a Pixel here, but there’s a silent recovery from the case that the fetch yields no results, so maybe we’re fine to just do the same here?

**Steps to test this PR**:
This is hard to reproduce and the fix doesn’t seem like it could go wrong. However perhaps someone could sanity check this for me in case there’s a reason we’d want to crash or at least make a bit more noise when this happens

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [x] iPhone X
* [ ] iPhone 14 Pro
* [ ] iPad

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16
* [x] iOS 17

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:


Core/PixelEvent.swift
```diff
@@ -1015,6 +1017,7 @@ extension Pixel.Event {
         // MARK: - Return user measurement
         case .debugReturnUserAddATB: return "m_debug_return_user_add_atb"
         case .debugReturnUserUpdateATB: return "m_debug_return_user_update_atb"
+        case .appRatingPromptFetchError: return "m_app_rating_prompt_fetch_error"
```
bwaresiak: Please use `m_debug_` :)

DuckDuckGo/AppRatingPrompt.swift
```diff
@@ -95,14 +101,20 @@ class AppRatingPromptCoreDataStorage: AppRatingPromptStorage {
...
-        
-        guard let results = try? context.fetch(fetchRequest) else {
-            fatalError("Error fetching AppRatingPromptEntity")
+
+        let results: [AppRatingPromptEntity]
+
+        do {
+            results = try context.fetch(fetchRequest)
+        } catch {
+            DailyPixel.fireDailyAndCount(pixel: .appRatingPromptFetchError, error: error)
```
bwaresiak: Does it send atb and appVersion? It should not.

bwaresiak (changes requested):

bwaresiak:
@graeme What about sending atb and appVersion parameters?

DuckDuckGo/AppRatingPrompt.swift
graeme: Oh I missed this comment. ATB is a good point but thought we’d perhaps want to know when it’s fixed by using appVersion. It’s included in the privacy triage.

bwaresiak (approved):
LGTM!

PR #2433, opened 2024-02-05
fix ddg cookies on fire button

brindy:
Task/Issue URL: https://app.asana.com/0/414709148257752/1206511646424647/f
Tech Design URL:
CC:

**Description**:

Keep DDG cookies when fire button is used.

**Steps to test this PR**:
1. Change 'safe search' to 'off' on the SERP (or via settings in SERP menu).  Change other settings too.
2. Use the fire button
3. Check the settings are retained. 

**DDG domain test**:
1. Login into use-devtesting2.duckduckgo.com (sorry)
5. Change some settings on that site
6. Ensure they are retained

### DIFF
--- a/Core/WebCacheManager.swift
+++ b/Core/WebCacheManager.swift
@@ -389,7 +389,7 @@ extension WKWebsiteDataStore: WebCacheManagerDataStore {
     public func preservedCookies(_ preservedLogins: PreserveLogins) async -> [HTTPCookie] {
         let allCookies = await self.httpCookieStore.allCookies()
         return allCookies.filter {
-            preservedLogins.isAllowed(cookieDomain: $0.domain)
+            "duckduckgo.com" == $0.domain || preservedLogins.isAllowed(cookieDomain: $0.domain)
         }
     }
 
--- a/Core/WebCacheManager.swift
+++ b/Core/WebCacheManager.swift
@@ -56,10 +56,6 @@ extension WebCacheManagerDataStore {
 }
 
 public class WebCacheManager {
-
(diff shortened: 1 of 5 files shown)


Core/WebCacheManager.swift
```diff
@@ -389,7 +389,7 @@ extension WKWebsiteDataStore: WebCacheManagerDataStore {
     public func preservedCookies(_ preservedLogins: PreserveLogins) async -> [HTTPCookie] {
         let allCookies = await self.httpCookieStore.allCookies()
         return allCookies.filter {
-            preservedLogins.isAllowed(cookieDomain: $0.domain)
+            "duckduckgo.com" == $0.domain || preservedLogins.isAllowed(cookieDomain: $0.domain)
```
dus7: There’s a similar check in `legacyDataClearing` function. Maybe you can reuse the `cookieDomainsToPreserve` or have an internal/fileprivate interface for the `WebCacheManager` to do the check against the allowed domains?

BTW, it would be awesome to add a unit test for this as well. The existing one (`testWhenClearedThenDDGCookiesAreRetained`) is only executing `legacyDataClearing`.

dus7 (approved):
Verified it works as expected. One suggestion to consider.

Core/WebCacheManager.swift
brindy: All this code needs to be cleaned up imo.  Just wanted to get a quick fix in for the issue.

dus7 (approved):
Looks good!

brindy:
Replaced by #2444

PR #2444, opened 2024-02-06
Keep DuckDuckGo Settings after Fire Button

brindy:
Task/Issue URL: https://app.asana.com/0/414709148257752/1206511646424647/f
Tech Design URL:
CC:

**Description**:
Keep DDG cookies when fire button is used.

Replace #2433

**Steps to test this PR**:
1. Change 'safe search' to 'off' on the SERP (or via settings in SERP menu).  Change other settings too.
2. Use the fire button
3. Check the settings are retained. 

**DDG domain test**:
1. Login into use-devtesting2.duckduckgo.com (sorry)
5. Change some settings on that site
6. Ensure they are retained

### DIFF
--- a/DuckDuckGoTests/FireButtonReferenceTests.swift
+++ b/DuckDuckGoTests/FireButtonReferenceTests.swift
@@ -44,7 +44,7 @@ final class FireButtonReferenceTests: XCTestCase {
         return url.host!
     }
 
-    func testClearData() async {
+    func testClearData() async throws {
         let preservedLogins = PreserveLogins.shared
         preservedLogins.clearAll()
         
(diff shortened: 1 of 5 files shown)


DuckDuckGoTests/FireButtonReferenceTests.swift
```diff
@@ -44,6 +44,54 @@ final class FireButtonReferenceTests: XCTestCase {
...
+        }
+        
+        let referenceTests = testData.fireButtonFireproofing.tests.filter {
+            $0.exceptPlatforms.contains("ios-browser") == false
+        }
+            
+        let cookieStorage = CookieStorage()
+        let idManager = DataStoreIdManager()
+        for test in referenceTests {
+            guard let cookie = cookie(for: test) else {
```
dus7: I find it better to use `try XCTUnwrap` with a throwing test function in such cases. It gives you automatic error message :).

dus7 (approved):
LGTM 👍

DuckDuckGoTests/FireButtonReferenceTests.swift
brindy: Ah very nice, will use that.

PR #2449, opened 2024-02-07
clean up clear data code

brindy:
Task/Issue URL: https://app.asana.com/0/414235014887631/1206545081824897/f
Tech Design URL:
CC: @loremattei 

**Description**:

* Cleans up the clear data code and makes it use the async/await API. 
* Removes some unnecessary and unused abstractions.
* Removes some obsolete diagnostics.
* Moves UI refresh to the end of the animation to improve consistency (should have been like this initially tbh).
* Fixes a bug where domains which remove all the cookies might get those cookies back later.
* setcookie.net is a convenient site for testing, but feel free to use any site that will drop a cookie (e.g. after login).
* It's important to test this on iOS 17 **and** an earlier version of iOS as there are two paths depending on OS version.
* Added a couple of maestro tests

**Steps to test this PR**:

***Maestro***
1. Run the tests under .maestro/data_clearing_tests

***Migration***
1. Check out an old version of the code (`main` will do)
2. Set some cookies on setcookie.net and fire proof the site 
3. Check the cookies are retained as expected
4. Checkout this branch and run the code
5. Check the cookies are still retained

***General Usage***
1. Go to setcookie.net and set a cookie (save it as a favourite too for convenience).  Refresh the page to confirm it has been set.
6. Use the fire button
7. Go back to setcookie.net - cookie should gone
8. Set another cookie and fire proof the site
9. Use the fire button
10. Cookie should be retained
11. Use the fire button a couple of times in a row
12. Cookie should still be present
13. Go to settings and remove the site from the list of fireproofed sites
14. Refresh the site and the cookie should have been removed
15. Do a search and on the SERP change a setting (e.g. safe search to off)
16. Use the fire button
17. Go back to the SERP - the setting should be retained
18. Login to some websites and fireproof - check that you stay logged in.
19. Remove fireproofing - check that you are logged out


bwaresiak (changes requested):
I've tested this a lot Today, and found two issues:

1. When using Auto Clear data, fireproofing doesn't seem to work when I kill and restart the app.
2. When data clear animation is set to "None" URL is not cleared from URL bar.

Other than this good job @brindy:

- Tested both iOS 15 and 17.
- Upgrading app works fine.
- With regular animation I could not reproduce the problem with URL bar glitch/data.
- Tested various scenarios, opening, navigating, clearing, reopening, restarting... all of these worked as expected.

DuckDuckGo/AppDelegate.swift
```diff
@@ -287,7 +287,11 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         }
 
         autoClear = AutoClear(worker: main)
-        autoClear?.applicationDidLaunch()
+//        Task { @MainActor in
+            // Internally this thing has always called 'async' functions but just never waited for them
+            //  so this is no different really.
+            // await autoClear?.applicationDidLaunch()
+//        }
```
bwaresiak: Remove? :)

DuckDuckGo/MainViewController.swift
```diff
@@ -154,6 +154,7 @@ class MainViewController: UIViewController {
     
     var postClear: (() -> Void)?
     var clearInProgress = false
+    var dataStoreReadiness: DataStoreWarmup? = DataStoreWarmup()
```
bwaresiak: readiness or a warmup? ;)

bwaresiak (changes requested):
Tested it this morning found two problems:
1. Auto clear doesn't seem to remove cookies on restart. :)
2. Blank Snapshot view has empty space instead of search bar.
IMG_0704

brindy:
@bwaresiak haven't addressed the missing bar on the empty view because I think that is not related to this change, right?

bwaresiak (approved):
Great job!

PR #2466, opened 2024-02-09
Send MM message in case of App Store Release failure

dus7:
Task/Issue URL: https://app.asana.com/0/0/1206553137165937/f

**Description**:
Sends MM message to actor in case of App Sore Release workflow failure.

**Steps to test this PR**:
1. Make App Store Release fail (not cancel)
2. Receive MM message

—
###### Internal references:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -108,7 +108,8 @@ jobs:
             --form "file=@${asana_dsyms_path};type=application/zip"
         fi
 
-    - name: Send Mattermost message
+      - name: Send Mattermost message
+      if: ${{ success() || failure() }} # Don't execute when cancelled
       env:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -108,7 +108,8 @@ jobs:
             --form "file=@${asana_dsyms_path};type=application/zip"
         fi
 
-    - name: Send Mattermost message
+      - name: Send Mattermost message
+      if: ${{ success() || failure() }} # Don't execute when cancelled
```
jaceklyp, line 112: perhaps we should just check: if: ! ${{ cancelled() }}

dus7: We could, but on the other hand if any other status shows up in the future, json won’t have mapping for it. Existing approach will allow only for those two explicit statuses to be processed. I’ll keep it how it is if you don’t mind.

jaceklyp (approved):

PR #2473, opened 2024-02-13
Move subscription Code to BSK

afterxleep:
Task/Issue URL:  https://app.asana.com/0/414235014887631/1206582087819536/f
Tech Design URL:
CC:

**Description**:
- Moves Subscriptions to BSK

**Steps to test this PR**:
1. Purchase/Restore a subscription


DuckDuckGo/AppDelegate.swift
```diff
@@ -33,6 +33,7 @@ import Configuration
 import Networking
 import DDGSync
 import SyncDataProviders
+import Subscription
```
miasma13, line 38: Question, shouldn't we guard those `import Subscription` in the `#if SUBSCRIPTION` checks?
This way we would ensure that the imports are not satisfied in the targets lacking the `SUBSCRIPTION` flag and we would ensure on compile time that no code using it would  be in those builds.

miasma13 (approved):

PR #2492, opened 2024-02-21
Alpha ad-hoc lane + workflow

ayoy:
Task/Issue URL:  https://app.asana.com/0/414709148257752/1206944335297057/f

**Description**:
- Adds an alpha build lane/workflow for ad-hoc builds


graeme (approved):
LGTM!

afterxleep:
@ayoy Finally had a moment to check this.  I think we can merge as is and parametrize later.  There are few moving/changing parts. WDYT? 

What would be the expected scenario for parametrization? Having a single ad-hoc build action and picking PROD/ALPHA alongside the branch?   🤔

ayoy:
@afterxleep thanks! Regarding parametrization I only meant reworking fastlane code to avoid duplication, but I agree that it shouldn't block us from merging this. GHA workflow can be added at a later time 👍🏻

Please merge it at your convenience then :) thanks for moving this forward 🙏🏻

cc @jaceklyp

PR #2497, opened 2024-02-23
Autofill support for deleting all passwords 

amddg44:
Task/Issue URL: https://app.asana.com/0/1201462886803403/1206303310228833/f
Tech Design URL:
CC:

**Description**:
Give users the ability to delete all passwords they have saved

**Steps to test this PR**:

**Delete all (multiple passwords saved)**

1. Go to the Passwords screen & save 2 or more passwords
2. Tap `Edit` and confirm the toolbar appears as per the figma design
3. Confirm the top section which contains the row `Save and autofill passwords` is removed
4. Confirm the toobar label has the correct password count
5. Tap `Delete All`
6. Confirm the alert dialog text includes the correct password count
7. If sync is enabled, confirm it is mentioned in the alert message or if sync is not enabled, confirm it is not mentioned in the alert message
8. Follow the prompts and confirm you are prompted to authenticate in order to delete all
9. Cancel / fail the authentication, making sure the delele-all flow is cancelled
10. Repeat and this time authenticate successfully
11. Confirm all passwords disappear from the UI
12. Confirm the toast prompt appears with the correct password count & an `Undo` button
13. Tap `Undo` and confirm the passwords re-appear
14. Back out of the screen and come back, confirming the passwords still appear correctly
15. Repeat but this time don't tap `Undo`.
16. Back out of the screen and come back, confirming there are no passwords saved

**Delete one password**

1. Tap `Edit` and delete a single password
2. Confirm the toolbar password count updates correctly

**Delete all (one password saved)**

The copy is slightly different in the case of the user having just a single password saved:
1. Ensure only a single password is saved
2. Tap `Delete All` and confirm the alert title mentions "this password" instead of the password count
3. Follow the complete `Delete All` flow and confirm the singular is used in all the copy in reference to passwords / accounts

**Password details screen**

1. Tap a password to see its details
2. Tap the `Delete` button and confirm the title & message are the same as the `Delete all (one password saved)` alert 
3. If sync is enabled, confirm it is mentioned in the alert message or if sync is not enabled, confirm it is not mentioned in the alert message

**Sync**

1. With Sync enabled, confirm the passwords are deleted on the other devices
2. Depending on whether or not you were testing with Sync enabled / disabled, turn it off / on and confirm the `Delete all` alert copy doesn't / does mention sync in the alert message

**Toolbar behaviour**

1. Confirm the toolbar only appears when in `Edit` mode and not in any of these cases:
- when table is empty
- when searching
- when accessing password details
- when screen is locked for authentication
2. Access the Passwords screen via Settings (not from the menu overflow)
3. Tap `Edit` and back out of the Passwords screen
4. Confirm the toolbar is correctly removed

**Authentication**

1. Repeat the `Delete all` flow but this time minimise the app at each stage
2. Confirm authentication is always required to view the list of passwords when returning to the app

**Save a never prompt site**

1. Visit `https://fill.dev/form/login-simple` and enter some credentials
2. Tap `Never Ask for this Site`
3. Go to the Passwords screen and confirm the row `Reset Excluded Sites` is now present in the top section
4. Delete all passwords & confirm this row disappears after the toast disappears 

**Additional smoke tests**

1. Lightly test search & confirm everything still works correctly

**Copy Testing**:

**Orientation Testing**:

* [x] Portrait
* [x] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPhone 14 Pro
* [ ] iPad

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16

**Theme Testing**:

* [x] Light theme
* [x] Dark theme

---
###### Internal references:


DuckDuckGo/AuthConfirmationPromptViewController.swift
```diff
@@ -0,0 +1,106 @@
...
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import UIKit
+import SwiftUI
+
+class AuthConfirmationPromptViewController: UIViewController {
```
aataraxiaa: Nitpick: Maybe could mark this `final`? 

```suggestion
final class AuthConfirmationPromptViewController: UIViewController {
```

DuckDuckGo/AuthConfirmationPromptViewController.swift
```diff
@@ -0,0 +1,106 @@
...
+    
+    typealias AuthConfirmationCompletion = (_ authenticated: Bool) -> Void
+    
+    private let didBeginAuthenticating: () -> Void
+    private let authConfirmationCompletion: AuthConfirmationCompletion
+    
+    private var viewModel: AuthConfirmationPromptViewModel?
+    
+    internal init(didBeginAuthenticating: @escaping () -> Void,
```
aataraxiaa: Curious: Why do we need to mark this `init` internal?

DuckDuckGo/AuthConfirmationPromptViewModel.swift
```diff
@@ -0,0 +1,60 @@
...
+import Foundation
+
+protocol AuthConfirmationPromptViewModelDelegate: AnyObject {
+    func authConfirmationPromptViewModelDidBeginAuthenticating(_ viewModel: AuthConfirmationPromptViewModel)
+    func authConfirmationPromptViewModelDidAuthenticate(_ viewModel: AuthConfirmationPromptViewModel, success: Bool)
+    func authConfirmationPromptViewModelDidCancel(_ viewModel: AuthConfirmationPromptViewModel)
+    func authConfirmationPromptViewModelDidResizeContent(_ viewModel: AuthConfirmationPromptViewModel, contentHeight: CGFloat)
+}
+
+class AuthConfirmationPromptViewModel: ObservableObject {
```
aataraxiaa: Nitpick: Could maybe mark this `final`?

DuckDuckGo/AutofillLoginDetailsViewController.swift
```diff
@@ -38,6 +39,7 @@ class AutofillLoginDetailsViewController: UIViewController {
     private let viewModel: AutofillLoginDetailsViewModel
     private var cancellables: Set<AnyCancellable> = []
     private var authenticator = AutofillLoginListAuthenticator(reason: UserText.autofillLoginListAuthenticationReason)
+    private let syncService: DDGSyncing
```
aataraxiaa: Maybe we can remove this? I think it is only passed into the ViewModel and then not referenced again.

aataraxiaa (approved):
Verified scenarios and left some very minor comments. LGTM @amddg44. 🥳

DuckDuckGo/AutofillLoginDetailsViewController.swift
amddg44: Good point! Done 🙂

DuckDuckGo/AuthConfirmationPromptViewController.swift
amddg44: Yup, no need for this to be marked internal; updated

PR #2503, opened 2024-02-26
Fixes the firing of `mh` pixel

brindy:
Task/Issue URL: https://app.asana.com/0/414709148257752/1206674614649925/f
Tech Design URL:
CC:

**Description**:
Fixes the firing of `mh` pixel

**Steps to test this PR**:
1. Launch the app, ensure `mh` pixel is fired when the the new tab is shown
2. Create tabs from the tab switcher.  `mh` should only be fired when the the user ends up on the Home Screen
3. Check RMF message still displays as expected

### DIFF
--- a/DuckDuckGo/HomeViewController.swift
+++ b/DuckDuckGo/HomeViewController.swift
@@ -264,12 +264,18 @@ class HomeViewController: UIViewController {
     
     override func viewDidAppear(_ animated: Bool) {
         super.viewDidAppear(animated)
-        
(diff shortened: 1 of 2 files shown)


amddg44 (approved):
LGTM! I've tested on iPhone & iPad
- Creating new tabs via the tab switcher, long press on tab bar button and via overflow menu
- Switching to new tabs
- No impact to RMF
- No impact to onboarding

PR #2508, opened 2024-02-28
14. Implemented Error handlers for all subscription scenarios

afterxleep:
Task/Issue URL: https://app.asana.com/0/414235014887631/1206707680638882/f

**Description**:
This adds different error handling 'handles' to attach pixels and events, and streamlines the overall error handling flow in the UI

**Steps to test this PR**:
Confirm the 'happy paths' work as expected..

For example:
1. Get a subscrioption
2. Add an email
3. Remove from device
4. Restore Via App Store connect
5. Remove from device
6. Restore via email

<!--
---
###### Internal references:


DuckDuckGo/Subscription/UserScripts/SubscriptionPagesUseSubscriptionFeature.swift
```diff
@@ -221,10 +235,29 @@ final class SubscriptionPagesUseSubscriptionFeature: Subfeature, ObservableObjec
...
+            case .success:
+
+                // Store the account data and mark as active
+                accountManager.storeAccount(token: accessToken,
+                                            email: accountDetails.email,
+                                            externalID: accountDetails.externalID)
+                emailActivationComplete = true
+                
+            case .failure(let error):
+                os_log(.info, log: .subscription, "Failed to restore subscription from Email")
```
federicocappelli: should this be an .error log?

DuckDuckGo/Subscription/UserScripts/SubscriptionPagesUseSubscriptionFeature.swift
```diff
@@ -221,10 +235,29 @@ final class SubscriptionPagesUseSubscriptionFeature: Subfeature, ObservableObjec
...
+                transactionError = .failedToRestoreFromEmail
+            }
         } else {
-            os_log(.info, log: .subscription, "Failed to restore subscription from Email")
+            os_log(.info, log: .subscription, "General error. Could not get account Details")
```
federicocappelli: same here, .error log?

DuckDuckGo/Subscription/UserScripts/SubscriptionPagesUseSubscriptionFeature.swift
```diff
@@ -237,6 +270,7 @@ final class SubscriptionPagesUseSubscriptionFeature: Subfeature, ObservableObjec
     func featureSelected(params: Any, original: WKScriptMessage) async throws -> Encodable? {
```
federicocappelli: (general comment, not only for this function)

Overall the pattern of having functions that express a failure returning a nil value is very Objective-C, the return value should be optional only if an optional value makes sense for a positive outcome in the context and not as a way to signal an error, the errors should be propagated by throwing exceptions. 
Additionally, modifying an external variable for storing errors makes the function's interface obscure.

Suggestions:
- make the return value not optional
- propagate errors throwing them and ctaching them in the function callers

DuckDuckGo/Subscription/UserScripts/SubscriptionPagesUseSubscriptionFeature.swift
```diff
@@ -67,10 +67,22 @@ final class SubscriptionPagesUseSubscriptionFeature: Subfeature, ObservableObjec
...
+             failedToRestorePastPurchase,
+             subscriptionExpired,
+             hasActiveSubscription,
+             generalError
+    }
         
     @Published var transactionStatus: SubscriptionTransactionStatus = .idle
-    @Published var hasActiveSubscription = false
-    @Published var purchaseError: AppStorePurchaseFlow.Error?
+    @Published var transactionError: UseSubscriptionError?
```
federicocappelli: (Suggestion, no need to act on it right now)
Storing a var that is accessible from everywhere and everyone is a non-pattern, it's difficult to debug (who set it, who is reading it?) and defeats the overall Swift error handling approach (throw and catch) that allows the error path to be linear.
If an error must be published I would suggest setting the variable only from a single function that is responsible for catching errors from every other function.

DuckDuckGo/Subscription/UserScripts/SubscriptionPagesUseSubscriptionFeature.swift
```diff
@@ -221,10 +235,29 @@ final class SubscriptionPagesUseSubscriptionFeature: Subfeature, ObservableObjec
...
            case let .success(accountDetails) = await accountManager.fetchAccountDetails(with: accessToken) {
-            accountManager.storeAccount(token: accessToken, email: accountDetails.email, externalID: accountDetails.externalID)
-            emailActivationComplete = true
+            switch await SubscriptionService.getSubscriptionDetails(token: accessToken) {
+            
+            // If the account is not active, display an error and logout
+            case .success(let response) where !response.isSubscriptionActive:
+                transactionError = .failedToRestoreFromEmailSubscriptionInactive
+                accountManager.signOut()
+                return nil
```
federicocappelli, line 265: Do any of these functions marked as "throws" effectively throw any error ever?

DuckDuckGo/Subscription/UserScripts/SubscriptionPagesUseSubscriptionFeature.swift
```diff
@@ -257,28 +291,37 @@ final class SubscriptionPagesUseSubscriptionFeature: Subfeature, ObservableObjec
 
     func pushAction(method: SubscribeActionName, webView: WKWebView, params: Encodable) {
         let broker = UserScriptMessageBroker(context: SubscriptionPagesUserScript.context, requiresRunInPageContentWorld: true )
-        
         broker.push(method: method.rawValue, params: params, for: self, into: webView)
     }
     
-    func restoreAccountFromAppStorePurchase() async -> Bool {
+    func restoreAccountFromAppStorePurchase() async throws {
```
federicocappelli, line 319: why "throws"?

DuckDuckGo/Subscription/ViewModel/SubscriptionFlowViewModel.swift
```diff
@@ -131,15 +136,49 @@ final class SubscriptionFlowViewModel: ObservableObject {
...
+            transactionError = .generalError
+        }
+    }
+    
     private func setupWebViewObservers() async {
         webViewModel.$scrollPosition
             .receive(on: DispatchQueue.main)
             .sink { [weak self] value in
+                guard let strongSelf = self else { return }
                 DispatchQueue.main.async {
```
federicocappelli, line 183: You already use `.receive(on: DispatchQueue.main)` , is this needed?

federicocappelli (commented):
Few general comments about the error handling approach

DuckDuckGo/Subscription/UserScripts/SubscriptionPagesUseSubscriptionFeature.swift
afterxleep: As for propagating errors via exceptions, that's the ideal scenario, sure; but in this context of the UserScripts architecture, class methods are invoked by external sources (webviews > userscripts), and view models do not directly control when these methods are executed. (Except for transaction restore)

Propagating errors and other state changes through published properties is more like a necessity rather than a choice.

DuckDuckGo/Subscription/UserScripts/SubscriptionPagesUseSubscriptionFeature.swift
afterxleep: We need it to be openly readable because of the reason explained above, but I could probably make it private(set), and use a single private function in the class to do that.

DuckDuckGo/Subscription/UserScripts/SubscriptionPagesUseSubscriptionFeature.swift
afterxleep: nop.  good call.  I'll just kill that.

DuckDuckGo/Subscription/UserScripts/SubscriptionPagesUseSubscriptionFeature.swift
afterxleep: This method is effectively called from viewModels, so was going to handle the error there instead of the publisher but missed it.  Good catch.

DuckDuckGo/Subscription/ViewModel/SubscriptionFlowViewModel.swift
afterxleep: good catch.

DuckDuckGo/Subscription/UserScripts/SubscriptionPagesUseSubscriptionFeature.swift
afterxleep: We're returning an optional, because this method is called from the `UserScriptMessageBroker`, which expects an optional message.     While I guess this could be updated, the scope of the subscription project did not include the development of these frameworks.-

DuckDuckGo.xcodeproj/project.pbxproj
```diff
@@ -9920,8 +9922,8 @@
 			isa = XCRemoteSwiftPackageReference;
 			repositoryURL = "https://github.com/DuckDuckGo/BrowserServicesKit";
 			requirement = {
-				kind = exactVersion;
-				version = 112.0.0;
+				branch = "daniel/subcription.user-cancel";
```
federicocappelli: I assume this can't be merged in the Main branch, right?

federicocappelli (changes requested):

DuckDuckGo.xcodeproj/project.pbxproj
afterxleep: No. This will be updated once the required BSK PR is merged and BSK is released:https://github.com/duckduckgo/BrowserServicesKit/pull/684

federicocappelli: I may have missed something here, this PR is open and is against Main, are you going to merge it now (soon) or wait for BSK PR? if you need to wait for BSK then you should tag the PR and follow the process in https://app.asana.com/0/1202500774821704/1200837094583426/f

afterxleep: @federicocappelli This is what we normally do for BSK changes

1.  We open 3 PRs.  BSK, macOS and iOS.  
2. Both macOS point to the BSK branch for testing
3. Once everything approved, we release BSK, and update the macOS and iOS PRs to point to that release
4. We merge both iOS and MacOS PR's together.

This PR is referenced in the BSK PR I mentioned and since these changes wont affect macOS immediately, Michal will open a macOS PR once we release BSK.

federicocappelli: Ok, I understand where the problem is here, I'm not used to the fact that a PR can be changed after is approved, and you want me to approve this and then change the BSK reference, I'm ok with that, I'm just not sure why the BSK reference is not already in when the BSK PR has been approved.
LGTM anyway

federicocappelli (approved):

DuckDuckGo.xcodeproj/project.pbxproj
afterxleep: @federicocappelli  As mentioned in MM, we often approve before releasing BSK, so the author can quickly merge all three PR's async. (macOS and iOS PRs are often opened by different people)

I've just pointed to 113.1.0 release in case you want to have a look before merging

federicocappelli: I'm good with that, free to merge

PR #2509, opened 2024-02-28
rename cookie and ensure timeout does not fire prematurely

brindy:
Task/Issue URL: https://app.asana.com/0/414235014887631/1206707849446765/f
Tech Design URL:
CC:

**Description**:

rename cookie and ensure timeout does not fire prematurely

**Steps to test this PR**:
1. Use the fire button - the cookie timeout pixel should not fire
1. Add `try? await Task.sleep(interval: 6.0)` to the `legacyDataClearing()` function before the timeout task gets cancelled
3. Use the fire button - the cookie timeout pixel will fire

### DIFF
--- a/Core/PixelEvent.swift
+++ b/Core/PixelEvent.swift
@@ -899,7 +899,7 @@ extension Pixel.Event {
             
         case .blankOverlayNotDismissed: return "m_d_ovs"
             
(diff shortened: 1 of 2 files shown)


quanganhdo (approved):
LGTM & tested on my phone 👍

PR #2519, opened 2024-02-29
Add simple behavior monitoring to better address potential breakages

jaceklyp:
Task/Issue URL: https://app.asana.com/0/0/1206706226010039/f

**Description**:
We want to identify behavioral signals that indicate a user is experiencing breakage.

**Steps to test this PR**:
1. Run UserBehaviorMonitorTests
1. Go through scenarios listed here: https://app.asana.com/0/0/1206706226010039/f and see if pixels are being sent.

### DIFF
--- a/Core/PixelEvent.swift
+++ b/Core/PixelEvent.swift
@@ -526,6 +526,14 @@ extension Pixel {
         case compilationFailed
 
         case appRatingPromptFetchError
+
+        case userBehaviorReloadTwice
+        case userBehaviorReloadAndRestart
+        case userBehaviorReloadAndFireButton
+        case userBehaviorReloadAndOpenSettings
+        case userBehaviorReloadAndTogglePrivacyControls
+        case userBehaviorFireButtonAndRestart
+        case userBehaviorFireButtonAndTogglePrivacyControls
     }
     
 }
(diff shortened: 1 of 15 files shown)


DuckDuckGo/UserBehaviorMonitor.swift
```diff
@@ -0,0 +1,89 @@
...
+
+    @UserDefaultsWrapper(key: .didRefreshTimestamp, defaultValue: .distantPast)
+    private var didRefreshTimestamp: Date?
+    @UserDefaultsWrapper(key: .didBurnTimestamp, defaultValue: .distantPast)
+    private var didBurnTimestamp: Date?
+
+    func handleAction(_ action: Action, date: Date = Date()) {
+        switch action {
+        case .refresh:
+            fireEventIfActionOccurredRecently(since: didRefreshTimestamp, eventToFire: .reloadTwice, within: 10.0)
```
samsymons, line 49: Do you think this also needs to take into account whether the reload has occurred on the same tab? This pixel can fire if you reload one tab, swap to another one, and then also reload that, which is likely not indicative of the sort of problem we're looking for.

samsymons: Marking resolved since this was discussed internally already.

Core/PixelEvent.swift
```diff
@@ -1024,6 +1032,15 @@ extension Pixel.Event {
         case .debugReturnUserUpdateATB: return "m_debug_return_user_update_atb"
             
         case .appRatingPromptFetchError: return "m_d_app_rating_prompt_fetch_error"
+
+        // MARK: - User behavior
+        case .userBehaviorReloadTwice: return "m_reload-twice"
```
samsymons, line 1037: One other thing to check with this pixel is that if the user refreshes three times in a row, the pixel will send twice. Do you think we need some buffer to make it so that this pixel really means "user tried to reload repeatedly to fix a problem", or is it okay?

PR #2520, opened 2024-02-29
15. Subscriptions: Properly sign out users and cache state

afterxleep:
Task/Issue URL: https://app.asana.com/0/0/1206707680638927/f

**Description**:
- If the subscription has become inactive, the user may still see the options in settings.  This forces a signout in that case and caches the subscription state data properly.

**Steps to test this PR**:
1.  Purchase a sub
2. Close setings
3.  Wait for it to expire
4.  Open the app
5. Subscription options should not be shown.  (Only purchase options)

---
###### Internal references:

### DIFF
--- a/DuckDuckGo/SettingsViewModel.swift
+++ b/DuckDuckGo/SettingsViewModel.swift
@@ -340,15 +340,14 @@ extension SettingsViewModel {
         // Fetch available subscriptions from the backend (or sign out)
         switch await SubscriptionService.getSubscriptionDetails(token: token) {
         case .success(let response) where !response.isSubscriptionActive:
-            AccountManager().signOut()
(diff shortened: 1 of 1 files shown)


miasma13:
For the described scenario it worked as expected. Though for the first time I left the app when subscription was active with the settings open. And returning to the app the settings didn't refresh. 
Not sure how hard it would be but WDYT about refreshing the subscription part of settings when returning to the app.

miasma13 (approved):

afterxleep:
> For the described scenario it worked as expected. Though for the first time I left the app when subscription was active with the settings open. And returning to the app the settings didn't refresh. Not sure how hard it would be but WDYT about refreshing the subscription part of settings when returning to the app.

Yeap, adding notification subscribers in the next PR

PR #2539, opened 2024-03-05
Use History on iOS

brindy:
Task/Issue URL: https://app.asana.com/0/414235014887631/1206524433066955/f
Tech Design URL:
CC:

**Description**:
Adds history collection to iOS

**Steps to test this PR**:
1. See https://github.com/duckduckgo/BrowserServicesKit/pull/693
2. In the manager's  computed var for the coordinator where the store is loaded, change code to set an error message and ensure that the preemptive crash alert is shown.


Core/HistoryCapture.swift
```diff
@@ -0,0 +1,74 @@
...
+}
+
+extension URL {
+
+    var urlOrDuckDuckGoCleanQuery: URL {
+        guard isDuckDuckGoSearch,
+                let searchQuery,
+                let url = URL.makeSearchURL(query: searchQuery)?.removingInternalSearchParameters() else { return self }
+        return url
+    }
```
brindy, line 67: This cleans up a search URL for consistency and so that the other parameters aren't stored in history (we'll re-add them anyway)

Core/HistoryManager.swift
```diff
@@ -17,21 +17,182 @@
...
 
-class HistoryManager {
+public protocol HistoryManaging {
 
-    let privacyConfig: PrivacyConfiguration
+    var historyCoordinator: HistoryCoordinating { get }
+
+}
+
+public class HistoryManager: HistoryManaging {
```
brindy, line 34: Manages the history coordinator so we can swap it out if the privacy config changes in real time.

Core/HistoryManager.swift
```diff
@@ -17,21 +17,182 @@
...
+        await withCheckedContinuation { continuation in
+            historyCoordinator.burnAll {
+                continuation.resume()
+            }
+        }
+    }
+
+}
+
+class NullHistoryCoordinator: HistoryCoordinating {
```
brindy, line 96: Basically a no-op history coordinator for use when history is disabled and in the tab previews.

Core/HistoryManager.swift
```diff
@@ -17,21 +17,182 @@
...
+        guard isHistoryFeatureEnabled() else {
+            currentHistoryCoordinator = nil
+            return NullHistoryCoordinator()
+        }
+
+        if let currentHistoryCoordinator {
+            return currentHistoryCoordinator
+        }
+
+        database.loadStore()
```
bwaresiak: Shouldn't we handle load error (pixel etc)? Also check how `shouldPresentInsufficientDiskSpaceAlertAndCrash` is used. :)

DuckDuckGo/HistoryDebugViewController.swift
```diff
@@ -0,0 +1,96 @@
...
+        let items = try? context.fetch(fetchRequest)
+        items?.forEach { obj in
+            context.delete(obj)
+        }
+        do {
+            try context.save()
+        } catch {
+            assertionFailure("Failed to save after delete all")
+        }
+        fetch()
```
samsymons, line 86: I think I'm slightly surprised to not see `HistoryStore` used to do the clearing here, since it should conceal the underlying implementation details that this function is having to care about - but given this is a debug view controller, I'm not bothered by it.

DuckDuckGo/TabViewController.swift
```diff
@@ -174,6 +175,7 @@ class TabViewController: UIViewController {
         didSet {
             updateTabModel()
             delegate?.tabLoadingStateDidChange(tab: self)
+            historyCapture.titleDidChange(title, forURL: url)
```
samsymons, line 179: For whatever reason, when I first started testing this by visiting `nytimes.com` I kept getting visits with no title. After I added some logging, I can't reproduce the issue even once. 🤷 Mentioning it here in case it sounds familiar, but if not then let's just ignore it.

DuckDuckGo/HistoryDebugViewController.swift
brindy: Yeah this thing is accessing the context directly.  That was deliberate tbh, because if there's a bug in history store you can end up with multiple entries for a history and this also lets us confirm that all history entries have actually been removed from the database.

DuckDuckGo/TabViewController.swift
brindy: I'll see if I can replicate on a clean build.

brindy: Fixed! Not really happy about it, but it's a side effect of how this code uses Futures and the fact that I want to support being able to change history state from privacy config on the fly rather than having to wait for a reload (unlike macOS which doesn't have privacy config controls for history).  In the future I think it'd be a good idea to refactor this all to use async/await instead.

samsymons (approved):

PR #2546, opened 2024-03-06
Validate correct environment

bwaresiak:
Task/Issue URL: https://app.asana.com/0/0/1206645876801807/f
Tech Design URL:
CC:

**Description**:

Fire Unique pixel in case we are on dev environment but we most likely should not be.

**Steps to test this PR**:

Smoke tests.

---
###### Internal references:


Core/PixelEvent.swift
```diff
@@ -1004,6 +1005,8 @@ extension Pixel.Event {
         case .syncDeleteAccountError: return "m_d_sync_delete_account_error"
         case .syncLoginExistingAccountError: return "m_d_sync_login_existing_account_error"
 
+        case .syncWrongEnvironment: return "m_d_sync_wrong_environment"
```
ayoy:
```
Fatal error: Unique pixel: must end with _u
```
:)

ayoy (approved):
Works as expected, but catches an assertion because of the reason I posted in the comment. Other than that it's good to go :) pending privacy triage 😇

Core/PixelEvent.swift
bwaresiak: This is confusing - DailyPixel appends _d and _c automatically but Unique does not, Going to update pixel name now.

PR #2552, opened 2024-03-07
Use History in Suggestions on iOS

brindy:
Task/Issue URL: https://app.asana.com/0/0/1206524433066958/f
Tech Design URL:
CC:

**Description**:
Surfaces history in the autocomplete.

**Steps to test this PR**:
1. Perform a search
2. Start typing part of the search, it should appear as a history item (with the clock icon) at the bottom of suggestions
3. Browse to some sites
4. Start typing part of some of the site titles, they should appear as history items at the bottom of the suggestions
5. Visit a couple of pages several times (ie more then 3)
6. Start typing part of those page titles, they should appear as history items at the top
7. Add bookmarks and favourites and try again.  Appropriate suggestions should be made.
8. Use the fire button, history should no longer appear in suggestions.


DuckDuckGo/AutocompleteViewController.swift
```diff
@@ -125,24 +143,29 @@ class AutocompleteViewController: UIViewController {
...
+        case .phrase(phrase: let phrase):
+            Pixel.fire(pixel: .autocompleteClickPhrase)
+        case .website(url: let url):
+            Pixel.fire(pixel: .autocompleteClickWebsite)
+        case .bookmark(title: let title, url: let url, isFavorite: let isFavorite, allowedInTopHits: let allowedInTopHits):
+            Pixel.fire(pixel: isFavorite ? .autocompleteClickFavorite : .autocompleteClickBookmark)
+        case .historyEntry(title: let title, url: let url, allowedInTopHits: let allowedInTopHits):
+            Pixel.fire(pixel: .autocompleteClickHistory)
+        case .unknown(value: let value):
+            assertionFailure("Unknown suggestion")
```
dus7: You don’t use all associated values here, so this can be cleaned up a tiny bit:
```suggestion
        case .phrase:
            Pixel.fire(pixel: .autocompleteClickPhrase)
        case .website:
            Pixel.fire(pixel: .autocompleteClickWebsite)
        case .bookmark(_, _, isFavorite: let isFavorite, _):
            Pixel.fire(pixel: isFavorite ? .autocompleteClickFavorite : .autocompleteClickBookmark)
        case .historyEntry:
            Pixel.fire(pixel: .autocompleteClickHistory)
        case .unknown:
            assertionFailure("Unknown suggestion")
        }
```

DuckDuckGo/AutocompleteViewController.swift
```diff
@@ -152,47 +175,30 @@ class AutocompleteViewController: UIViewController {
...
+                  url.isValid else {
+                return nil
             }
+
+            return url
+        })
+        pendingRequest = true
+
+        loader?.getSuggestions(query: query) { [weak self] result, error in
+            guard error == nil else { return }
```
dus7, line 202: `pendingRequest` remains set to true in case of error. I see for now it’s read only for key commands but I’d still reset the value here to make the state consistent.

Also I’d consider clearing suggestions in such case. Now when you have suggestions list populated but next load ends with an error the list is still visible. Consider the following scenario:
1. Type “foo” in search - list shows up
2. Replace search with “bar” - this code fails and suggestions for “foo” are still visible.

DuckDuckGo/CachedBookmarkSuggestions.swift
```diff
@@ -0,0 +1,69 @@
...
+//  limitations under the License.
+//
+
+import Foundation
+import CoreData
+import Bookmarks
+import Suggestions
+import Persistence
+
+class CachedBookmarks {
```
dus7: nitpick - it could be `final`.

DuckDuckGo/CachedBookmarkSuggestions.swift
```diff
@@ -0,0 +1,69 @@
...
+
+    let context: NSManagedObjectContext
+    lazy var all: [BookmarkSuggestion] = {
+        let fetchRequest = NSFetchRequest<NSFetchRequestResult>(entityName: "BookmarkEntity")
+        fetchRequest.predicate = NSPredicate(
+            format: "%K = false AND %K == NO",
```
dus7: It works, but the mixed usage of `false` and `NO` looks weird to me :).

dus7 (commented):
Looks good! Just a few things to consider.

DuckDuckGo/CachedBookmarkSuggestions.swift
brindy: It should be yeah - we used to lint this 🤔

DuckDuckGo/AutocompleteViewController.swift
brindy: Agree on the pending request.  I think the chance of error is high (macOS seems to assume it will happen which is where I copied this from), but I'll test it and see what happens.

DuckDuckGo/CachedBookmarkSuggestions.swift
brindy: Copy pasted from existing code, but I agree :)

dus7 (approved):

PR #2557, opened 2024-03-07
Update secure vault error pixels to fire daily

amddg44:
Task/Issue URL: https://app.asana.com/0/414709148257752/1206271742997031/f
Tech Design URL:
CC:

**Description**:
Updates the noisiest secure vault pixels to only fire once per day to get an accurate sense of the number of users impacted

Also updates DailyPixel key to also take error values into account so that the same pixel firing with different errors will fire once for each different error (not just once per pixel)

**Steps to test this PR**:
1. Launch the app, filter for `pixel` in the console and confirm there are no secure vault error related pixels
2. Update BSK to use this branch: `anya/secure-vault-error-pixels-hack` which will throw 2 alternating `secureVaultInitFailed` errors (`noL1Key` and `noL2Key`)
3. Add a logging breakpoint that prints: `Pixel NOT fired for @key@`
 
image

4. Re-launch the app and navigate around a couple of websites. Only 2 `secureVaultInitFailedError` pixels should fire (with errors `noL1Key` and `noL2Key` respectively). No more `secureVaultInitFailedError` pixels should fire and the console should have a long log of all the pixel events that were prevented from being fired 

image
 

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPhone 14 Pro
* [ ] iPad

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:


Core/DailyPixel.swift
```diff
@@ -46,20 +46,28 @@ public final class DailyPixel {
     private static let storage: UserDefaults = UserDefaults(suiteName: Constant.dailyPixelStorageIdentifier)!
 
     /// Sends a given Pixel once per day.
-    /// This means a pixel will get sent twice the first time it is called per-day.
```
amddg44, line 49: Removed this line as it doesn't make sense (I think it was a bad copy / paste from `fireDailyAndCount` comment)

DuckDuckGo/SecureVaultErrorReporter.swift
```diff
@@ -27,34 +27,41 @@ final class SecureVaultErrorReporter: SecureVaultErrorReporting {
     static let shared = SecureVaultErrorReporter()
     private init() {}
 
+    @MainActor
```
amddg44, line 30: Moved the application state check here and explicitly set it to run on `main` since I noticed while testing that it was sometimes being called from other threads 
image

samsymons: Nice, good find :+1:

Core/DailyPixel.swift
```diff
@@ -97,19 +107,27 @@ public final class DailyPixel {
...
             return Date().isSameDay(lastFireDate)
         }
         return false
     }
-    
+
+    private static func createSortedStringOfValues(from dict: [String: String], maxLength: Int = 50) -> String {
+        let sortedKeys = dict.keys.sorted()
+
+        let uniqueString = sortedKeys.map { "\(dict[$0]!)" }.joined(separator: ";")
```
samsymons: We could probably use `compactMap` here to filter out nil values and avoid the force unwrap, but tbh I can't think of a scenario where this would fail.

samsymons (approved):

PR #2559, opened 2024-03-07
Prevent redeeming invite codes after the VPN test ends

samsymons:
Task/Issue URL: https://app.asana.com/0/414235014887631/1206789264380520/f
Tech Design URL:
CC:

**Description**:

This is the iOS equivalent of https://github.com/duckduckgo/macos-browser/pull/2347.

**Steps to test this PR**:

This is harder to test since the staging waitlist endpoint won't work here, but you can do the following:

1. Change `let inviteCode = waitlistStorage.getWaitlistInviteCode()` to be hardcoded to a real string (doesn't have to be a real invite code, just needs to not be nil)
2. Change `waitlistBetaActive` to be hardcoded to false
3. Run the app and check that it doesn't try (and fail) to redeem the code, it should stop due to `waitlistBetaActive` not being true

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16

---
###### Internal references:

### DIFF
--- a/DuckDuckGo/AppDelegate+Waitlists.swift
+++ b/DuckDuckGo/AppDelegate+Waitlists.swift
@@ -22,6 +22,7 @@ import Core
 import BackgroundTasks
 import NetworkProtection
 import Waitlist
+import BrowserServicesKit
 
 extension AppDelegate {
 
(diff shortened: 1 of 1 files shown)


DuckDuckGo/AppDelegate+Waitlists.swift
```diff
@@ -51,12 +52,17 @@ extension AppDelegate {
 
         VPNWaitlist.shared.fetchInviteCodeIfAvailable { [weak self] error in
             guard error == nil else {
-#if !DEBUG
```
samsymons, line 54: Removed this since there's no reason this code shouldn't run in debug code any more. It was better this way early on, but now we can enable this.

jotaemepereira (approved):

PR #2603, opened 2024-03-15
Updated settings

tomasstrba:
Task/Issue URL: https://app.asana.com/0/0/1206668965195606/f

**Description**:
New Settings layout with Privacy Protections, Main Settings, Next Steps and others. It also includes the implementation of pixel experiment for monitoring the usage of old and new Settings to make sure we don't harm the experience.

ℹ️  The experiment will be activated in a follow-up PR after running experiments on iOS are over.

**Steps to test this PR**:

**Test the old Settings**
1. Manually set the cohort in PixelExperiment.swift file

```
var cohort: PixelExperiment? {
    return .control

    guard isInstalled else { return nil }
...
```
2. Make sure every setting work as expected and is persisted properly

**Test the new Settings**
1. Manually set the cohort in PixelExperiment.swift file

```
var cohort: PixelExperiment? {
    return .newSettings

    guard isInstalled else { return nil }
...
```
2. Make sure every setting work as expected and is persisted properly

**Verify pixels**
1. Open [list of pixels](https://app.asana.com/0/0/1206831531680581/f)
2. Make sure both old Settings and new Settings emit the relevant pixels

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPhone 14 Pro
* [ ] iPad

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:


Core/PixelExperiment.swift
```diff
@@ -0,0 +1,119 @@
...
+        if PixelExperiment(rawValue: allocatedCohort) == nil {
+            return true
+        }
+        return false
+    }
+
+    /// Enables this experiment for new users when called from the new installation path.
+    public static func install() {
+        // Disable the experiment until all other experiments are finished
+        // logic.install()
```
tomasstrba: To avoid having a PR open for weeks and keeping it conflict free, I'd rather merge it without enabling new Settings (without running an experiment). 

Once everyting is ready, I'd just open a new PR with this line uncommented and that would trigger the experiment after the merge.

Are there any strong objections about this approach?

Core/PixelExperiment.swift
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    static func cleanup() {
+        logic.cleanup()
+    }
+
+    // These are the variants. Rename or add/remove them as needed.  If you change the string value
+    //  remember to keep it clear for privacy triage.
+    case control
+    case newSettings = "variant1"
```
brindy: This will result in:

/t/pixel?cohort=variant1

or 

/t/pixel?cohort=control

depending on which cohort they end up in.  I'd recommend dropping the "variant1" so that it just ends up as "newSettings".  Mainly from a privacy triage perspective.  If you've already agreed that 'variant1' is good enough that's fine.

Core/PixelExperiment.swift
```diff
@@ -0,0 +1,119 @@
...
+        self.fire = fire
+    }
+
+    func install() {
+        isInstalled = true
+    }
+
+    func fireEnrollmentPixel() {
+        Pixel.fire(pixel: .pixelExperimentEnrollment,
+                   withAdditionalParameters: PixelExperiment.parameters)
```
brindy: Was the idea to call the fire function passed in?  In the original macOS code (which I see no longer has this) the idea was to be able to mock calls to pixels so that we could test it.

Talking of which - I don't see any unit tests? I think we should because now it's in the code base there's a good chance this will end up being the defacto standard approach moving forward.

Core/PixelExperiment.swift
```diff
@@ -0,0 +1,119 @@
...
+        guard isInstalled else { return nil }
+
+        if let allocatedCohort,
+           // if the stored cohort doesn't match, allocate a new one
+           let cohort = PixelExperiment(rawValue: allocatedCohort) {
+            return cohort
+        }
+
+        // For now, just use equal distribution of all cohorts.
+        let cohort = PixelExperiment.allCases.randomElement()!
```
tomasstrba: As Stefano pointed out in [his comment](https://app.asana.com/0/0/1206778521934560/1206933640056700/f), this logic will change based on new values we discussed

brindy (commented):
Mostly looking good just a couple of comments to discuss before I approve.

Also:

1/ "About", "Share Feedback", "DuckDuckGo on other platforms" items don't fire pixels - is that intended?  (Looks like it is)

2/ Email protection has a pixel in the new settings, but not the old settings?

Core/PixelExperiment.swift
tomasstrba: Sounds good 👍

Core/PixelExperiment.swift
tomasstrba: Since the experiment uses existing pixels and it adds many pixels on top, the current approach made more sense to me.

- Mocking would create approximately 2x30 methods which would be hard to maintain.
- Adding existing pixels to the experiment or removal of a pixel from the experiment would require to swap the standard Pixel.fire call with the specific method.

Do you think wrapper calls are that important? I don't have a strong opinion and can generate them (uncle GPT can help with that)

tomasstrba:
Thank you for the review, @brindy! We're close to the finish line with the design and copy approvals. As soon as the new copy is translated, I'll reach out to you for the final review.

1/ Correct, no pixels are planned for those sections
2/ Are you referring to `settingsEmailProtectionOpen`? My understanding is that the old Settings don't have that view. The email button in the old Settings is basically 'Manage Account' button in the new Settings and they both don't have a pixel.

tomasstrba:
@brindy, translations are in place for new Settings and I also corrected the main view which had a rendering problem because of a different length of labels.

Also, this PR contains translations for Privacy Pro, which is expected and approved by @miasma13.

Core/PixelExperiment.swift
brindy: I agree with your approach to firing pixels elsewhere, but regarding this class specifically there's a function passed into this class which you can use rather than calling it directly. If it's not needed then please remove it.  

But I assume point of that function was that it makes it "easy" to this logic from a unit test as you can do something like:

    func testPixelFiredOnEnrolment() {
    
       // create expectation
    
       let logic = PixelExperimentLogic() { pixel in
          // assert the expected pixel is fired
    
         // fulfil expectation 
       }
    
       logic.install()
       XCTAssertNotNil(logic.cohort)
    
       // wait for expectations 
    
    }

And I don't see those kind of tests here?

DuckDuckGo/SettingsPrivacyProtectionsView.swift
```diff
@@ -0,0 +1,79 @@
...
+import UIKit
+import Core
+
+struct SettingsPrivacyProtectionsView: View {
+
+    @EnvironmentObject var viewModel: SettingsViewModel
+
+    var body: some View {
+        Section(header: Text("Privacy Protections")) {
+            // Default Browser
```
brindy, line 30: These kind of comments are redundant.

DuckDuckGo/SettingsViewModel.swift
```diff
@@ -559,7 +735,31 @@ extension SettingsViewModel {
...
+    }
+
+    func openEmailSupport() {
+        UIApplication.shared.open(URL.emailProtectionSupportLink,
+                                  options: [:],
+                                  completionHandler: nil)
+    }
+
+    func openOtherPlatforms() {
+        UIApplication.shared.open(URL.apps,
```
brindy, line 752: This opens the default browser,  should it just open in a tab instead via quickLink?

brindy (commented):
These don't fire pixels on the old version (which might be correct):

* unprotected
* fireproof sites
* application lock
* keyboard
* long press previews
* open links in associated apps
* email 
* mac
* windows
* about 
* share feedback

I think we maybe had a miscommunication about the firePixel in the PixelExperiment class. Let me know if the latest comment doesn't make sense.

And also one comment about using a quick link instead of raw URL. 

Otherwise all looks good - will approve once we iron those out.

Core/PixelExperiment.swift
tomasstrba: I see, thanks for clarification! 👍 I missed this part. I have modified it slightly and added couple of tests to ensure pixel is fired when a cohort is allocated. Also, there is a verification the experiment logic returns appropriate pixel parameters.

DuckDuckGo/SettingsViewModel.swift
tomasstrba: Good catch! 👍 I was not aware of this. Corrected ✅

DuckDuckGo/SettingsPrivacyProtectionsView.swift
tomasstrba: I added header comments based on the original macOS implementation and find them pretty useful. In more complicated views, the navigation is a lot easier. For consistency, they are above each item. Do you mind if I leave them there?

tomasstrba:
@brindy, thanks for clarification, everything is clear now 👍 

**1) Pixels**
Most pixels you mentioned are not important for the experiment, as stated in [list of pixels](https://app.asana.com/0/0/1206831531680581/f). The email is slightly confusing, because the flow is different in the old Settings. In the end, I decided to include it there too so we have data for comparison in the experiment

**2) Unit Tests** 
You were right, this logic could see some testing, since it's important for the experiment. Added the one you suggested and couple of more to ensure parameters are correct.

**3) Quick Links**
Good catch! That would break a lot of things! Fixed ✅ 

The PR is ready for another review. Thanks for patience with this

brindy (approved):
LGTM!

PR #2605, opened 2024-03-16
23. Subscriptions:  - Feature Flags

afterxleep:
Task/Issue URL: https://app.asana.com/0/0/1206807550798011/f
Plan: https://app.asana.com/0/0/1206820386373649/f
BSK PR: https://github.com/duckduckgo/BrowserServicesKit/pull/717/files

**Description**:
- Adds feature flags for Subscriptions availability (Core logic in BSK)
- Adds compiler flags for Subscription environment

**How to test (Feature Availability)**:
- Build DDG-Alpha target
- Confirm Subscription is available in settings
- Go to Debug Settings > UnCheck Internal User State
- Confirm Subscription is NOT available in settings

**How to test (Disallowing Purchases)**:
- Invert conditional in `SubscriptionPagesUseSubscriptionFeature > getSubscriptionOptions()` so it returns an `SubscriptionOptions.empty`
- Go to `Settings > Privacy Pro > Get Privacy Pro`
- Confirm Purchase buttons are not visible. (See image below)

**Optional:**
Test with a custom privacy config URL > Flags available in: https://github.com/duckduckgo/privacy-configuration/pull/1887

Screenshot 2024-03-16 at 1 18 35 AM


DuckDuckGo/AppDependencyProvider.swift
```diff
@@ -22,6 +22,7 @@ import Core
 import BrowserServicesKit
 import DDGSync
 import Bookmarks
+import Subscription
```
quanganhdo, line 25: Does this need to be inside `#if SUBSCRIPTION` block?

miasma13: I think we should not, the imported class basically implements the runtime check equivalent to how we had previously used #if SUBSCRIPTION.
Plus for the release all targets/configuration will have the SUBSCRIPTION flag added (on macOS they already have, on iOS we only lack it in Release build config. atm). 
Plus it is likely that we will consider a cleanup of the DBP/NETWORK_PROTECTION/SUBSCRIPTION compile time feature flags some time after the launch.

quanganhdo (approved):

PR #2611, opened 2024-03-18
Web pixels (handlers + pixels)

federicocappelli:
Task/Issue URL: https://app.asana.com/0/72649045549333/1205469290776415/f
Implementation sub-task: https://app.asana.com/0/72649045549333/1206697233126329/f
CC: @miasma13

**Description**:

The FE needs to fire pixels after some user actions, this PR implements the web handlers where the related pixels are fired.

**Steps to test this PR**:
1. 4 pixels are implemented (see https://app.asana.com/0/72649045549333/1206697233126329/f)
2. For each Pixel recreate the action described in the `Trigger` and see if the right pixel is fired.

### DIFF
--- a/.swiftlint.yml
+++ b/.swiftlint.yml
@@ -7,6 +7,7 @@ disabled_rules:
   - todo
   - unused_capture_list
   - trailing_comma
+  - cyclomatic_complexity
 
 opt_in_rules:
   - closure_end_indentation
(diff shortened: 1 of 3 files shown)


.swiftlint.yml
```diff
@@ -7,6 +7,7 @@ disabled_rules:
   - todo
   - unused_capture_list
   - trailing_comma
+  - cyclomatic_complexity
```
federicocappelli, line 10: This rule is uber annoying and useless, even a simple method (`func handler(forMethodNamed methodName: String) -> Subfeature.Handler?`) with a switch and 12 cases is marked.
Additionally, `cyclomatic_complexity` is already disabled manually 8 times in the project

miasma13 (approved):
LGTM 👍

PR #2627, opened 2024-03-22
Rework the caching logic for subscription and entitlements 

miasma13:
Task/Issue URL: https://app.asana.com/0/414235014887631/1206909771441779/f

**Description**:

- Change default staleness time interval for subscription and entitlement cache to 20min
- When storing subscription in the cache use the date based on default interval or its date of renewalOrExpiry - whatever is earlier
- Make the VPN code use cache and its logic for doing fetch if data is stale
- On applicationDidBecomeActive refresh subscription details and entitlements
- In the app rely on subscription details and entitlements based on cache logic (cached if still valid else fetch)

---
###### Internal references:

### DIFF
--- a/DuckDuckGo/AppDelegate.swift
+++ b/DuckDuckGo/AppDelegate.swift
@@ -634,6 +634,10 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
             await autoClear?.applicationWillMoveToForeground()
             showKeyboardIfSettingOn = true
             syncService.scheduler.resumeSyncQueue()
+
+#if SUBSCRIPTION
+            await AccountManager().refreshSubscriptionAndEntitlements()
+#endif
         }
     }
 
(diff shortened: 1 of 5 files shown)


DuckDuckGo/AppDelegate.swift
```diff
@@ -435,7 +435,7 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
             }
 #endif
             SubscriptionPurchaseEnvironment.current = .appStore
-            await AccountManager().checkSubscriptionState()
+            await AccountManager().refreshSubscriptionAndEntitlements()
```
federicocappelli: What about moving this from didFinishLaunchingWithOptions to applicationDidBecomeActive so we can refresh the cache every time the app comes back from the background? didFinishLaunchingWithOptions could not happen for days

federicocappelli (approved):
Just an idea in the comment, LGTM, ready for BSK update

DuckDuckGo/AppDelegate.swift
quanganhdo: +1 to having this invoked more often.

quanganhdo (approved):

DuckDuckGo/AppDelegate.swift
miasma13: There is already call below in `applicationWillEnterForeground` but will move it to `applicationDidBecomeActive`

PR #2641, opened 2024-03-26
Fix forceful sign outs during purchase

miasma13:
Task/Issue URL: https://app.asana.com/0/0/1206926700687559/f

**Description**:
Due to recent changes we are force refreshing subscription and entitlements cache on `applicationDidBecomeActive`. 
We also did a check to sign out the user in case expired subscription was found.
This may cause problems during the purchase process as apple's modals are multiple times making the app non active and then active again causing the `applicationDidBecomeActive` to be called. 

When doing the purchase we attempt to retrieve past account from Apple ID transaction. If the account has been used previously and has an expired subscription there is a change we will trigger a sign out during the purchase flow. This will result in purchase seemingly going through, the purchase view being preemptively dismissed and the settings being in unsubscribed state. Attempt at restoring the purchase fixes it and brings back the recent purchase.

**Steps to test this PR**:
- Smoke test the successful purchase using Apple ID account that has expired/cancelled subscriptions
- Confirm if when the subscription is cancelled and the app is left in the settings after the cancellation is registered in the BE and coming back to the app the state will be reloaded accordingly (sign out) 

---
###### Internal references:

### DIFF
--- a/DuckDuckGo/AppDelegate.swift
+++ b/DuckDuckGo/AppDelegate.swift
@@ -524,8 +524,6 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
                                                                                            cachePolicy: .reloadIgnoringLocalCacheData) {
                 if subscription.isActive {
                     DailyPixel.fire(pixel: .privacyProSubscriptionActive)
-                } else {
-                    accountManager.signOut()
                 }
             }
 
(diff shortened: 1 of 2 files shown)


federicocappelli (approved):
LGTM, I tried to simulate possible sign-off but all seems ok.

PR #2664, opened 2024-04-03
Fix VPN internal user bugs

samsymons:
Task/Issue URL: https://app.asana.com/0/414235014887631/1206986434624665/f
Tech Design URL:
CC:

**Description**:

This PR fixes an issue where the VPN Thank You modal was interrupting the setup flow.

**Steps to test this PR**:
1. Go into AppDelegate and change `stopTunnelAndShowThankYouMessagingIfNeeded` to always return true on its `vpnFeatureVisibility` check - basically we want to always make the alert show
2. Run the app and check that you get the alert - if you don't, you probably have an account already signed in
3. Sign into an account and try to activate the VPN
4. Check that it works

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPhone 14 Pro
* [ ] iPad

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:

### DIFF
--- a/DuckDuckGo/AppDelegate.swift
+++ b/DuckDuckGo/AppDelegate.swift
@@ -524,7 +524,9 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
                 let controller = NetworkProtectionTunnelController()
 
                 if await controller.isConnected {
-                    DailyPixel.fireDailyAndCount(pixel: .privacyProVPNBetaStoppedWhenPrivacyProEnabled)
+                    DailyPixel.fireDailyAndCount(pixel: .privacyProVPNBetaStoppedWhenPrivacyProEnabled, withAdditionalParameters: [
+                        "reason": "thank-you-dialog"
+                    ])
                 }
 
                 await controller.stop()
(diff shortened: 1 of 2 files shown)


DuckDuckGo/AppDelegate.swift
```diff
@@ -512,14 +512,21 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     }
 
     private func stopTunnelAndShowThankYouMessagingIfNeeded() {
+        if AccountManager().isUserAuthenticated {
+            tunnelDefaults.vpnEarlyAccessOverAlertAlreadyShown = true
+            return
+        }
```
samsymons, line 518: This is the change that fixes all of this - the issue was caused by this modal trying to appear during the setup flow, but the alert couldn't be presented so we were never able to set this flag. Now we make sure to set this flag if this is attempted any time after the user already has an account.

quanganhdo (approved):
Works well for me 👍

PR #2666, opened 2024-04-03
fix double clearing on autoclear

brindy:
Task/Issue URL: https://app.asana.com/0/414709148257752/1206731143449260/f
Tech Design URL: 
CC:

**Description**:
* Refactoring to inject tabs model into the main view controller
* Prevents double clearing when using auto-clear + timer
* Also fixes [Bug: Unable to launch URL properly when auto clear enabled](https://app.asana.com/0/414235014887631/1206454582708636/f)

**Steps to test this PR**:
***Test standard fire button and fire proofing usage***
1. Open some tabs, set cookies (e.g. login to a site), use fire button ensure data is cleared
2. Open some tabs, set cookies (e.g. login to a site), add to fire proofing, ensure data except fireproofed is cleared
3.  Open some tabs, set cookies (e.g. login to a site), terminate the app and reload, ensure tab loads as expected

***Autoclear with Terminate***
1. Update settings to enable auto clear on terminate
2. Open some tabs, set cookies (e.g. login to a site)
3. Terminate the app
4. Launch the app, ensure tabs and data are cleared

***Autoclear with Terminate and Delay***
1.  Update settings to enable auto clear on terminate with a delay
2. Hack the code to return true for the selected delay in the `shouldClearData` function
3. Background the app
4. Use another app (e.g news) to launch a URL in DDG (either through default browser or by share extension)
5. Ensure tabs and data are cleared

***Autoclear with tabs only***
1. Repeat above tests but with tabs only selected in auto clear settings
2. Ensure tabs are closed but data is retained


DuckDuckGo/MainViewController.swift
```diff
@@ -692,40 +709,6 @@ class MainViewController: UIViewController {
...
-
-        let isPadDevice = UIDevice.current.userInterfaceIdiom == .pad
-
-        let tabsModel: TabsModel
-        if let settings = AutoClearSettingsModel(settings: appSettings) {
-            // This needs to be refactored so that tabs model is injected and cleared before view did load,
-            //  but for now, ensure this happens in the right order by clearing data here too, if needed.
-            tabsModel = TabsModel(desktop: isPadDevice)
-            tabsModel.save()
-            previewsSource.removeAllPreviews()
```
bwaresiak, line 705: Don't we need to still remove previews?

DuckDuckGo/MainViewController.swift
```diff
@@ -998,6 +982,7 @@ class MainViewController: UIViewController {
         currentTab?.progressWorker.progressBar = nil
         currentTab?.chromeDelegate = nil
             
+        tab.delegate = self
```
bwaresiak: I see what you mean by delaying the moment delegate is attached. The problem here is that before delegate is set, the view is apparently already loaded. The most likely place that could fail are the ones that use delegate methods which return a value like `tabDidRequestSearchBarRect` or `tabCheckIfItsBeingCurrentlyPresented` which could glitch the TabUI itself during its setup depending on the code flows (there could be other glitches too, but these are most likely). 

We could investigate code paths to "fix" this now, but overall this is not a foolproof solution. I'm afraid that with TabManager being a Factory, a binding must occur as it was.

But looking at the code, I think it could remain as it was - just maybe don't pass self it through TabManager init as it's awkward for weak vars but explicitly invoke a manager.tabDelegate = self in MVC init.

DuckDuckGo/TabViewController.swift
```diff
@@ -74,7 +74,11 @@ class TabViewController: UIViewController {
         }
     }
     
-    weak var delegate: TabDelegate?
+    weak var delegate: TabDelegate? {
+        didSet {
+            assert(delegate != nil)
```
bwaresiak: This is getting triggered when using fire button, as `prepareForDataClearing` sets delegate to nil

bwaresiak (approved):
Apart from that assert: FIreproofing works. I could not reproduce startup issue either. Smoke tested fire button, tab creation, auto clear etc. Good job!

PR #2680, opened 2024-04-04
Fix missing tab delegate on launch

brindy:
Task/Issue URL: https://app.asana.com/0/414235014887631/1207004361509188/f
Tech Design URL:
CC:

**Description**:
Fixes a bug in recent PR which means the settings menu doesn't work when launching the app with a tab selected.

**Steps to test this PR**:
1. Open a tab and browse to a site
2. Terminate the app
3. Launch the app - page should load
4. Access browsing menu, select settings - settings should load
5. Open some more tabs
6. Close some tabs
7. Use the tab switcher a bit


samsymons (approved):
LGTM, bashed on the tab switching and state restoration a bunch, couldn't reproduce the original issue.

PR #2711, opened 2024-04-11
soft revert history suggestions

brindy:
Task/Issue URL: https://app.asana.com/0/414235014887631/1207038972999258/f
Tech Design URL:
CC:

**Description**:
Reverts change to new logic for existing users not in the experiment. 

**Steps to test this PR**:
1. Run the app and check you're not allocated to a variant (e.g. because you're a return user, which should be the case unless you reset the simulator) - the output from Xcode will say 
2. Ensure previous logic works, ie matching bookmarks should appear in the top two slots of suggestions.
3. Run the app with the `md` variant (via launch arguments)
4. Ensure new logic with history works
5. Run the app with the `mc` variant
6. Ensure new logic without history works
7. Reset the simulator so you're a new user and check you get allocated to a variant

### DIFF
--- a/Core/DefaultVariantManager.swift
+++ b/Core/DefaultVariantManager.swift
@@ -26,8 +26,8 @@ extension FeatureName {
     // Define your feature e.g.:
     // public static let experimentalFeature = FeatureName(rawValue: "experimentalFeature")
 
-    // Experiment coming soon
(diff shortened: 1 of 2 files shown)


dus7 (approved):
LGTM

PR #2720, opened 2024-04-11
Add support for uploading crash reports to Sentry

ayoy:
Task/Issue URL: https://app.asana.com/0/72649045549333/1205862129634403/f

**Description**:
This change introduces uploading crash reports to DDG's Sentry instance. Users are prompted
to opt in to crash collection upon launching the app after it crashed. A user can decide either
to always send reports or never send them. A switch is added to Settings to control their preference.

**Steps to test this PR**:
1. Install the ipa file uploaded to this Asana task.
2. Identify as internal user.
3. Go to Debug Menu -> Crash (fatal error)
4. Relaunch the app
5. Verify that you're being presented with a crash reporting dialog
6. Verify that expanding and collapsing the dialog works as expected (note that there may be a delay/lag when rendering crash report data).
7. Verify that if you close the dialog and crash the app again, the dialog reappears on the next app launch.
8. Verify that if you choose "Always Send" or "Never Send", the dialog won't appear again after a subsequent crash.
9. Select Debug Menu -> Reset Upload Crash Logs Prompt to have the dialog appear again as needed.

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPhone 14 Pro
* [ ] iPad

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:


DuckDuckGo/AppDelegate.swift
```diff
@@ -133,8 +136,20 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
             Configuration.setURLProvider(AppConfigurationURLProvider())
         }
 
-        CrashCollection.start {
-            Pixel.fire(pixel: .dbCrashDetected, withAdditionalParameters: $0, includedParameters: [])
+        crashCollection.start { pixelParameters, payloads, sendReport in
+            pixelParameters.forEach { params in
```
ayoy, line 137: Previously the callback was called for every crash report found. Now we have a single callback that can contain multiple crash reports. We still send pixels for all crash reports, and we present onboarding for Sentry as needed using the most recent crash report (this is the same behavior as on macOS for both DMG and App Store builds).

DuckDuckGo/CrashCollectionOnboardingView.swift
```diff
@@ -0,0 +1,167 @@
...
+                                        .foregroundColor(Color(designSystemColor: .container))
+                                        .frame(maxWidth: .infinity, maxHeight: .infinity)
+                                        .cornerRadius(4.0)
+
+                                    Text(reportDetails)
+                                        .multilineTextAlignment(.leading)
+                                        .font(.crashReport)
+                                        .foregroundColor(Color(designSystemColor: .textSecondary))
+                                        .padding(24)
+                                }
```
samsymons, line 76: I was wondering whether it would make sense to put this in a scroll view which is horizontally scrollable since text wrapping can make it awkward to read, but TBH there's no right approach here – happy to leave this as-is.

DuckDuckGo/CrashCollectionOnboardingView.swift
```diff
@@ -0,0 +1,167 @@
...
+            }
+            .background(Color(designSystemColor: .backgroundSheets))
+        }
+    }
+
+    var contents: some View {
+        VStack(spacing: 0) {
+            ScrollView {
+                VStack(spacing: 24) {
+                    Image("Breakage-128")
```
samsymons: We could use [`Image(decorative:)`](https://developer.apple.com/documentation/swiftui/image/init(decorative:bundle:)) here to make sure that VoiceOver API doesn't read the image name (it does this in the current implementation).

samsymons (approved):
LGTM, no blocking comments. The expansion of the crash report details was pretty slow on my iPhone X, but I don't think it's a blocker since this isn't a core flow and is only presented to the user once.

Nice work on this, really excited for it to ship!

DuckDuckGo/CrashCollectionOnboardingView.swift
ayoy: That was my first idea and implementation, but we agreed with Bryan that it actually shouldn't be scrolled horizontally ([see the discussion](https://app.asana.com/0/0/1206992056177669/1207072176348345/f)).

That said, the scroll view approach makes rendering faster (not fast, but faster) 🤷

DuckDuckGo/CrashCollectionOnboardingView.swift
ayoy: Of course :) thanks for pointing that out :) I'll fix it right away.

ayoy:
Thanks for the review @samsymons! Yes, the crash report is very slow to load (also relatively slow on a modern device), but there's not much we can do about it. The slowness comes from the UI rendering, not from processing crash report payload – so it can't be offloaded to a background thread. Showing a spinner wouldn't really work either (or the spinner wouldn't spin).

The working solution would be to break the text into lines and load in a list with reusable cells but that's too much :) But as you said, it's only presented once to the user and also it's not a core flow.

PR #2722, opened 2024-04-12
Require device auth to be set in order to use Sync

bwaresiak:
Task/Issue URL: https://app.asana.com/0/414235014887631/1206876054968715/f
Tech Design URL:
CC:

**Description**:

Require device auth to be set in order to use Sync

**Steps to test this PR**:

Test this on a device that has no password set, or test by changing `UserAuthenticator.canAuthenticate()` to return false.

When sync is not set up:
* [x] Connecting with device should be protected.
* [x] Setting sync for current device should be protected.
* [x] Recovering data should be protected.

When sync is set up:
* [x] Connecting with device should be protected.
* [x] Saving recovery code should be protected.

**Copy Testing**:

**Device Testing**:

* [ ] iPhone 
* [ ] iPad

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16

---
###### Internal references:

### DIFF
--- a/LocalPackages/SyncUI/Sources/SyncUI/Views/SyncSettingsViewExtension.swift
+++ b/LocalPackages/SyncUI/Sources/SyncUI/Views/SyncSettingsViewExtension.swift
@@ -21,6 +21,8 @@ import DesignResourcesKit
 import DuckUI
 import SwiftUI
 
+// swiftlint:disable file_length
+
 extension SyncSettingsView {
 
     @ViewBuilder
@@ -413,3 +415,5 @@ extension View {
         closure(self)
     }
 }
+
+// swiftlint:enable file_length
--- a/LocalPackages/SyncUI/Sources/SyncUI/Views/SyncSettingsViewExtension.swift
+++ b/LocalPackages/SyncUI/Sources/SyncUI/Views/SyncSettingsViewExtension.swift
@@ -74,11 +74,6 @@ extension SyncSettingsView {
 
     @ViewBuilder
     func otherOptions() -> some View {
-        let alertBinding = Binding<Bool>(
-            get: { model.shouldShowPasscodeRequiredAlert },
-            set: { value in model.shouldShowPasscodeRequiredAlert = value }
-        )
-        
         Section {
             Button(UserText.syncAndBackUpThisDeviceLink) {
                 Task { @MainActor in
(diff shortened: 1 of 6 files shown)


LocalPackages/SyncUI/Sources/SyncUI/Views/SyncSettingsViewExtension.swift
```diff
@@ -97,6 +107,18 @@ extension SyncSettingsView {
                 })
             })
             .disabled(!model.isAccountRecoveryAvailable)
+            .alert(isPresented: alertBinding) {
```
ayoy: `alertBinding` can be replaced with `$model.shouldShowPasscodeRequiredAlert` here (and save you the hassle of defining the binding).

```suggestion
            .alert(isPresented: $model.shouldShowPasscodeRequiredAlert) {
```

LocalPackages/SyncUI/Sources/SyncUI/Views/SyncSettingsViewExtension.swift
```diff
@@ -97,6 +107,18 @@ extension SyncSettingsView {
...
+                Alert(
+                    title: Text("Secure Your Device to Use Sync & Backup"),
+                    message: Text("A device password is required to use Sync & Backup."),
+                    dismissButton: .default(Text("Go to Settings"), action: {
+                        UIApplication.shared.open(URL(string: UIApplication.openSettingsURLString)!,
+                                                  options: [:],
+                                                  completionHandler: nil)
+                        model.shouldShowPasscodeRequiredAlert = false
+                    })
+                )
```
ayoy: This alert is not presented when Sync is enabled. To have it show up also when "Sync With Another Device" under devices list is tapped, this alert should be defined in SyncSettingsView as a modifier on the top-level List (around the line 91 in SyncSettingsView.swift).

Additionally, the dialog is not shown when tapping "Save Recovery PDF" when Sync is enabled, because that action calls delegate directly instead of wrapping it in `commonAuthenticate()`:

```swift
func saveRecoveryPDF() {
    delegate?.shareRecoveryPDF()
}
```

This is probably because the delegate (the VC) performs authentication on its own, but it doesn't report failure back to the view model.

ayoy (commented):
Hey @bwaresiak, the code looks solid but I could not get alerts to display while Sync was enabled. Following our Zoom chat I'm leaving my comments here. Thanks!

ayoy (approved):
Works great now, thanks! :)

PR #2789, opened 2024-04-25
Fix bug that shows selected tab in strange position in the switcher

brindy:
Task/Issue URL: https://app.asana.com/0/414709148257752/1207153191136827/f
Tech Design URL:
CC:

**Description**:

Fix bug that shows selected tab in strange position in the switcher

**Steps to test this PR**:
1. Open a new few tabs on various pages
2. Use the tab switcher, ensure the current tab is as central as it can be
3. Change tab switcher modes and retest
4. Also check the two options when going between home pages


amddg44 (approved):
LGTM! Tested on iPhone & iPad in both switcher modes, portrait & landscape; one and many tabs, all perfect!

PR #2833, opened 2024-05-08
support marketplace kit

brindy:
Task/Issue URL: https://app.asana.com/0/0/1207266309510426/f
Tech Design URL:
CC:

**Description**:
Adds marketplace kit support.  Installing alternative app stores doesn't appear to work with an adhoc build so need to test in the wild.  Internal users only.

The final version should use our current link handling logic, this direct approach is just to validate it works.

**Steps to test this PR**:
1. With an iOS **17.4** device or simulator with internal user flag enabled access https://chrisbrind.rocks/testing/ and click on the marketplace URL and you should get a toast.  Check that tapping COPY copies the URL into the pasteboard.
2. Try the above where any of the conditions are false (e.g. older version of iOS, internal user flag off, etc).  Nothing should happen when clicking on the link.

---
###### Internal references:


fastlane/Matchfile
```diff
@@ -6,7 +6,6 @@ platform "ios"
 type "appstore"
 app_identifier ["com.duckduckgo.mobile.ios", "com.duckduckgo.mobile.ios.ShareExtension", "com.duckduckgo.mobile.ios.OpenAction2", "com.duckduckgo.mobile.ios.Widgets", "com.duckduckgo.mobile.ios.NetworkExtension"]
 generate_apple_certs false
-template_name "Default Web Browser iOS (Dist)"
```
brindy, line 9: This forces it to use the wrong profile.  The entitlements are now driven by the app id configuration.

aataraxiaa (approved):
LGTM! 🚀

Minor clarification as discussed on MM: When clicking on the link with internal user disabled, something does happen, which is that the following toast is displayed

Screenshot 2024-05-08 at 13 19 55

PR #2845, opened 2024-05-13
Subscription attribution

alessandroboron:
Task/Issue URL: https://app.asana.com/0/0/1205406454505227/f

**Description**:

This PR adds a new pixel to track the attribution when a Privacy Pro subscription is purchased.
The new Pixel will have the device Locale and an `origin` string as parameters. The origin will be defined as a query parameter in the `/pro` URLs. 

Currently, when a `duckduckgo.com/pro` URL is input in the browser we redirect to `duckduckgo.com/subscriptions/welcome`. We need to ensure that if we have `/pro?origin=test` we redirect to `/subscriptions/welcome?origin=test`.

If the `origin` is defined, the pixel will send the origin parameter in its request. If the `origin` is not defined the Pixel will be fired without the `origin` parameter.

Changes in the PR:

I refactored the code previously written to inject the origin and the redirect URL.

I covered as much as I could with tests. Unfortunately, not everything is testable, and the AppStorePurchaseFlow dependency cannot be controlled.

**Note:**

The PR points temporarily to a BSK branch. I will update it once the BSK PR is merged in. 

**Steps to test this PR**:
1. Create a sandbox user to test in App Purchase on [appstoreconnect](https://appstoreconnect.apple.com/). You can use aliases so you can create a sandbox user with an email that looks like <your_ddg_username>[+sandbox@duckduckgo.com](mailto:+sandbox@duckduckgo.com).
2. Ensure that the environment used is `staging` and clear the subscription details. 
3. Type in the address bar www.duckduckgo.com/pro?origin=test
4. Ensure that the url is redirected to https://duckduckgo.com/subscriptions?environment=staging&origin=test.
5. Click the 9.99 / month or 99.99 / month button.
6. Click on the Subscribe button.
7. If the sandbox user is not logged in Apple will ask you to sign in with that account.
8. Accept Apple alert about subscription being successful.
9. After the in app purchase is successful search on the console log for `m_subscribe` and make sure that the locale and origin parameters are sent.

**Copy Testing**:

**Orientation Testing**:

* [x] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [x] iPhone X
* [ ] iPhone 14 Pro
* [ ] iPad

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [x] iOS 16

---
###### Internal references:


DuckDuckGo/TabURLInterceptor.swift
```diff
@@ -79,13 +79,23 @@ extension TabURLInterceptorDefault {
...
+        switch interceptedURL {
+
             // Opens the Privacy Pro Subscription Purchase page (if user can purchase)
             case .privacyPro:
                 if SubscriptionPurchaseEnvironment.canPurchase {
-                    NotificationCenter.default.post(name: .urlInterceptPrivacyPro, object: nil)
+                    // If URL has an `origin` query parameter, append it to the `subscriptionPurchase` URL.
+                    // Also forward the origin as it will need to be sent as parameter to the Pixel to track subcription attributions.
+                    let originQueryItem = queryItems?.first(where: { $0.name == AttributionParameter.origin })
+                    let purchaseURL = URL.subscriptionPurchase
```
miasma13: I do not like that the `URL.subscriptionPurchase` appears here in the interception logic. I think it fits more to be contained at the purchase flow level (and appending it with the origin param can happen there).
WDYT about removing the passing of the URL in the `.urlInterceptPrivacyPro` notification and instead replace it with just raw `origin` value?

alessandroboron: To be honest that was my first approach. Then I thought to be more explicit by passing the URL but I can see where you’re coming from. I’ve implemented the changes and pass just the origin.

DuckDuckGo/Subscription/ViewModel/SubscriptionContainerViewModel.swift
```diff
@@ -31,11 +32,18 @@ final class SubscriptionContainerViewModel: ObservableObject {
...
-         subFeature: SubscriptionPagesUseSubscriptionFeature = SubscriptionPagesUseSubscriptionFeature()) {
+    init(
+        origin: String?,
+        userScript: SubscriptionPagesUserScript,
+        subFeature: SubscriptionPagesUseSubscriptionFeature
+    ) {
+        var purchaseURL = URL.subscriptionPurchase
+        if let origin {
+            purchaseURL = purchaseURL.appendingParameter(name: AttributionParameter.origin, value: origin)
+        }
```
miasma13: I'd probably avoid handling creation of `purchaseURL` here. Instead just take the `origin`  and and pass it into `SubscriptionFlowViewModel`'s init (instead of `purchaseURL`)

miasma13 (approved):

DuckDuckGo/Subscription/ViewModel/SubscriptionContainerViewModel.swift
alessandroboron: Done!

PR #2850, opened 2024-05-14
Lower unit test workflow timeout to 15 minutes

samsymons:
These runs take around 7 minutes, so the current 30 minute timeout is excessive. CI does occasionally get stuck so we have it wait so long to time out when we know after 15 minutes that something is wrong.

Task/Issue URL: https://app.asana.com/0/414235014887631/1207310444690858/f
Tech Design URL:
CC:

**Description**:

This PR lowers the PR Checks workflow timeout to 15 minutes, down from 30. This is because our workflows typically run in under 10 minutes, so waiting an extra 20 minutes on top of that is excessive.

See [this workflow run](https://github.com/duckduckgo/iOS/actions/runs/9077404703) for an example where a lower timeout would have saved us time.

**Steps to test this PR**:
1. Just check that there are no other 30 minute timeouts that we should be changing - I deliberately focused on the unit test workflow for this change since it's the one that occasionally gets stuck

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16

---
###### Internal references:

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -40,7 +40,7 @@ jobs:
     name: Unit Tests
 
     runs-on: macos-13-xlarge
-    timeout-minutes: 30
+    timeout-minutes: 15
 
     outputs:
       commit_author: ${{ steps.fetch_commit_author.outputs.commit_author }}


brindy (approved):
Let's do this.

PR #2886, opened 2024-05-22
remove gradient logic and transaction flush

brindy:
Task/Issue URL: https://app.asana.com/0/414709148257752/1207372554889903/f
Tech Design URL:
CC:

**Description**:
Clean up the progress bar now that we're only using a single color and remove the unnecessary call to CATransaction flush 

**Steps to test this PR**:
1. Load some pages
2. Switch between tabs and load pages
3. Swipe between tabs and load pages


brindy:
We're not sure about losing the gradient and unclear on the impact of removing the flush so will come back to it later.

PR #2891, opened 2024-05-24
fire RMF message updated notification if we make any change to the store

brindy:
Task/Issue URL: https://app.asana.com/0/414235014887631/1207400212393238/f
Tech Design URL:
CC:

**Description**:
Refresh if there's any save to the RMF store.

Note that logo hiding/showing is a bit fuzzy at the moment as it's done a in a few places, but we can fix that in the upcoming new tab page project.

**Steps to test this PR**:
1. Add some code to the app delegate refresh the RMF every 10 seconds 
2. Point the RMF at some json you can change easily
2. Open a new home view
3. Update json to add a message
4. Wait for message to appear
5. Update json to remove the message
6. Wait for message to disappear


amddg44 (approved):
LGTM, thanks!

PR #2892, opened 2024-05-24
Remove usage of token store from iOS VPN

quanganhdo:
Task/Issue URL: https://app.asana.com/0/414235014887631/1207178910368608/f
Tech Design URL:
CC:

**Description**:

Following Privacy Pro integration, the iOS version no longer uses the VPN auth token to interact with the backend. During the transition, the existing token store is left intact but now it's fully deprecated.

This task removes the token store usage from Startup Options and other places.

**Steps to test this PR**:
1. Make sure the VPN still works properly
2.

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPhone 14 Pro
* [ ] iPad

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:


DuckDuckGo/NetworkProtectionAccessController.swift
```diff
@@ -105,7 +105,9 @@ struct NetworkProtectionAccessController: NetworkProtectionAccess {
     }
 
     func revokeNetworkProtectionAccess() {
+#if os(macOS)
         try? networkProtectionKeychainTokenStore.deleteToken()
+#endif
```
samsymons: Any reason to keep the code inside the `os(macOS)` checks around given that this is iOS client-side code?

DuckDuckGo/AppDependencyProvider.swift
```diff
@@ -149,11 +149,15 @@ class AppDependencyProvider: DependencyProvider {
...
             return { nil }
         }()
+#if os(macOS)
         networkProtectionKeychainTokenStore = NetworkProtectionKeychainTokenStore(keychainType: .dataProtection(.unspecified),
                                                                                   serviceName: "\(Bundle.main.bundleIdentifier!).authToken",
                                                                                   errorEvents: .networkProtectionAppDebugEvents,
                                                                                   isSubscriptionEnabled: accountManager.isUserAuthenticated,
                                                                                   accessTokenProvider: accessTokenProvider)
+#else
+        networkProtectionKeychainTokenStore = NetworkProtectionKeychainTokenStore(accessTokenProvider: accessTokenProvider)
```
samsymons, line 159: Ideally we could remove this entirely from iOS, though I understand we still have some code at the BSK layer that expects it.

samsymons (changes requested):
Left a couple comments, and also it looks like tests are failing

PacketTunnelProvider/NetworkProtection/NetworkProtectionPacketTunnelProvider.swift
```diff
@@ -285,11 +285,16 @@ final class NetworkProtectionPacketTunnelProvider: PacketTunnelProvider {
...
             }
-            return { nil } }()
+            return { nil }
+        }()
+#if os(macOS)
         let tokenStore = NetworkProtectionKeychainTokenStore(keychainType: .dataProtection(.unspecified),
                                                              errorEvents: nil,
                                                              isSubscriptionEnabled: true,
                                                              accessTokenProvider: accessTokenProvider)
+#else
```
samsymons: We can probably simplify this down to only use the `else` case since the `if` case will never be true, right?

samsymons (approved):
LGTM, I think there are some small pieces of code we can simplify but they aren't blockers.

PR #2971, opened 2024-06-19
Custom DNS for VPN

quanganhdo:
Task/Issue URL: https://app.asana.com/0/1107339934162485/1204085249612942/f
Tech Design URL:
CC:

**Description**:

Adds custom DNS settings for VPN.

**Steps to test this PR**:
1. Go to VPN Settings > DNS Server
2. Input a custom DNS > Apply
3. Verify

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPhone 14 Pro
* [ ] iPad

**OS Testing**:

* [ ] iOS 14
* [ ] iOS 15
* [ ] iOS 16

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:


DuckDuckGo/NetworkProtectionDNSSettingsViewModel.swift
```diff
@@ -0,0 +1,79 @@
...
+            DailyPixel.fireDailyAndCount(pixel: .networkProtectionDNSUpdateCustom)
+        } else {
+            settings.dnsSettings = .default
+            DailyPixel.fireDailyAndCount(pixel: .networkProtectionDNSUpdateDefault)
+        }
+    }
+
+    func updateApplyButtonState() {
+        if isCustomDNSSelected {
+            isApplyButtonEnabled = !customDNSServers.isEmpty && customDNSServers.isValidIpv4Host
```
samsymons, line 71: `isValidIpv4Host` will use leading zeroes if you don't provide a full address, i.e. if you enter the string `1` it will allow you to apply it and expand it to `0.0.0.1`. This expansion behaviour is fine since it expands to a legit IP, but it may be confusing for users since we don't show the expanded value - a knowledgable user would likely guess correctly, but I doubt all users would.

When the "Apply" button is tapped, we could use the IP string to initialize an `IPv4Address` then get the full string being used out of that, and save _that_ to storage. In the case of the string `1`, you can call `String(reflecting: address)` to get back `0.0.0.1`.

samsymons: One other kind of strange thing that the `IPv4Address` string initializer does is take a subset of the string you provide it to extract the IP. For example, you can use `IPv4Address("123412341234123412341234123412341234")` and get back an IP address of `37.12.1.242`, but in the UI we will show you that full obviously-incorrect string.

quanganhdo: Interesting. I updated it so that what you see is what you get at least.

samsymons (approved):

PR #2975, opened 2024-06-20
update copy for autocomplete settings and some minor clean up

brindy:
Task/Issue URL: https://app.asana.com/0/1206767823185684/1207592116853788/f
Tech Design URL:
CC:

**Description**:
Explicitly call out what recently visited sites toggle does to make it clearer to users so they are less compelled to just disable suggestions entirely.  See [Copy task](https://app.asana.com/0/38424471409662/1207592116853789/f)

**Steps to test this PR**:
1. Edit `HistoryMessageManager` and set `shouldShow` to return true
2. Launch app and start typing in the search field.  Message should show with updated copy.
3. Open settings and go to General. Private Search section should have correct copy for each toggle and no footer
4. Open settings and go to Private Search. Search Suggestions section should have correct copy for each toggle and no footer


amddg44 (approved):
LGTM!

PR #2980, opened 2024-06-21
Add feature flag for improved New Tab Page 

dus7:
Task/Issue URL: https://app.asana.com/0/1206226850447395/1207630604691910/f
Tech Design URL:
CC:

**Description**:

Defines an internal flag to keep existing implementation and make new one possible to enable via debug menu. This allows to do incremental development changes into main branch.

**Steps to test this PR**:
1. Make sure there are no changes to Home Screen with and without internal user flag set.
1. Setting local flag via debug menu should display a new home tab controller with "Empty" text in a new tab.

---
###### Internal references:


DuckDuckGo/HomeTabViewController.swift
```diff
@@ -0,0 +1,56 @@
+//
+//  ImprovedHomeViewController.swift
```
brindy: SwiftLint - presume you renamed the file or something during dev

DuckDuckGo/HomeTabViewController.swift
```diff
@@ -0,0 +1,56 @@
...
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import SwiftUI
+
+final class HomeTabViewController: UIHostingController<HomeTabView>, HomeViewControllerProtocol {
```
brindy: Wonder if we should just go all in on calling it 'new tab page' ?  e.g.

`NewTabPageViewController` 

?

DuckDuckGo/HomeViewControllerProtocol.swift
```diff
@@ -0,0 +1,35 @@
...
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import UIKit
+
+protocol HomeViewControllerProtocol: UIViewController {
```
brindy: Let's avoid calling protocols "Protocol" - I know we have some in the code base already, but imo they need refactored.  From the API Design Guidelines:

* Protocols that describe what something is should read as nouns (e.g. Collection).

* Protocols that describe a capability should be named using the suffixes able, ible, or ing (e.g. Equatable, ProgressReporting).

I think the protocol here is simply `NewTabPage` but open to suggestion.

DuckDuckGo/RootDebugViewController.swift
```diff
@@ -41,6 +42,7 @@ class RootDebugViewController: UITableViewController {
         case openVanillaBrowser = 670
         case resetSendCrashLogs = 671
         case refreshConfig = 672
+        case homeTabImprovements = 674
```
brindy: I'm not overly keen on this.  Would be better to capture the cell as an outlet and check that. However, looks like we're invested in this approach.  We should probably refactor this debug view controller at some point. :) 

(No action required)

brindy (changes requested):
Hey, looks good and works. Needs some renaming, SwiftLint, and resolve conflicts. Give me a ping for a follow up review.

DuckDuckGo/HomeTabViewController.swift
dus7: Yeah, probably. I was hesitating and left one similar to what's now, but maybe it's better to keep consistency with internal names.

DuckDuckGo/MainViewController.swift
```diff
@@ -83,14 +83,14 @@ class MainViewController: UIViewController {
     }()
     
     var homeViewController: HomeViewController?
-    var homeTabViewController: HomeTabViewController?
-    var homeController: HomeViewControllerProtocol? {
-        homeViewController ?? homeTabViewController
+    var newTabPageViewController: NewTabPageViewController?
+    var homeController: NewTabPage? {
```
dus7, line 87: I left `homeController` here deliberately to avoid extensive renaming. It'll be straighten out after the old code is cleaned up.

brindy (approved):
FYI I pushed a swiftlint fix

LGTM!

PR #2986, opened 2024-06-24
Remove noise from wake pixel

diegoreymendez:
Task/Issue URL: https://app.asana.com/0/1206580121312550/1207646371500325/f

macOS PR: https://github.com/duckduckgo/macos-browser/pull/2899
BSK PR: https://github.com/duckduckgo/BrowserServicesKit/pull/860

## IMPORTANT

We're merging this against `release/7.126.0` as soon as the branch goes live... I'll be updating the target branch before merging.

## Description

We're emoving known noise from wake pixels, and tunnel update pixels, plus potentially removing unnecessary calls to backend and other code handling that shouldn't happen when the VPN is disconnected.

## Testing

I couldn't really test this for iOS because I couldn't figure out how to trigger the device going to sleep.

That said I believe it's reasonable to expect the logic makes sense, [as long as it works well for macOS too](https://github.com/duckduckgo/macos-browser/pull/2899).

Let me know if you disagree or if you understand how we can force the device to sleep.

---
###### Internal references:


samsymons (approved):
This looks good to me, though note that I can reproduce a wake attempt and success pixel pair firing off every 10 seconds or so. For this to occur the device has to be unplugged from power entirely, and have the screen locked. If you want to see the pixels my device was firing, look up `qs_appVersion:7.125.2.2` in Kibana.

As long as you're okay with the pixel volume, we can merge this, but let me know if you want to take some action to prevent iOS from sending them so often.

PR #2989, opened 2024-06-25
[DuckPlayer] - 2. Overlay Views

afterxleep:
Task/Issue URL: https://app.asana.com/0/1204099484721401/1207650110399906/f

**Description**:
- Implements Overlay Script for iOS
- Allows user to see video inline or open Duck Player
- Cleans up UserScripts (messaging) to match with what we do in other places on iOS (like subscriptions)

ℹ️   Known issue:  The FE is sometimes starting navigation instead of pushing a WKMessage, which causes you to have double tap the "Watch in DuckPlayer" button sometimes.   This is [reported](https://app.asana.com/0/1204099484721401/1207658032936628/f) and should be fixed as part of the FE work

**Steps to test this PR**:

⚠️  0. Update privacy config in Debug to use: https://jsonblob.com/1253721903368888320 and restart the app

1.  Go to Settings > Duck Player, and Select "Ask every time"
2.  Open a New tab and go to Youtube.com
3. Tap on a video
4. Confirm the Overlay shows on top of the video
5. Tap Watch here, and confirm video starts playing
6. Refresh the page
7. Tap Watch in Duck Player and confirm video plays in DP
8. Hit the back button
9. Confirm you're taken to the video page (not the youtube page)


DuckDuckGo/DuckPlayer/DuckPlayer.swift
```diff
@@ -112,73 +91,77 @@ struct InitialSetupSettings: Codable {
...
+    public struct OriginDomains {
+        static let duckduckgo = "duckduckgo.com"
+        static let youtubeWWW = "www.youtube.com"
+        static let youtube = "youtube.com"
+        static let youtubeMobile = "m.youtube.com"
+    }
+    
+    var mode: DuckPlayerMode {
+        get {
+            AppDependencyProvider.shared.appSettings.duckPlayerMode
```
Bunn: Maybe we can inject a DependencyProvider with a default value of AppDependencyProvider.shared.appSettings? Just to make the injection a bit clear from the callsite and make testing easier if necessary

DuckDuckGo/DuckPlayer/DuckPlayer.swift
```diff
@@ -112,73 +91,77 @@ struct InitialSetupSettings: Codable {
...
-
-    @Published var mode: DuckPlayerMode
-
-    var overlayInteracted: Bool {
-        true
-    }
 
-    init() {
-        mode = .enabled
+    init(settings: DuckPlayerSettings = DuckPlayerSettings()) {
```
Bunn, line 136: 👍

Bunn (approved):
LGTM. I've seen some issues I'll describe bellow, but not sure if it's part of this PR to solve them so I'm approving and letting you decide.

**1. Bad text wrapping.**
I believe this is expected since we're using the macOS page. But worth mentioning
Screenshot 2024-06-26 at 11 29 23

**2. Issue reloading**
When reloading a video from duck player, instead of opening duck player again, it goes to the no-cookies URL

Simulator Screen Recording - iPhone 15 Pro - 2024-06-26 at 11 30 05

afterxleep:
@Bunn Yes, both issues reported are expected. First is macOS UI related, correct.   Second will be fixed by next PR (which takes care of the URL management).

DuckDuckGo/DuckPlayer/DuckPlayer.swift
afterxleep: Good call

PR #3000, opened 2024-06-26
scripts for running UI tests

brindy:
Task/Issue URL: https://app.asana.com/0/392891325557410/1207589761442405/f
Tech Design URL:
CC: @bwaresiak 

**Description**:
Adds scripts for running the UI tests locally in a way that's similar to how they run on CI.

**Setting up UI tests**:
1. Run `.maestro/setup_ui_tests.sh` from the `.maestro` folder. It should tell you to run it from the root. If there are other errors with your environment (such as missing commands) it will also fail.
2. Run `.maestro/setup_ui_tests.sh` from the root. It'll build the app to a local `DerivedData` folder, create a simulator and install the app
3. Run `.maestro/setup_ui_tests.sh --skip-build`.  It'll skip the build, but create a simulator and install the app if one is available in the right location

**Running UI tests**:
1. Run `.maestro/run_ui_tests.sh`. It'll fail because you need to specific a location for the flow files to run.
6. Run `.maestro/run_ui_tests.sh --flow-location .maestro\x` which will fail because the location is wrong.
7. Run `.maestro/run_ui_tests.sh --flow-location .maestro\release_tests\` it will run the tests and create a file called run_log.txt in `DerivedData`. Once it finishes it'll dump that file to the console.  If they all passed it will show a success message.
8. Run `.maestro/run_ui_tests.sh --run-flow` it'll fail because you need to specify a flow file
9. Run `.maestro/run_ui_tests.sh --run-flow .maestro/release_tests/bookmarks.yaml` it'll run just that test
10. Run `.maestro/run_ui_tests.sh --flow-location .maestro\release_tests\` and hit Ctrl-C.  This will make it appear like the test has failed.  Do this for a few test and allow others to pass.  The final message should indicate that some failures occurred.

**Cleaning up**:
1. Run `.maestro/clean_simulators.sh` and it'll kill the simulators and then remove the ones that have 'maestro' in the name. Some errors might get printed for where they don't need to be shutdown but this can be ignored.


.maestro/setup_ui_tests.sh
```diff
@@ -0,0 +1,112 @@
+### Set up environment for UI testing
```
loremattei, line 3: Should we add the shebang?

.maestro/setup_ui_tests.sh
```diff
@@ -0,0 +1,112 @@
...
+
+## Constants
+derived_data_path=DerivedData
+app_location=$derived_data_path/Build/Products/Debug-iphonesimulator/DuckDuckGo.app
+device_uuid_path="$derived_data_path/device_uuid.txt"
+
```
loremattei, line 8: Since these should stay in sync with run_ui_tests, should we put them in a shared file and source it here and in the other script?

.maestro/setup_ui_tests.sh
```diff
@@ -0,0 +1,112 @@
...
+## Functions
+
+fail() {
+    echo "‼️ $1"
+    exit 1
+}
+
+check_is_root() {
+    if [ ! -d ".maestro" ]; then
```
loremattei: I can make this fail, for example by invoking from a parent folder, but it works in the most common scenarios for sure. 
Checking it in a more robust way is much more verbose, so I'm inclined to say this is good enough. 
But what's the worst thing that can happen if this is invoked from other directories?

.maestro/setup_ui_tests.sh
```diff
@@ -0,0 +1,112 @@
...
+echo
+echo "ℹ️  Checking environment for UI testing with maestro"
+
+## If maestro is not installed run the following commands
+#
+# curl -Ls "https://get.maestro.mobile.dev" | bash
+# brew tap facebook/fb
+# brew install facebook/fb/idb-companion
+#
+check_command maestro
```
loremattei: I'd prefer this to print the instructions, but I guess it's good enough given that it's something that shouldn't happen more than once to people.

loremattei: I also wonder if we should check versions here, to avoid people getting mad because of an old/wrong/buddy version of maestro. 
We can add this later, though.

.maestro/setup_ui_tests.sh
```diff
@@ -0,0 +1,112 @@
...
+while [[ "$#" -gt 0 ]]; do
+    case $1 in
+        --skip-build) skip_build=1; shift ;;
+        *)
+    esac
+    shift
+done
+
+if [ ! -n "$skip_build" ]; then
+    echo "Skipping build"
```
loremattei, line 87: 🤔 Is this message flipped, or am I misunderstanding?

.maestro/setup_ui_tests.sh
```diff
@@ -0,0 +1,112 @@
...
+    build_app
+fi
+
+echo "ℹ️ Closing all simulators"
+killall Simulator
+
+device_uuid=$(xcrun simctl create "$target_device $target_os (maestro)" "com.apple.CoreSimulator.SimDeviceType.$target_device" "com.apple.CoreSimulator.SimRuntime.$target_os")
```
loremattei, line 98: Can we add a message to say that we are trying to restart the simulator? 
I was initially confused from the log I saw in my console after launching the script and not following the run because it says "Closing all simulators", but then there's work with simulators and the simulator itself is open.

.maestro/setup_ui_tests.sh
brindy: It is hah, thanks

.maestro/setup_ui_tests.sh
brindy: Hmm, not much tbh, it'll fail because it can't build the app and find some folders and stuff, but wanted to avoid people having to work that out.

.maestro/setup_ui_tests.sh
brindy: Yep! 😅

.maestro/setup_ui_tests.sh
brindy: I did this anyway :)

brindy:
@loremattei addressed your comments - ready for another look, thanks.

.maestro/setup_ui_tests.sh
```diff
@@ -0,0 +1,120 @@
...
+    if [ $? -ne 0 ]; then
+        echo "‼️ Unable to build app into $derived_data_path"
+        exit 1
+    fi
+
+}
+
+## Main Script
+
+check_is_root
```
loremattei: This can't be invoked unless the "source" operation above is successful, but it will be successful only if the script is invoked from the root folder (otherwise ".maestro/common.sh" will be invalid). 
So, if invoked from other folders, the script will print an error message that this command is unknown, but it will continue and try to run the next instructions even if it's on the wrong folder. 

An option could be to use dirname "$0" to get the path of this file and use it in the source command instead of the static path.

.maestro/run_ui_tests.sh
```diff
@@ -0,0 +1,137 @@
+#!/bin/zsh
+
+### Run UI tests
+
+source .maestro/common.sh
```
loremattei: This has the same issue described for the other file.

.maestro/run_ui_tests.sh
```diff
@@ -0,0 +1,137 @@
...
+app_bundle="com.duckduckgo.mobile.ios"
+
+## Functions
+
+fail() {
+    echo "‼️ $1"
+    exit 1
+}
+
+check_is_root() {
```
loremattei: This is technically duplicated now... even though, considering the issue in the source command, it's actually the definition that it's used when the script is invoked from the wrong folder.

.maestro/run_ui_tests.sh
```diff
@@ -0,0 +1,137 @@
...
+			flow_location="$2"
+			if [ ! -d "$flow_location" ]; then
+				fail "Invalid flow location $flow_location, use --flow-location .maestro/flow_folder "
+			fi
+			shift ;;
+        *) fail "Unknown parameter passed: $1" ;;
+    esac
+    shift
+done
+
```
loremattei, line 72: Very minor and not a blocker to merge: if the user doesn't pass any arg, the script will fail with a "no matches found: /*.yaml" error message which is not too difficult, neither straightforward to understand. 
Should we print a "help" page with usage if the user invokes the script without any parameter?

.maestro/run_ui_tests.sh
```diff
@@ -0,0 +1,137 @@
...
+rm $run_log
+
+log_message $run_log "START"
+
+if [ -n "$run_flow" ]; then
+	if [ ! -f $run_flow_file ]; then
+		fail "$run_flow_file is not a file"
+	fi
+
+	run_flow $device_uuid $run_flow_file
```
loremattei: This is weird, but this command always fail in my environment 🤔 with an error that the driver didn't start in time. 
But the same command actually consistently works if run from the "else" branch 🤷‍♂️

It looks like a timing issue where the simulator is not ready yet and the small delay makes maestro timeout. 
As a quick test, a "sleep 1" before "run_flow" solved the problem for me, but maybe there are better, more robust solutions.

brindy:
@loremattei 

Just pushed a few changes
* You can now run the scripts from anywhere 
* Removed the parameters from the run_ui_tests script, it now takes a single parameter
  * if it's a file it runs a single flow
  * if it's a folder it iterates over all the items in the folder.  It'll give some weird output if you choose a folder with no flows, but should be obvious enough why

.maestro/setup_ui_tests.sh
brindy: Basically made it so you can run these scripts from anywhere so this is fixed.

.maestro/run_ui_tests.sh
brindy: so now it takes 1 parameter which is the name of the flow or a folder of flows, should resolve this

.maestro/run_ui_tests.sh
brindy: Let me know if you still see this after my recent changes.

loremattei (approved):
@brindy this looks great!
Thanks for all the updates!
:shipit:

PR #3017, opened 2024-07-01
Integrate RemoteMessaging with NewTabPage

dus7:
Task/Issue URL: https://app.asana.com/0/1206226850447395/1207703703130204/f
Tech Design URL:
CC:

**Description**:

Adds RemoteMessaging capability to new `NewTabPage` layout. 

`PixelFiring` protocol got reorganized and extended, allowing to abstract this functionality and make depending logic testable. So far it was used only in `AdAttributionPixelReporter`.

**Steps to test this PR**:
1. Change `RemoteMessagingClient.endpoint` value to `https://www.jsonblob.com/api/1257306235471781888`.
2. Restart app.
3. Make sure Internal User flag is set and NewTabPage sections are enabled in debug menu.
4. Test functionality of all 5 messages.
5. Check if proper pixels are fired on dismiss, appear, close and actions.
6. Verify if message does not appear twice after it's dismissed.

**Device Testing**:

* [ ] iPhone 14 Pro
* [ ] iPad

---
###### Internal references:

### DIFF
--- a/Core/PixelFiring.swift
+++ b/Core/PixelFiring.swift
@@ -1,0 +1,50 @@
+//
+//  PixelFiring.swift
+//  DuckDuckGo
+//
+//  Copyright © 2024 DuckDuckGo. All rights reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//  http://www.apache.org/licenses/LICENSE-2.0
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+import Networking
+
+public protocol PixelFiring {
+
+    static func fire(_ pixel: Pixel.Event,
+                     withAdditionalParameters params: [String: String],
+                     includedParameters: [Pixel.QueryParameters],
+                     onComplete: @escaping (Error?) -> Void)
+
+    static func fire(_ pixel: Pixel.Event,
+                     withAdditionalParameters params: [String: String])
+}
+
+extension Pixel: PixelFiring {
+    public static func fire(_ pixel: Pixel.Event,
+                            withAdditionalParameters params: [String: String],
+                            includedParameters: [Pixel.QueryParameters],
+                            onComplete: @escaping (Error?) -> Void) {
+        
+        Pixel.fire(pixel: pixel,
+                   withAdditionalParameters: params,
+                   includedParameters: includedParameters,
+                   onComplete: onComplete)
+    }
+
+    public static func fire(_ pixel: Pixel.Event,
+                            withAdditionalParameters params: [String: String]) {
+        Pixel.fire(pixel: pixel, withAdditionalParameters: params)
+    }
+}
--- a/Core/PixelFiringAsync.swift
+++ b/Core/PixelFiringAsync.swift
@@ -1,0 +1,43 @@
+//
+//  PixelFiringAsync.swift
+//  DuckDuckGo
+//
+//  Copyright © 2024 DuckDuckGo. All rights reserved.
+//
+//  Licensed under the Apache License, Version 2.0 (the "License");
+//  you may not use this file except in compliance with the License.
+//  You may obtain a copy of the License at
+//
+//  http://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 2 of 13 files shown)


DuckDuckGo/NewTabPageMessagesModel.swift
```diff
@@ -0,0 +1,121 @@
...
+    private func refresh() {
+        homePageMessagesConfiguration.refresh()
+        updateHomeMessageViewModel()
+    }
+
+    private func updateHomeMessageViewModel() {
+        self.homeMessageViewModels = homePageMessagesConfiguration.homeMessages.compactMap(self.homeMessageViewModel(for:))
+    }
+
+    private func homeMessageViewModel(for message: HomeMessage) -> HomeMessageViewModel? {
```
dus7, line 73: This function is copied from `HomeMessageViewSectionRenderer`.

DuckDuckGo/NewTabPageView.swift
```diff
@@ -19,13 +19,32 @@
 
 import SwiftUI
 import DuckUI
+import RemoteMessaging
```
dus7, line 22: Required for Preview

Core/PixelFiringAsync.swift
dus7, line 1: Moved from `AdAttributionPixelReporter`.

Core/PixelFiring.swift
```diff
@@ -0,0 +1,50 @@
...
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+import Networking
+
+public protocol PixelFiring {
+
+    static func fire(_ pixel: Pixel.Event,
```
dus7, line 25: Had to make first param unnamed to avoid name ambiguity, but I think API still makes sense, only downside is that when using `Pixel` directly there are both variants available in autocomplete. In the end I believe we should be using protocol exclusively.

DuckDuckGo/NewTabPageView.swift
```diff
@@ -51,9 +72,54 @@ struct NewTabPageView: View {
 
     private struct Constant {
         static let sectionPadding = EdgeInsets(top: 16, leading: 24, bottom: 16, trailing: 24)
+
+        static let messageMaximumWidth: CGFloat = 380
+        static let messageMaximumWidthPad: CGFloat = 455
```
dus7, line 77: Used values from `HomeMessageCollectionViewCell`.

brindy (approved):
The pixel thing confused me for a moment. Took me a minute to understand why we need the async one, but I get it. If you think this a pattern we should propagate then please create a diary entry for it.  Not sure how compatible it is with the PixelKit work?

Anyway, aside from that, LGTM!

dus7:
>  Not sure how compatible it is with the PixelKit work?

@federicocappelli any thoughts on this? Do you think it clashes with PixelKit in any way?

federicocappelli:
> > Not sure how compatible it is with the PixelKit work?
> 
> @federicocappelli any thoughts on this? Do you think it clashes with PixelKit in any way?

Hi @dus7, from my understanding you wanted to:
1. Add swift concurrency to the Pixel.fire(...) to fire a pixel synchronously instead of the usual fire-and-forget
2. Be able to mock pixel firing for testing purposes
Correct?

Your implementation seems good BUT I'm a bit worried about the constant diverging of our approaches across platforms, this could have been a good occasion for adopting PixelKit (only for `.appleAdAttribution`), adding swift concurrency to it and using the tools in `PixelKitTestingUtilities` for your unit tests.

My point is that I would like to see more collaboration and various teams working on shared tools instead of each of us re-implementing its own thing. This code becomes immediately tech debt, and another blocker, as soon someone tries to adopt PixelKit in iOS.

dus7:
Thanks @federicocappelli, this makes sense to me. Completely agreed we should not be diverging our approaches and it wasn't my goal at all while writing this. I wasn't aware about the existence of `PixelKitTestingUtilities`, nor there's any usage of `PixelKit` yet in iOS app so I just went and solved my own problem which has now escalated to multiple places :).

I'll merge this PR and add a [task](https://app.asana.com/0/1206226850447395/1207725427404718/f) for myself to add swift concurrency to PixelKit and refactor this code to use PixelKit instead.

federicocappelli:
> Thanks @federicocappelli, this makes sense to me. Completely agreed we should not be diverging our approaches and it wasn't my goal at all while writing this. I wasn't aware about the existence of `PixelKitTestingUtilities`, nor there's any usage of `PixelKit` yet in iOS app so I just went and solved my own problem which has now escalated to multiple places :).
> 
> I'll merge this PR and add a task for myself to add swift concurrency to PixelKit and refactor this code to use PixelKit instead.

Perfectly understandable! Thank you for putting in the effort!

PR #3019, opened 2024-07-01
improve bookmarks and favorites UI tests

brindy:
Task/Issue URL: https://app.asana.com/0/392891325557410/1207195241923822/f
Tech Design URL:
CC: @bwaresiak 

**Description**:
Fleshed out and tweaked bookmarks and favorites UI tests

Notes:
* The scripts are being reviewed in https://github.com/duckduckgo/iOS/pull/3000 so you can largely ignore them.
* You should be able to run the scripts from anywhere, just adjust the paths below accordingly
* Fixes a bug found during testing in the bookmarks swift code

**Steps to test this PR**:
* Confirm End to End tests have passed: https://github.com/duckduckgo/iOS/actions/runs/9750711656
* In terminal cd into .maestro
* Run `setup_ui_tests.sh`
* Optional: run all the release tests in one go using: run_tests_tests.sh release_tests
* Run `run_ui_tests.sh release_tests/bookmarks.yaml`
* Run `run_ui_tests.sh release_tests/bookmarks-multi.yaml`
* Run `run_ui_tests.sh release_tests/favorites.yaml`


DuckDuckGo/Base.lproj/Bookmarks.storyboard
```diff
@@ -432,7 +432,7 @@
         <scene sceneID="gmt-6C-Bi5">
             <objects>
                 <tableViewController storyboardIdentifier="BookmarkFoldersViewController" id="o1q-bZ-7v3" userLabel="Bookmark Folders" customClass="BookmarkFoldersViewController" customModule="DuckDuckGo" customModuleProvider="target" sceneMemberID="viewController">
-                    <tableView key="view" clipsSubviews="YES" contentMode="scaleToFill" alwaysBounceVertical="YES" dataMode="prototypes" style="insetGrouped" allowsSelectionDuringEditing="YES" rowHeight="-1" estimatedRowHeight="-1" sectionHeaderHeight="18" sectionFooterHeight="18" id="V1z-bn-1Gk">
+                    <tableView key="view" clipsSubviews="YES" contentMode="scaleToFill" alwaysBounceVertical="YES" keyboardDismissMode="onDrag" dataMode="prototypes" style="insetGrouped" allowsSelectionDuringEditing="YES" rowHeight="-1" estimatedRowHeight="-1" sectionHeaderHeight="18" sectionFooterHeight="18" id="V1z-bn-1Gk">
```
loremattei, line 435: Is this intentional?

DuckDuckGo/BookmarksViewController.swift
```diff
@@ -444,6 +438,8 @@ class BookmarksViewController: UIViewController, UITableViewDelegate {
                 completion(true)
             }
             present(alertController, animated: true)
+        } else if bookmark.isFolder {
+            delete()
```
loremattei, line 442: This is to avoid the debug assertion, right?
Why aren't we calling the completion handler here?

loremattei (approved):
@brindy LGTM!
I left a couple small questions just to double check, but feel free to merge if things look good to you.

DuckDuckGo/Base.lproj/Bookmarks.storyboard
brindy: Yep - without it, on some screens when you do

- assertVisible: "Delete"
- tapOn: "Delete" 

Then maestro thinks that has worked when it hasn't and scrolling on that page doesn't bring the delete button up fair enough.  The usual practice is to dismiss the keyboard on swiping a list anyway so I think it's OK.

DuckDuckGo/BookmarksViewController.swift
brindy: Good catch, I'll check that.

brindy: There's a few paths that don't but I think this one should so I've added it.  I had a quick look and can't really see what the implication of not calling it is.  It seems that if you don't then the buttons shown on swipe don't get hidden, but since we're removing the row anyway it doesn't matter *too* much.

PR #3030, opened 2024-07-03
Fix issue when cancelling a download

Bunn:
Task/Issue URL: https://app.asana.com/0/414235014887631/1207722065639812/f

**Description**:
Fix issue where the wrong item was being cancelled in the download list

**Steps to test this PR**:
1. Open https://www.thinkbroadband.com/download
2. Select multiple items to download
3. Cancel the downloads and check if the correct ones in the list are being removed
4. To make it easier, you can also change the line [here](https://github.com/duckduckgo/iOS/blob/b33782933a33c959d01e9cdfc096ae93aeae014c/DuckDuckGo/DownloadsList.swift#L206) to something like `Text("\(UserText.cancelDownloadAlertDescription) \(row.id)")` so you can see what item is being selected

### DIFF
--- a/DuckDuckGo/DownloadsList.swift
+++ b/DuckDuckGo/DownloadsList.swift
@@ -24,9 +24,8 @@ struct DownloadsList: View {
     @Environment(\.presentationMode) var presentationMode
     @ObservedObject var viewModel: DownloadsListViewModel
     @State var editMode: EditMode = .inactive
-    
-    @State private var isCancelDownloadAlertPresented: Bool = false
-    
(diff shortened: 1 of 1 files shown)


federicocappelli:
LGTM, tested cancelling 6 different downloads.

Unrelated note, the alert action when I try to remove the download seems **VERY** confusing to me!
IMG_7664

They should be:
- Are you sure you want to cancel this download?
- Yes/Confirm
- No

the `Resume` option makes no sense to me

federicocappelli (approved):
LGTM

PR #3053, opened 2024-07-07
Hide all connection details when the VPN isn’t connected

samsymons:
Task/Issue URL: https://app.asana.com/0/1199333091098016/1207750283125995/f
Tech Design URL:
CC:

**Description**:

This PR fixes an issue with the VPN showing connection details for the last connection, even after disconnecting.

See https://github.com/duckduckgo/BrowserServicesKit/pull/858/files for where the issue was introduced - rather than modify that code, I'm adding checks on the iOS VPN UI side to ensure we only show the status row when appropriate.

**Steps to test this PR**:
1. Open the VPN UI, make sure that no connection details are showing
2. Disconnect the VPN, make sure that IP address and throughout disappear
3. Reconnect the VPN, make sure that connection details reappear and look accurate
4.

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPhone 14 Pro
* [ ] iPad

**OS Testing**:

* [ ] iOS 15
* [ ] iOS 16
* [ ] iOS 17

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:

### DIFF
--- a/DuckDuckGo/NetworkProtectionStatusView.swift
+++ b/DuckDuckGo/NetworkProtectionStatusView.swift
@@ -40,7 +40,7 @@ struct NetworkProtectionStatusView: View {
             toggle()
             locationDetails()
 
-            if statusModel.shouldShowConnectionDetails {
+            if statusModel.isNetPEnabled && statusModel.shouldShowConnectionDetails && statusModel.ipAddress != nil {
                 connectionDetails()
             }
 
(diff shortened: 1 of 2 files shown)


DuckDuckGo/NetworkProtectionStatusView.swift
```diff
@@ -40,7 +40,7 @@ struct NetworkProtectionStatusView: View {
             toggle()
             locationDetails()
 
-            if statusModel.shouldShowConnectionDetails {
+            if statusModel.isNetPEnabled && statusModel.shouldShowConnectionDetails && statusModel.ipAddress != nil {
                 connectionDetails()
             }
```
samsymons, line 45: This should be changed before merging, the correct fix is to make sure that `shouldShowConnectionDetails` is accurate. I'm opening the PR now anyway since I need these changes for another branch, but will aim to update this by EOD.

diegoreymendez: I posted this in MM but cross-posting here so it's easy to track the conversation.  I think the GUI logic is just wrong here, and I wouldn't change BSK.

Showing or hiding UI shouldn't rely directly on notifications, but on local state.  In this case it should rely on connection status.  This is the way macOS does it.

We don't ever want to show details when disconnected, and we always want to show details when connected.

diegoreymendez (approved):
Approved, great catch - dunno how I missed this.

Still I think the issue is really in the UI code: see my comment.

diegoreymendez:
I'm going ahead to merge so this goes out in the internal.

PR #3054, opened 2024-07-08
Onboarding intro ship review design feedback

alessandroboron:
Task/Issue URL: https://app.asana.com/0/1206329551987282/1207752279048482/f

**Description**:

This PR:
1. Update the Shadows for Dax Dialogue according to Designer feedback.
2. Implement animation when the landing screen transitions to the first Dax intro dialog

**Steps to test this PR**:

Shadows are being checked by designer doing the design review.

**SCENARIO 1 - Dax Animation**
1. Launch the Onboarding Intro flow 
2. Observe the transition between “Welcome screen” and the first Dax Dialog
**Expected Result**: The Dax Icon should translate in position, then the Dax dialogue should appear, and the text should start animating.

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPhone 14 Pro
* [ ] iPad

**OS Testing**:

* [ ] iOS 15
* [ ] iOS 16
* [ ] iOS 17

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:


DuckDuckGo/OnboardingExperiment/OnboardingIntro/OnboardingView.swift
```diff
@@ -82,17 +88,25 @@ struct OnboardingView: View {
             }
             .frame(width: geometry.size.width, alignment: .center)
             .offset(y: geometry.size.height * Metrics.dialogVerticalOffsetPercentage.build(v: verticalSizeClass, h: horizontalSizeClass))
+            .onAppear {
+                DispatchQueue.main.asyncAfter(deadline: .now() + Metrics.daxDialogVisibilityDelay) {
```
alessandroboron, line 92: It looks like if I use `with animation (.linear.delay(0.5)) {}`, the content of the closure is executed immediately; hence, the text appears as if it is already typed.

brindy: The way to do this would be to change `AnimatableTypingText` view to support the` Animatable` protocol which tbh I should have caught early on when you started this, sorry.

Would need a bit of a refactor because passing in true to that class starts a timer, but you don't need any of that if you use Animatable - just need to know what value you are interpolating to, so rather than starting a timer you have your start index at zero and target index at the length of the string and let SwiftUI interpolate between them for you using the `animatableData` property required by the protocol (an empty default is provided).

In places you don't use `withAnimation` then the text will just appear in full by default, so kind of the opposite of what's happening here.

It's another one of those 'mindset' changes :) 

It's also fairly advanced topic so we can skip it for now, I think.  It'd be nice to do it "right" at some point though.

brindy (approved):
LGTM! I left a note about using `Animatable` but let's circle back to that in future (maybe).

DuckDuckGo/OnboardingExperiment/OnboardingIntro/OnboardingView.swift
alessandroboron: Hey @brindy no worries and thanks for pointing this out. Would like to give it a go when I have some down time!

PR #3068, opened 2024-07-11
ensure only one history database context

brindy:
Task/Issue URL: https://app.asana.com/0/392891325557410/1207755453158856/f
Tech Design URL:
CC:

**Description**:
Ensures there's only 1 database context for the history store to avoid having merge conflicts.

**Steps to test this PR**:

***Fire button:***
1. Make sure feature is enabled
2. Visit some sites and generate history
3. Check history appears
4. Use the fire button
5. Check history has been deleted

***Settings:***
2. Visit some sites and generate history.
3. Disable the autocomplete setting
4. Check history not visible and visit different sites
6. Enable the autocomplete setting
7. Check previous history not visible
8. Repeat but by turning off the recently visited sites setting

***Autoclear:***
1. Turn on autoclear and set timer to 5 seconds in the code 
2. Visit some sites
3. Background the app and trigger autoclear 
4. Check sites not in history
5. Visit more sites
6. Kill the app
7. Check sites not in history

### DIFF
--- a/Core/HistoryManager.swift
+++ b/Core/HistoryManager.swift
@@ -25,9 +25,11 @@ import Common
 import Persistence
 
 public protocol HistoryManaging {
-
+    
     var historyCoordinator: HistoryCoordinating { get }
-    func loadStore(onCleanFinished: @escaping () -> Void) throws
+    func isHistoryFeatureEnabled() -> Bool
+    var isEnabledByUser: Bool { get }
+    func removeAllHistory() async
 
 }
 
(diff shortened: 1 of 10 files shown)


Core/HistoryManager.swift
```diff
@@ -25,9 +25,11 @@ import Common
 import Persistence
 
 public protocol HistoryManaging {
-
+    
     var historyCoordinator: HistoryCoordinating { get }
-    func loadStore(onCleanFinished: @escaping () -> Void) throws
+    func isHistoryFeatureEnabled() -> Bool
+    var isEnabledByUser: Bool { get }
+    func removeAllHistory() async
```
brindy, line 32: protocol was not in use widely so added these functions which are used and made all dependents use the protocol

Core/HistoryManager.swift
```diff
@@ -106,25 +103,12 @@ public class HistoryManager: HistoryManaging {
 
     public func removeAllHistory() async {
         await withCheckedContinuation { continuation in
-            historyCoordinator.burnAll {
+            dbCoordinator.burnAll {
```
brindy, line 106: Always use the db coordinator when burning, even if disabled in settings.

Core/HistoryManager.swift
```diff
@@ -244,3 +228,59 @@ class HistoryStoreEventMapper: EventMapping<HistoryStore.HistoryStoreEvents> {
         fatalError("Use init()")
     }
 }
+
+extension HistoryManager {
+
+    /// Should only be called once in the app
+    public static func make(isAutocompleteEnabledByUser: @autoclosure @escaping () -> Bool,
```
brindy, line 235: Moved creation logic here.  These auto closures mean I don't need to pass in AppSettings into a Core class.

DuckDuckGo/AppDelegate.swift
```diff
@@ -346,6 +344,29 @@ import WebKit
...
+                                   privacyConfigManager: ContentBlocking.shared.privacyConfigurationManager) {
+
+        case .failure(let error):
+            Pixel.fire(pixel: .historyStoreLoadFailed, error: error)
+            if error.isDiskFull {
+                self.presentInsufficientDiskSpaceAlert()
+            } else {
+                self.presentPreemptiveCrashAlert()
+            }
+            return NullHistoryManager()
```
brindy, line 363: Will never actually get used because the app will get killed, but gotta return something for the compiler.

bwaresiak (approved):
There is some linter error, but its minor. 

I also think we could limit the work we do on startup (or move it off main thread in some cases) but I've perf-tested this with 5k visits on the device and it still performed well, so I'm not that concerned. 

Great job :)

PR #3081, opened 2024-07-15
fix ui tests broken by new onboarding and use shared setup flow

brindy:
Task/Issue URL: https://app.asana.com/0/392891325557410/1207808692245660/f
Tech Design URL:
CC:

**Description**:

Also: https://app.asana.com/0/392891325557410/1207809241818381/f

Fixes UI test breakage by skipping assertion for title of onboarding screen.

**Steps to test this PR**:
1. Confirm E2E tests pass: https://github.com/duckduckgo/iOS/actions/runs/9974333207
3. Confirm Sync tests pass: https://github.com/duckduckgo/iOS/actions/runs/9974343415
5. Run the `.maestro/setup_ui_tests.sh` script.
6. Run some of the UI tests using the `.maestro/run_ui_tests.sh` script.  Should consistently pass.

### DIFF
--- a/Core/DefaultVariantManager.swift
+++ b/Core/DefaultVariantManager.swift
@@ -91,7 +91,7 @@ public protocol VariantRNG {
 public class DefaultVariantManager: VariantManager {
 
     public var currentVariant: Variant? {
-        let variantName = ProcessInfo.variantName
+        let variantName = ProcessInfo.variantName ?? storage.variant
         return variants.first(where: { $0.name == variantName })
     }
 
--- a/Core/DefaultVariantManager.swift
+++ b/Core/DefaultVariantManager.swift
@@ -182,13 +182,12 @@ public class Arc4RandomUniformVariantRNG: VariantRNG {
 extension ProcessInfo {
 
     static var variantName: String? {
-        guard let variantArg = ProcessInfo().arguments.first(where: {
-            $0.starts(with: "VARIANT=")
+        let p = ProcessInfo()
+        guard let variantArgIndex = p.arguments.firstIndex(where: {
+            $0.starts(with: "VARIANT")
         }) else { return nil }
-
-        let components = variantArg.components(separatedBy: "=")
-        guard components.count == 2 else { return nil }
-        return components[1]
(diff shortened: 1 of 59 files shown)


Core/DefaultVariantManager.swift
```diff
@@ -91,7 +91,7 @@ public protocol VariantRNG {
 public class DefaultVariantManager: VariantManager {
 
     public var currentVariant: Variant? {
-        let variantName = ProcessInfo.processInfo.environment["VARIANT", default: storage.variant ?? "" ]
```
alessandroboron: Out of curiosity, how come we changed from environment variable to launch argument?
Maestro lets set you the env variable as per [Maestro doc](https://maestro.mobile.dev/api-reference/configuration/flow-configuration)

alessandroboron:
Hey Brindy:

When I tested manually I got the expected behaviour. When I run the test in some instance I got into this scenario. The launch argument variant is “ru” but the onboarding shown is the new one.

Screenshot 2024-07-16 at 1 07 52 PM

Core/DefaultVariantManager.swift
brindy: Ah, I just didn't know.  This is even better.  Will revert to the env version

brindy: OK, had a look at this.

1/ Env doesn't appear to actually work.  I added some extra UI in the debug screen to check it (haven't committed that).

2/ We'd have to put it in every file since env is specific to the flow and I'm trying to have a standard 'set up' flow.

Interestingly the VARIANT argument gets changed to `-VARIANT` for some reason. Tested this with autoclear and that argument doesn't, but maybe because it's a bool?

Anyway, specifying an override prevents dax dialogs from showing and I think we want that because we want these tests to run against a something as close to production as possible.  So I am going to change the tests to just get past the onboarding and do everything else as expected.

Once we look to merge the new dax dialogs we'll have to update the tests again.

alessandroboron: Ah bummer, I saw that it was available so I suggested it. Thanks for the explanation, makes sense.

alessandroboron (approved):
Works as expected!

PR #3086, opened 2024-07-16
Expand AdHoc build workflow, add debug bookmarks screen

bwaresiak:
Task/Issue URL: https://app.asana.com/0/414235014887631/1207821981275160/f

https://app.asana.com/0/856498667320406/1207823615637251/f

**Description**:

Expand AdHoc build workflow to have an option of alpha and regular builds. Add debug bookmarks screen.

**Steps to test this PR**:

Check the workflow file for any obvious mistakes, validate history of executions for any mistakes.

Validate debug screen works as expected.

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPhone 14 Pro
* [ ] iPad

**OS Testing**:

* [ ] iOS 15
* [ ] iOS 16
* [ ] iOS 17

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:

### DIFF
--- a/.github/workflows/adhoc.yml
+++ b/.github/workflows/adhoc.yml
@@ -1,6 +1,7 @@
 name: Make ad-hoc build
 
 on:
+  push:
   workflow_dispatch:
     inputs:
       suffix:
@@ -49,7 +50,7 @@ jobs:
           APPLE_API_KEY_ISSUER: ${{ secrets.APPLE_API_KEY_ISSUER }}
           MATCH_PASSWORD: ${{ secrets.MATCH_PASSWORD }}
         run: |
-          lane_to_use="adhoc"
+          lane_to_use="release_adhoc"
           if [[ "${{ github.event.inputs.is-alpha }}" == "Release" ]]; then
             lane_to_use = "release_adhoc"
           fi
--- a/.github/workflows/adhoc.yml
+++ b/.github/workflows/adhoc.yml
@@ -10,7 +10,7 @@ on:
         description: "Asana task URL"
         required: false
         type: string
-      is-alpha:
+      build-type:
         description: "Build Configuration"
         type: choice
         required: true
@@ -50,7 +50,7 @@ jobs:
           MATCH_PASSWORD: ${{ secrets.MATCH_PASSWORD }}
         run: |
           lane_to_use="adhoc"
-          if [[ "${{ github.event.inputs.is-alpha }}" == "Release" ]]; then
+          if [[ "${{ github.event.inputs.build-type }}" == "Release" ]]; then
             lane_to_use = "release_adhoc"
           fi
           
--- a/.github/workflows/adhoc.yml
+++ b/.github/workflows/adhoc.yml
@@ -11,13 +11,13 @@ on:
         required: false
         type: string
       is-alpha:
-        description: "Is alpha build"
+        description: "Build Configuration"
         type: choice
         required: true
-        default: 'true'
+        default: 'Alpha'
         options:
-           - true
-           - false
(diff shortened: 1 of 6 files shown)


.github/workflows/adhoc.yml
```diff
@@ -10,6 +10,14 @@ on:
...
         required: false
         type: string
+      is-alpha:
+        description: "Is alpha build"
+        type: choice
+        required: true
+        default: 'true'
+        options:
+           - true
+           - false
```
ayoy: GHA supports boolean type so we can use it here. It would then use a checkbox in the UI, instead of a dropdown. The call site doesn't need updating:
```suggestion
      is-alpha:
        description: "Whether the build uses Alpha or Release configuration"
        type: boolean
        required: true
        default: true
```

bwaresiak: I know, but I've seen people complaining that sometimes these behave like booleans, sometimes like strings, there's a whole thread https://github.com/actions/runner/issues/1483 so I'd prefer to explicitly use strings.

ayoy: Yes, that's why the call site still needs to treat them as strings :) just pointing it out because it's possible to use a checkbox here, although on a second thought we could make it a dropbox with more descriptive options, like "Configuration: Alpha/Release". I'm not strong on any of my suggestions though, as long as the message is clear.

samsymons (commented):
@bwaresiak When testing the debug menu I saved a couple favorites, but the debug menu doesn't show their metadata - it does show the root folders correctly.

Bookmarks

Then when deleting all bookmarks and re-running the app, I ran into this assertion failure:

Assertion Failure

A second run of the app worked, and I got the root folders recreated, but I still wasn't getting bookmark metadata in the debug menu.

DuckDuckGo/AppDelegate.swift
```diff
@@ -117,6 +117,8 @@ import WebKit
     // swiftlint:disable:next function_body_length
     func application(_ application: UIApplication, didFinishLaunchingWithOptions launchOptions: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
 
+        OSLog.enabledLoggingCategories = ["Content Blocking"]
+
```
samsymons: Is this change intentional for this PR?

bwaresiak: No, I've missed this, thanks

.github/workflows/adhoc.yml
bwaresiak: I've tweaked the options and description a bit.

bwaresiak:
@samsymons thanks, I've fixed the context (as it was being deallocated, making entries lose the content) and removed the delete option - since bookmarks can be managed by sync we don't want to just wipe them from db.

.github/workflows/adhoc.yml
ayoy: Thanks, that looks much clearer. Sorry to be a pain, but I'd also rename `is-alpha` now that it's not a boolean anymore. How about `build-type`? `configuration` is the most correct, but it also is a broad term in general and may be confusing.

bwaresiak: Good suggestion, done :)

samsymons (approved):
LGTM!

ayoy (approved):
LGTM :)

bwaresiak:
Thank you!

PR #3092, opened 2024-07-17
Attach params to PPro pixels

quanganhdo:
Task/Issue URL: https://app.asana.com/0/414235014887631/1207759380490241/f
Tech Design URL:
CC:

**Description**:

Attach a randomized subset of parameters to certain PPro pixels.

**Steps to test**:

1. Use https://www.jsonblob.com/api/1263251505917321216 for remote config
2. Use https://www.jsonblob.com/api/1263231573443862528 for remote message
3. Check if the attached params are fired, and if their values are calculated correctly
4. I also have a Debug menu to show the param values

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPhone 14 Pro
* [ ] iPad

**OS Testing**:

* [ ] iOS 15
* [ ] iOS 16
* [ ] iOS 17

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:


DuckDuckGo/Subscription/PrivacyProDataReporting.swift
```diff
@@ -0,0 +1,310 @@
...
+    private let variantManager: VariantManager
+    private let userDefaults: UserDefaults
+    private let emailManager: EmailManager
+    private let tutorialSettings: TutorialSettings
+    private let appSettings: AppSettings
+    private let statisticsStore: StatisticsStore
+    private let secureVault: (any AutofillSecureVault)?
+    private var syncService: DDGSyncing?
+    private var tabsModel: TabsModel?
+    private let dateGenerator: () -> Date
```
quanganhdo: `syncService` and `tabsModel` are the trickiest to initialize, so I chose to inject them at runtime.

DuckDuckGo/Subscription/PrivacyProDataReporting.swift
```diff
@@ -0,0 +1,310 @@
...
+
+    func saveFireproofingUsed() {
+        guard !isFireproofingUsed() else { return }
+        userDefaults.set(true, forKey: Key.isFireproofingUsedKey)
+    }
+
+    func saveWidgetAdded() async {
+        let isInstalled = await appSettings.isWidgetInstalled()
+        if isInstalled != isWidgetAdded() {
+            userDefaults.set(isInstalled, forKey: Key.isWidgetAddedKey)
```
quanganhdo, line 256: This is cached so that `isWidgetAdded()` doesn't need to be `async` and that reduces the complexity when call sites are updated.

DuckDuckGo/Subscription/PrivacyProDataReporting.swift
```diff
@@ -0,0 +1,310 @@
...
+        case .origin(let origin):
+            guard let includedOrigins, let origin, includedOrigins.contains(origin) else { return [:] }
+        case .debug:
+            break
+        }
+
+        /// Wait for all the injected dependencies to be available
+        if !isReady && !ProcessInfo().arguments.contains("testing") {
+            return randomizedParameters(for: useCase)
+        }
```
quanganhdo: We run into this easily when the app launches and a remote message with the matching ID is shown.

amddg44: @quanganhdo testing using the provided configs worked as expected, however when I was regression testing with updated config id + messages, I kept ending up in an endless loop here. I was re-launching the app with an updated message config while on the new tab page with a remote message already showing (it looks like syncService is the one never being initialised)

image

amddg44 (approved):
LGTM, very nice work! 👏

DuckDuckGo/Subscription/PrivacyProDataReporter.swift
```diff
@@ -0,0 +1,289 @@
...
+    private let emailManager: EmailManager
+    private let tutorialSettings: TutorialSettings
+    private let appSettings: AppSettings
+    private let statisticsStore: StatisticsStore
+    private let secureVault: (any AutofillSecureVault)?
+    private var syncService: DDGSyncing?
+    private var tabsModel: TabsModel?
+    private let dateGenerator: () -> Date
+
+    init(config: PrivacyConfiguration = ContentBlocking.shared.privacyConfigurationManager.privacyConfig,
```
samsymons: The `PrivacyConfig` object is a struct, which means that we'll save a snapshot of it at initialization. If the Privacy Config is later updated, this object will not see those changes, so we should only refer to the manager and get the config dynamically each time.

DuckDuckGo/Subscription/PrivacyProDataReporter.swift
```diff
@@ -0,0 +1,289 @@
...
+        static let includedOrigins = "origins"
+    }
+
+    enum UseCase {
+        case messageID(String)
+        case origin(String?)
+        case debug
+    }
+
+    public static let shared = PrivacyProDataReporter()
```
samsymons: Let's replace the singleton with an instance of this class that's stored on the app delegate if possible - we're aiming to introduce no new singletons given their harm on testability. There are places in this PR that the singleton usage can impact unit testing, so I think it's worth resolving.

samsymons (approved):
New changes look good, thanks a lot!

PR #3097, opened 2024-07-17
Reduce VPN manager instances

samsymons:
Task/Issue URL: https://app.asana.com/0/72649045549333/1207151621945908/f
Tech Design URL:
CC:

**Description**:

This PR reduces the number of VPN manager instances that we create through the lifetime of the app.

**Steps to test this PR**:
1. Test that you can install and enable the VPN as expected, and that all features work
2. Test that removing the VPN configuration from Settings.app then returning to the VPN UI shows the VPN as disconnected, and when you try to enable it then you get prompted to install it
3. With the debugger connected, enable the VPN and stay on the status screen, then check that you don't see logs about the connection status once per second

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPhone 14 Pro
* [ ] iPad

**OS Testing**:

* [ ] iOS 15
* [ ] iOS 16
* [ ] iOS 17

---
###### Internal references:


DuckDuckGo/NetworkProtectionConvenienceInitialisers.swift
```diff
@@ -28,7 +28,7 @@ import Subscription
 
 private class DefaultTunnelSessionProvider: TunnelSessionProvider {
     func activeSession() async -> NETunnelProviderSession? {
-        try? await ConnectionSessionUtilities.activeSession()
+        return await AppDependencyProvider.shared.networkProtectionTunnelController.activeSession()
     }
 }
```
samsymons, line 33: We could remove `DefaultTunnelSessionProvider` entirely now, since it just delegates through to the tunnel controller. I'll push a commit to solve this.

DuckDuckGo/NetworkProtectionStatusViewModel.swift
```diff
@@ -346,7 +346,7 @@ final class NetworkProtectionStatusViewModel: ObservableObject {
     }
 
     private func refreshDataVolumeTotals() async {
-        guard let activeSession = try? await ConnectionSessionUtilities.activeSession() else {
+        guard let activeSession = await tunnelController.activeSession() else {
```
samsymons, line 352: This is the core change of this PR, we now reuse the active session from the tunnel controller. Because we use the tunnel controller from the app dependency provider, it means we have one session reused throughout the entire app.

DuckDuckGo/NetworkProtectionTunnelController.swift
```diff
@@ -184,10 +223,8 @@ final class NetworkProtectionTunnelController: TunnelController {
         }
     }
 
-    /// Reloads the tunnel manager from preferences.
-    ///
-    private func reloadTunnelManager() {
-        internalTunnelManager = nil
+    private func clearInternalManager() {
+        internalManager = nil
     }
```
samsymons, line 228: Renamed this function since `reload` was misleading - `clearInternalManager` is the function name on macOS so it's consistent now.

DuckDuckGo/NetworkProtectionTunnelController.swift
```diff
@@ -311,6 +317,31 @@ final class NetworkProtectionTunnelController: TunnelController {
...
+                    guard let manager = await self.internalManager else {
+                        return
+                    }
+
+                    do {
+                        try await manager.loadFromPreferences()
+
+                        if manager.connection.status == .invalid {
+                            self.clearInternalManager()
+                        }
```
samsymons, line 336: This is the key to making sure that the shared manager is always valid. This publisher will clear it out, and the next time it's accessed then a new one will be created (assuming we have a configuration at all).

diegoreymendez: Looking good - exactly as in macOS so I'm hoping that means it won't cause trouble.

diegoreymendez (commented):
The code changes look good, but I'm getting a crash on launch.

The issue is that `AppDependencyProvider` tries to instance `ConnectionStatusObserverThroughSession` which tries to access `AppDependencyProvider.shared`.

I have to say I don't love this `AppDependencyProvider` approach we're taking as it makes circular dependencies between singletons really hard to spot and likely to fail at runtime, like in this case.

I'm attaching the stack trace.

[Thread 1 Queue - com.apple.m.zip](https://github.com/user-attachments/files/16281944/Thread.1.Queue.-.com.apple.m.zip)

DuckDuckGoTests/MockDependencyProvider.swift
```diff
@@ -46,6 +46,7 @@ class MockDependencyProvider: DependencyProvider {
     var networkProtectionKeychainTokenStore: NetworkProtectionKeychainTokenStore
     var networkProtectionTunnelController: NetworkProtectionTunnelController
     var connectionObserver: NetworkProtection.ConnectionStatusObserver
+    var serverInfoObserver: NetworkProtection.ConnectionServerInfoObserver
```
samsymons, line 49: We need to keep an instance of this alive from the very beginning so that it can listen for connection changes. Otherwise, when we enter the VPN UI, we won't have any server info to display.

diegoreymendez (approved):
Works great!

PR #3141, opened 2024-07-25
tentatively fix content blocking UI test

brindy:
Task/Issue URL: https://app.asana.com/0/392891325557410/1207863848732212/f
Tech Design URL:
CC:

**Description**:
Tries to fix flakey UI test by adding some wait time.  Otherwise it's not really clear what the problem is.  As the dashboard just seems to appear and then immediately disappear.

**Steps to test this PR**:
1. Check this flow passes https://github.com/duckduckgo/iOS/actions/runs/10080371461
2. Run flow locally using `run_ui_test release/content_blocking.yaml`
3. Optionally run all the end to end flows with `run_ui_test release/`

### DIFF
--- a/.github/workflows/end-to-end.yml
+++ b/.github/workflows/end-to-end.yml
@@ -75,7 +75,7 @@ jobs:
           --header "Accept: application/json" \
           --header "Authorization: Bearer ${{ secrets.ASANA_ACCESS_TOKEN }}" \
           --header "Content-Type: application/json" \
-          --data ' { "data": { "name": "GH Workflow Failure - End to end tests (${{ vars.GITHUB_REF_NAME }})", "workspace": "${{ vars.GH_ASANA_WORKSPACE_ID }}", "projects": [ "${{ vars.GH_ASANA_IOS_APP_PROJECT_ID }}" ], "notes" : "The end to end workflow has failed. See https://github.com/duckduckgo/iOS/actions/runs/${{ github.run_id }}. For instructions on how to handle the failure(s), check https://app.asana.com/0/0/1206423571874502/f" } }'
+          --data ' { "data": { "name": "GH Workflow Failure - End to end tests", "workspace": "${{ vars.GH_ASANA_WORKSPACE_ID }}", "projects": [ "${{ vars.GH_ASANA_IOS_APP_PROJECT_ID }}" ], "notes" : "The end to end workflow has failed. See https://github.com/duckduckgo/iOS/actions/runs/${{ github.run_id }}. For instructions on how to handle the failure(s), check https://app.asana.com/0/0/1206423571874502/f" } }'
 
     - name: Upload logs when workflow failed
       uses: actions/upload-artifact@v4
--- a/.github/workflows/end-to-end.yml
+++ b/.github/workflows/end-to-end.yml
@@ -75,7 +75,7 @@ jobs:
           --header "Accept: application/json" \
           --header "Authorization: Bearer ${{ secrets.ASANA_ACCESS_TOKEN }}" \
           --header "Content-Type: application/json" \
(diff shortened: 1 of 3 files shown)


.github/workflows/end-to-end.yml
```diff
@@ -75,7 +75,7 @@ jobs:
           --header "Accept: application/json" \
           --header "Authorization: Bearer ${{ secrets.ASANA_ACCESS_TOKEN }}" \
           --header "Content-Type: application/json" \
-          --data ' { "data": { "name": "GH Workflow Failure - End to end tests", "workspace": "${{ vars.GH_ASANA_WORKSPACE_ID }}", "projects": [ "${{ vars.GH_ASANA_IOS_APP_PROJECT_ID }}" ], "notes" : "The end to end workflow has failed. See https://github.com/duckduckgo/iOS/actions/runs/${{ github.run_id }}. For instructions on how to handle the failure(s), check https://app.asana.com/0/0/1206423571874502/f" } }'
+          --data ' { "data": { "name": "GH Workflow Failure - End to end tests (${{ vars.GITHUB_REF_NAME }})", "workspace": "${{ vars.GH_ASANA_WORKSPACE_ID }}", "projects": [ "${{ vars.GH_ASANA_IOS_APP_PROJECT_ID }}" ], "notes" : "The end to end workflow has failed. See https://github.com/duckduckgo/iOS/actions/runs/${{ github.run_id }}. For instructions on how to handle the failure(s), check https://app.asana.com/0/0/1206423571874502/f" } }'
```
loremattei: Is this intentional?

.maestro/release_tests/content-blocking.yaml
```diff
@@ -49,4 +49,9 @@ tags:
 
 - tapOn: 
     id: PrivacyIcon
-- assertVisible: "Protections are OFF for this site"
+- extendedWaitUntil: 
+    visible: "Protections are OFF for this site"
```
loremattei: @brindy I'm not sure this fixes the issue, even though I do agree it's a timing issue. 
From what I see, switching protections on and off triggers a page refresh "after a while". If this refresh happens when the dashboard is open, the dashboard is dismissed. This refresh seems to happen even if you trigger a manual refresh, as we do in the test. 
I can reproduce this even manually if I'm fast enough with all the steps. 
So, I _think_ we should wait for the automatic refresh to happen _before_ opening the privacy dashboard (I don't know if we can actually wait for that event in Maestro, but a delay before tapping on the icon should work). 
Does it make sense to you?

brindy: That sounds like a dashboard bug then right?

.github/workflows/end-to-end.yml
brindy: kinda, but I'll roll it back

.maestro/release_tests/content-blocking.yaml
loremattei: Is it a bug? I'd say it's something we could improve, e.g. by triggering the automatic refresh faster or not triggering if a manual refresh has happened in the meantime. Or by pushing an updated state to the dashboard after a refresh instead of dismissing it. But I don't know this is a big problem for users in every day usage and I don't know if any of these things are relatively easy to do. 

But, you are also more familiar than me with the expected behaviour of the dashboard, so if you may see a bug in this behaviour that I don't.

.github/workflows/end-to-end.yml
loremattei: I'm fine with it if it's intentional. No need to rollback. I just wanted to make sure you intended to put it in this PR.

.maestro/release_tests/content-blocking.yaml
brindy: Yeah, I'd say this is a bug. I think it's intentional to close it, but I don't think the user should be able to open it and have it re-close on them.  I will create a task (but I think it's on @jaceklyp  who is out for a bit).  So I will add a temporary work around here by adding a delay if I can and will add to the task to adjust the UI test.

loremattei (approved):
Code changes make sense and the test run successfully locally and on CI.
:shipit:

PR #3173, opened 2024-08-01
Show Intro message for existing users on New Tab Page

dus7:
Task/Issue URL: https://app.asana.com/0/1206226850447395/1207951131408752/f
Tech Design URL:
CC:

**Description**:

Adds a message for existing users about redesign of New Tab Page. I'm using the presence of install date during app launch to determine if the user is new. 
Existing implementation also checks whether the feature flag is set to be releasable remotely and does nothing if it is not. Altering the code is necessary to verify final behavior.

**Steps to test this PR**:
1. Install and run any other build (without the intro message). If you have any other build run already you can skip this step.
1. Change `NewTabPageManager.isAvailableInPublicRelease` to return `true`, to override setup restrictions. Run build from this branch.
1. Set _internal user state_ and enable New Tab Page Sections feature flag in Debug menu.
1. Intro message should be visible on new tab page as well as for next two new tab page impressions.
1. Tapping on X button should hide message forever.
1. Remove app.
1. Launch and enable the feature in Debug menu again - intro message should not be visible.

**Finally:** With code unchanged message should not initialize at all. It can be reset and verified in Debug menu.

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:


federicocappelli:
Pre-review comment:
I've seen the other PR for translations but, as said in https://app.asana.com/0/1205842948507349/1206727265537758/f : 
`Maintaining a Shippable State: To ensure the main branch remains in a shippable state, pull requests containing copy changes should only be merged after the necessary translations have been included.`
Juan is adding this requirement to our Definition of Done right now.
I would suggest to unify the two PRs and merge in one shot. Ping me if the reasons are not clear.

dus7:
Thanks @federicocappelli. I understand the reasons, though technically this is in a shippable state. Whole feature as it is right now is not available publicly unless somebody is an internal user **and** manually switches the flag in Debug menu. At this point there are a few other strings already defined for this feature on main branch with translations pending, so I don't feel holding this one back will have a big influence.

federicocappelli:
- The show-not show behaviour seems to work correctly
- Is the scroll-bar in the right position? it seems a bit far away from the border 
Simulator Screenshot - iPhone 15 Pro Max - 2024-08-02 at 10 01 04
- I'm sure this is not related to your PR but looks like that a new empty tab can't be closed:
Simulator Screenshot - iPhone 15 Pro Max - 2024-08-02 at 10 08 33
If the tab is populated then it can be closed.

federicocappelli:
> Thanks @federicocappelli. I understand the reasons, though technically this is in a shippable state. Whole feature as it is right now is not available publicly unless somebody is an internal user **and** manually switches the flag in Debug menu. At this point there are a few other strings already defined for this feature on main branch with translations pending, so I don't feel holding this one back will have a big influence.

The problem is not feature availability to the general public, merging features with untranslated strings is against our localisation guidelines and DoD because that could cause conflicts with other developers' work, for instance:
- you merge an unlocalised feature
- another developer tries to localise new strings for their feature and ends up with all your strings

This happened in the past and we discussed it at the last weekly Apple meeting.

dus7:
> Is the scroll-bar in the right position? it seems a bit far away from the border

It looks off indeed, but I double-checked and it seems it's a SwiftUI thing.

> I'm sure this is not related to your PR but looks like that a new empty tab can't be closed

Yes, that's a known issue, not related to this PR. Is not affecting release and planned to be fixed: https://app.asana.com/0/72649045549333/1207840822110801/f

DuckDuckGo/it.lproj/Localizable.strings
```diff
@@ -1708,6 +1708,12 @@
 /* Title for the VPN Settings screen. */
 "network.protection.vpn.settings.title" = "Impostazioni VPN";
 
+/* Information message about New Tab Page redesign */
+"new.tab.page.intro.message.body" = "Personalizza i tuoi Preferiti e le funzionalità principali. Riorganizza gli elementi o nascondili per mantenere tutto in ordine.";
+
+/* Title of information message about New Tab Page redesign */
+"new.tab.page.intro.message.title" = "La tua pagina Nuova scheda è... Nuova!";
```
federicocappelli, line 1715: Oh god, if all translations are like this one we sound a bit dumb XD The translation is not terrible but doesn't sound great either. This isn't this PR concern, we don't have a localisation review so no idea how to approach this. Let's ignore it for this PR's sake.

federicocappelli (approved):
LGTM

PR #3186, opened 2024-08-05
new pixel and parameter for monitoring clearing of data stores

brindy:
Task/Issue URL: https://app.asana.com/0/392891325557410/1207974821438701/f
Tech Design URL:
CC:

**Description**:
Adds new pixel fired if there were previous left overs but we have cleared all containers.  Also adds a parameter to show last left over count.

**Steps to test this PR**:
1. Browse some sites
2. Use the fire button
3. Check data was cleared.  
4. If datastores pixel is fired check it has left_over_count parameter

### DIFF
--- a/Core/WebCacheManager.swift
+++ b/Core/WebCacheManager.swift
@@ -96,9 +96,9 @@ public class WebCacheManager {
 extension WebCacheManager {
 
     @available(iOS 17, *)
-    private func checkForLeftBehindDataStores(leftOversCount: Int) async {
+    private func checkForLeftBehindDataStores(previousLeftOversCount: Int) async {
         let params = [
-            "left_overs_count": "\(leftOversCount)"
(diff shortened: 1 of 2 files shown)


Core/WebCacheManager.swift
```diff
@@ -113,10 +119,11 @@ extension WebCacheManager {
         dataStore = nil
 
         let uuids = await WKWebsiteDataStore.allDataStoreIdentifiers
+        let count = max(0, uuids.count - 1) // -1 because there should be a current store
```
miasma13: I would suggest more descriptive naming e.g. `count` -> `leftOverDataStoreCountBeforeClearing` (maybe also pixel param name)

And `ids.count` -> `leftOverDataStoreCountAfterClearing`

miasma13 (approved):

miasma13 (approved):

PR #3191, opened 2024-08-05
ensure new data clearing api is being used

brindy:
Task/Issue URL: https://app.asana.com/0/392891325557410/1207976962313197/f
Tech Design URL: 
CC: @bwaresiak 

**Description**:
Ensure allocation of container id after clearing if on iOS 17 and fix relevant tests.

**Steps to test this PR**:
1. Clean app - browse some sites, use the fire button.  Container id should be nil when data clearing function is called
2. Browse some sites and use the fire button.  Container id should NOT be nil when data clearing function is called
3. Browse some sites and use the fire button.  Container id should NOT be nil when data clearing function is called and should be different from old one
4. Ensure tests all pass

### DIFF
--- a/Core/WKWebViewConfigurationExtension.swift
+++ b/Core/WKWebViewConfigurationExtension.swift
@@ -56,21 +56,29 @@ extension WKWebViewConfiguration {
 
 }
 
-public class DataStoreIdManager {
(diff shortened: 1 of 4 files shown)


miasma13 (approved):
LGTM

Issue #3228, opened 2024-08-13
Add Multipath TCP (MPTCP) support


Aperence:
MPTCP allows a better network reliability by permitting a seamless transition between different network interfaces in case of deterioration of the connection on one of them. Typically, if the connection becomes slow on the Wi-Fi interface, we can transition to the cellular interface without having to recreate the connection. Another use-case could be to use both interface in parallel to speed-up the download of files on this browser.

MPTCP can be enabled by using the [multipathServiceType](https://developer.apple.com/documentation/foundation/urlsessionconfiguration/improving_network_reliability_using_multipath_tcp) option in the URLSessionConfiguration with, for example, the handover mode. Examples on how to enable MPTCP in swift on iOS can be found [on this repository](https://github.com/mptcp-apps/mptcp-hello/tree/main/swift)

See https://mptcp.dev for details about how MPTCP works.

PR #3243, opened 2024-08-15
Issue #3228 Add Multipath TCP (MPTCP) support

awaan05:
#3228 solved.

### DIFF
--- a/DuckDuckGo/AppDelegate.swift
+++ b/DuckDuckGo/AppDelegate.swift
@@ -30,6 +30,7 @@ import Bookmarks
 import Persistence
 import Crashes
 import Configuration
+import Network      // Added later
 import Networking
 import DDGSync
 import RemoteMessaging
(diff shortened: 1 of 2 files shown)


awaan05:
I have not done the testing.

matttbe:
Thanks! Looking forward to seeing MPTCP being supported in this Duckduck Go app!

samsymons:
@awaan05 Hey - thanks for opening this! I'm going to close it as Not Planned for now since additional work is needed for this change to work. Specifically:

* The Multipath entitlement is required
* The approach taken in this PR adds a new `session` property that isn't directly used, and cannot be used from other parts of the codebase
* `multipathServiceType` is only being set on a single `URLSessionConfiguration` instance; any other instances that are created (even from `.default`) will use the default value, not `handover`

However, I've raised this suggestion as a ticket internally so that we can discuss and prioritize the change. Thanks!

PR #3249, opened 2024-08-16
[DuckPlayer] 21. Allow First video

afterxleep:
Task/Issue URL:  https://app.asana.com/0/1204099484721401/1208057529481154/f
**Description**:
Do not show the video overlays when navigating to YouTube via "Watch in YouTube" links.  

**Steps to test this PR**:
1. Make yourself internal user 
-- All Debug Options > Internal User state
2. Add a custom Privacy config. 
--  All Debug Options > Configurations URL.
--  Tap in the first item, and paste the following URL: https://jsonblob.com/api/1253721903368888320
3. Close and re-open settings
4. Confirm DuckPlayer is in "Ask every time" mode
5. Navigate to a Youtube video and when presented with the overlay, tap "Turn On Duck Player" (Do not enable "remember my choice)
6. Once in DuckPlayer, tap the "Watch in Youtube" button
7. Confirm the video loads in Youtube and there is no overlay
8. Refresh the page and confirm there's still no overlay
9. Scroll down and tap another video
10. Confirm the overlay appears again

### DIFF
--- a/DuckDuckGo/DuckPlayer/DuckPlayerNavigationHandler.swift
+++ b/DuckDuckGo/DuckPlayer/DuckPlayerNavigationHandler.swift
@@ -108,6 +108,9 @@ final class DuckPlayerNavigationHandler {
             return
         }
         
+        // This setting allows the FE to skip showing the overlay on some videos
+        duckPlayer.settings.allowFirstVideo = false
+        
         if let (videoID, _) = url.youtubeVideoParams,
             videoID == lastHandledVideoID {
             os_log("DP: URL (%s) already handled, skipping", log: .duckPlayerLog, type: .debug, url.absoluteString)
@@ -118,7 +121,8 @@ final class DuckPlayerNavigationHandler {
          // These should not be handled by DuckPlayer
         if url.isYoutubeVideo,
             url.hasWatchInYoutubeQueryParameter {
-                 return
+                duckPlayer.settings.allowFirstVideo = true
+            return
          }
                 
         if url.isYoutubeVideo,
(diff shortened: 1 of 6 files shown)


DuckDuckGo/DuckPlayer/DuckPlayerNavigationHandler.swift
```diff
@@ -108,6 +108,9 @@ final class DuckPlayerNavigationHandler {
             return
         }
         
+        // This setting allows the FE to skip showing the overlay on some videos
```
Bunn, line 111: Certainly! Here's the improved version of the comment:

"I think this comment could be improved. 'Some videos' is a bit unclear. Does the FE apply any heuristic to decide when to show or not show the overlay? Or is it simpler, as in, if we send this as true, the FE doesn't display the overlay?

The property name itself could also be improved. 'allowFirstVideo' is not very clear to me. What is a 'first video'? If the flag allows the native side to hide the overlay, maybe something like 'skipOverlay' would be better?"

Bunn: I think this comment could be improved. 'Some videos' is a bit unclear. Does the FE apply any heuristic to decide when to show or not show the overlay? Or is it simpler, as in, if we send this as true, the FE doesn't display the overlay?

The property name itself could also be improved. 'allowFirstVideo' is not very clear to me. What is a 'first video'? If the flag allows the native side to hide the overlay, maybe something like 'skipOverlay' would be better?

Bunn (approved):

DuckDuckGo/DuckPlayer/DuckPlayerNavigationHandler.swift
afterxleep: Agree on poor property name, but it matches what's defined in the FE JSON so kept it that way.  I'll improve the comment before merging

PR #3255, opened 2024-08-20
Make remote config accessible to background agents

SlayterDev:
Task/Issue URL: https://app.asana.com/0/1203581873609357/1207165680693234/f
Tech Design URL:
CC:

**Description**:

**Steps to test this PR**:
1. See https://github.com/duckduckgo/BrowserServicesKit/pull/947
2.

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPhone 14 Pro
* [ ] iPad

**OS Testing**:

* [ ] iOS 15
* [ ] iOS 16
* [ ] iOS 17

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:


Core/FileStore.swift
```diff
@@ -22,16 +22,30 @@ import Configuration
...
+                Pixel.fire(pixel: .fileStoreWriteFailed, error: error, withAdditionalParameters: ["config": configuration.rawValue])
+                writeError = error
+            }
+        }
+
+        if let writeError {
+            throw writeError
+        }
+        if let coordinatorError {
+            Pixel.fire(pixel: .fileStoreWriteFailed, error: coordinatorError, withAdditionalParameters: ["config": configuration.rawValue])
```
samsymons: As we discussed, this can be updated to use a coordinator-specific pixel

Core/FileStore.swift
```diff
@@ -22,16 +22,30 @@ import Configuration
...
-        do {
-            try data.write(to: persistenceLocation(for: configuration), options: .atomic)
-        } catch {
-            Pixel.fire(pixel: .fileStoreWriteFailed, error: error, withAdditionalParameters: ["config": configuration.rawValue])
-            throw error
+        let file = persistenceLocation(for: configuration)
+        var coordinatorError: NSError?
+        var writeError: Error?
+
+        NSFileCoordinator().coordinate(writingItemAt: file, options: .forReplacing, error: &coordinatorError) { fileUrl in
```
samsymons, line 34: Let's check that this `persist` function overall is never called on the main thread, since the `coordinate` function will block that thread until it can complete what it's trying to do.

Core/PixelEvent.swift
```diff
@@ -1231,6 +1234,9 @@ extension Pixel.Event {
         case .networkProtectionVPNConfigurationRemoved: return "m_netp_vpn_configuration_removed"
         case .networkProtectionVPNConfigurationRemovalFailed: return "m_netp_vpn_configuration_removal_failed"
 
+        case .networkProtectionConfigurationInvalidPayload: return "m_netp_vpn_configuration_invalid_payload"
```
samsymons: Let's update this to add the `Configuration` name as a suffix.

DuckDuckGo/AppDelegate.swift
```diff
@@ -293,7 +293,7 @@ import os.log
             bookmarksDatabase: bookmarksDatabase,
             appSettings: AppDependencyProvider.shared.appSettings,
             internalUserDecider: AppDependencyProvider.shared.internalUserDecider,
-            configurationStore: ConfigurationStore.shared,
+            configurationStore: ConfigurationStore(),
```
samsymons: Since this used to be a singleton (and I agree it shouldn't be any more), we can probably play it safe and put an instance of it on the `AppDependencyProvider`, just to make sure there are no assumptions that other parts of the codebase are making.

DuckDuckGo/Configuration/ConfigurationManager.swift
```diff
@@ -163,3 +190,39 @@ struct ConfigurationManager {
...
+        updateTrackerBlockingDependencies()
+    }
+
+    override func presentedSubitemDidChange(at url: URL) {
+        guard url == store.fileUrl(for: .privacyConfiguration) else { return }
+        updateTrackerBlockingDependencies()
+    }
+
+    func subscribeToLifecycleNotifications() {
+        NotificationCenter.default.addObserver(self, selector: #selector(addPresenter), name: UIApplication.didBecomeActiveNotification, object: nil)
```
samsymons: Per [file presenter documentation](https://developer.apple.com/documentation/foundation/nsfilecoordinator), we can use `willEnterForegroundNotification` here instead.

PacketTunnelProvider/NetworkProtection/ConfigurationStore.swift
```diff
@@ -0,0 +1,119 @@
...
+struct ConfigurationStore: ConfigurationStoring {
+
+    enum Error: Swift.Error {
+        case unsupportedConfig
+    }
+
+    private var defaults: KeyValueStoring
+
+    private var privacyConfigurationEtagKey: String {
+        return "privacyConfiguration"
```
samsymons, line 35: This key can likely be updated to mention that it's an ETag:

```suggestion
        return "privacyConfigurationEtag"
```

PacketTunnelProvider/NetworkProtection/NetworkProtectionPacketTunnelProvider.swift
```diff
@@ -321,6 +325,20 @@ final class NetworkProtectionPacketTunnelProvider: PacketTunnelProvider {
 
         let settings = VPNSettings(defaults: .networkProtectionGroupDefaults)
 
+        Configuration.setURLProvider(VPNAgentConfigurationURLProvider())
+        configuratioManager = ConfigurationManager()
```
samsymons:
```suggestion
        configurationManager = ConfigurationManager()
```

PacketTunnelProvider/NetworkProtection/NetworkProtectionPacketTunnelProvider.swift
```diff
@@ -321,6 +325,20 @@ final class NetworkProtectionPacketTunnelProvider: PacketTunnelProvider {
...
+        configuratioManager.start()
+        let privacyConfigurationManager = VPNPrivacyConfigurationManager.shared
+        // Load cached config (if any)
+        let configStore = ConfigurationStore()
+        privacyConfigurationManager.reload(etag: configStore.loadEtag(for: .privacyConfiguration), data: configStore.loadData(for: .privacyConfiguration))
+
+        if privacyConfigurationManager.privacyConfig.isSubfeatureEnabled(BackgroundAgentPixelTestSubfeature.pixelTest)
+            && !UserDefaults.configurationGroupDefaults.bool(forKey: BackgroundAgentPixelTestSubfeature.pixelTest.rawValue) {
+            Pixel.fire(pixel: .networkProtectionConfigurationPixelTest)
```
samsymons: Since this is in the init function of the VPN, it won't necessarily react to Privacy Config changes quickly - the VPN is intended to run indefinitely, so we'll want to instead observe config refreshes and fire it there.

We may be able to use the `UniquePixel` API as well, to save some UserDefaults boilerplate.

samsymons: I'd actually consider using the `DailyPixel` API here instead since it'll give us an accurate look each day at the exact rollout of the flag.

PacketTunnelProvider/NetworkProtection/ConfigurationStore.swift
SlayterDev: This was set to match the main app which has more etags then just remote config. None of those keys mention the etag so if we update this one we should update them all in the main app.

Core/FileStore.swift
SlayterDev: I tracked the persist calls all the way up to [here](https://github.com/duckduckgo/iOS/blob/1c736e9c4d9befca0a42ed280b4d47c2a5f38800/DuckDuckGo/AppConfigurationFetch.swift#L172). Basically the full fetch then store process is initiated by the `AppConfigurationFetch` within a `Task`. AFAIK that should put it off the main thread.

PacketTunnelProvider/NetworkProtection/NetworkProtectionPacketTunnelProvider.swift
```diff
@@ -322,6 +328,20 @@ final class NetworkProtectionPacketTunnelProvider: PacketTunnelProvider {
...
+        configurationManager.start()
+        let privacyConfigurationManager = VPNPrivacyConfigurationManager.shared
+        // Load cached config (if any)
+        privacyConfigurationManager.reload(etag: configurationStore.loadEtag(for: .privacyConfiguration), data: configurationStore.loadData(for: .privacyConfiguration))
+
+        configuationSubscription = privacyConfigurationManager.updatesPublisher
+            .sink {
+                if privacyConfigurationManager.privacyConfig.isSubfeatureEnabled(BackgroundAgentPixelTestSubfeature.pixelTest) {
+                    DailyPixel.fire(pixel: .networkProtectionConfigurationPixelTest)
+                }
```
samsymons, line 342: I think it's possible we'll still miss calls to this pixel - this publisher is only called when we receive updates, so if no Privacy Config update is published on a given day then we'll get no events coming through.

We could probably add the `BackgroundAgentPixelTestSubfeature.pixelTest` check and pixel call somewhere that we _know_ will get called regularly, like `handleConnectionStatusChange`. That said, we only really care about getting these calls on days where we're changing the rollout param and at that point we _will_ get this publisher called, so I'd be fine with it staying as-is.

PacketTunnelProvider/NetworkProtection/VPNPrivacyConfigurationManager.swift
```diff
@@ -0,0 +1,137 @@
...
+                    "exceptions": [],
+                    "settings": {}
+                }
+            },
+            "unprotectedTemporary": []
+        }
+    """
+        let data = configString.data(using: .utf8)
+        return data!
+    }
```
samsymons, line 49: I think we should close this gap as quickly as possible; from my perspective we can live with this in the short term, at least as part of running the pixel rollout test, but I'd advocate for us getting real embedded data working ASAP. Would you be able to scope the changes required for this for both iOS and macOS?

PacketTunnelProvider/NetworkProtection/ConfigurationStore.swift
samsymons: No worries, I think we can leave this one in that case - thanks!

PacketTunnelProvider/NetworkProtection/VPNPrivacyConfigurationManager.swift
samsymons, line 1: This class could use unit tests, though I acknowledge there's no good test suite set up for it right now. I'll defer to @diegoreymendez on whether he'd like to block on this, as primary of the VPN AOR.

samsymons (approved):
Approving based on the behavior working well for me across a bunch of different test cases.

PacketTunnelProvider/NetworkProtection/VPNPrivacyConfigurationManager.swift
SlayterDev: Follow up opened here: https://app.asana.com/0/1200890834746050/1208305934034181/f

Core/EtagStorage.swift
```diff
@@ -31,7 +31,7 @@ public protocol BlockerListETagStorage {
 
 public struct UserDefaultsETagStorage: BlockerListETagStorage {
 
-    private let defaults = UserDefaults(suiteName: "com.duckduckgo.blocker-list.etags")
+    private let defaults = UserDefaults(suiteName: "\(Global.groupIdPrefix).app-configuration")
```
diegoreymendez, line 34: Is this a risky transition / migration?  We're moving from using an existing user defaults suite, to using a new one.

Is there anything stored here that will get reset that might cause trouble for users?

Core/FileStore.swift
```diff
@@ -22,16 +22,30 @@ import Configuration
 
 public class FileStore {
 
-    private let groupIdentifier: String = ContentBlockerStoreConstants.groupName
+    private let groupIdentifier: String = ContentBlockerStoreConstants.configurationGroupName
```
diegoreymendez, line 25: Since this is used for the file paths: is there anything that will get reset here that might cause trouble for users?

Core/UserDefaults+NetworkProtection.swift
```diff
@@ -27,6 +27,14 @@ public extension UserDefaults {
         }
         return defaults
     }
+
+    static var configurationGroupDefaults: UserDefaults {
+        let suiteName = "\(Global.groupIdPrefix).app-configuration"
```
diegoreymendez: Can this be replaced by `ContentBlockerStoreConstants. configurationGroupName`?

DuckDuckGo/AppDelegate.swift
```diff
@@ -519,6 +519,7 @@ import os.log
             ContentBlocking.shared.contentBlockingManager.scheduleCompilation()
             AppConfigurationFetch.shouldScheduleRulesCompilationOnAppLaunch = false
         }
+        AppDependencyProvider.shared.configurationManager.loadPrivacyConfigFromDiskIfNeeded()
```
diegoreymendez, line 522: Should this happen before `AppDependencyProvider.shared.configurationManager` is used in line 295?

PacketTunnelProvider/NetworkProtection/VPNPrivacyConfigurationManager.swift
diegoreymendez: Agree that it'd be ideal to have tests here, but I'm not confident the effort to implement those tests is worth additional delay here.  Especially because the effort involved isn't completely clear to me and I don't want to be blocking things here.

PacketTunnelProvider/NetworkProtection/NetworkProtectionPacketTunnelProvider.swift
diegoreymendez: Struggling a bit to follow what `pixelTest` is.

diegoreymendez (commented):
Added some comments.  Mostly questions I have.

Core/EtagStorage.swift
SlayterDev: I don't think there's a risk here. Configuration starts up and fetches all configs at app launch. For an app update this will look like a fresh install for the configs. It will see no etags and just fetch the current versions and store them. I haven't run in to any issues during testing which involved a "migration".

Core/FileStore.swift
SlayterDev: Same as the etags. The module will see no files and download fresh ones. In the case of content blocking/privacy config we have embedded fallbacks while we wait for downloads.

DuckDuckGo/AppDelegate.swift
SlayterDev: It doesn't matter. We don't use the built in timer of the configuration manager because of background app refresh. And even if we did the only purpose of this method is to diff the stored etag vs the etag in the privacy config manager and load fresh from disk if needed.

diegoreymendez (approved):
Nothing more to add, my main doubts have been covered here.

PR #3263, opened 2024-08-21
usage segmentation

brindy:
Task/Issue URL: https://app.asana.com/0/392891325557410/1206955287217877/f
Tech Design URL: https://app.asana.com/0/392891325557410/1208090005627335/f
CC:

**Description**:
This PR ads usage segmentation calculations to the app so that we can monitor usage across segments in a private way, without having to send granular usage information to the server. 

Deviation from tech design:
* Added "Calculation" class / protocol in order to encapsulate the calculation logic exclusively and code it so that it's similar to the python implementation.

**Steps to test this PR**:

***First run, new user***:
1. Reset the simulator (so that the keychain is cleared)
1. Run the app
2. No `m_retention_segments` pixel should be fired

***Subsequent run a few days later***
1. Reset the simulator (so that the keychain is cleared)
1. Modify StatisticsLoader add the following line at the start of the `load` function:

            self.statisticsStore.atb = "445-1"

4. Run the app
5. An assertion failure should crash the app saying this is not a valid ATB. 
6. Update the above line as follows and reset the simulator again:

            self.statisticsStore.atb = "v445-1"

8. Ensure that `m_retention_segments` pixel is fired. There will be various parameters but one should indicate "app_use" activity
9. Re-launch the app
10. No additional pixel should be fired
11. Perform a search
12. The pixel will be fired again and there should be a parameter indicating search activity.
13. Perform another search
14. No additional pixel should be fired

***Return user test***
1. Repeat the above test bit with the following atb.  The pixels sent should include `reinstaller` in the segments parameter.

            self.statisticsStore.atb = "v445-1ru"


Core/Atb.swift
```diff
@@ -19,9 +19,105 @@
...
+    var droppingVariant: String {
+        return String(version.prefix(Atb.template.count))
+    }
+
+}
+
+private extension String {
+
+    func substring(_ range: ClosedRange<Int>) -> String {
+        let startIndex = self.index(self.startIndex, offsetBy: range.lowerBound)
```
aataraxiaa, line 117: Should we account for out-of-bounds here with something like:

```
guard range.lowerBound >= 0, 
              range.upperBound < self.count else {
     return nil
}   
```

and then return an Optional String.

Core/StatisticsLoader.swift
```diff
@@ -141,4 +152,16 @@ public class StatisticsLoader {
...
+        guard let installAtbValue = statisticsStore.atb else { return }
+        let installAtb = Atb(version: installAtbValue, updateVersion: nil)
+        self.usageSegmentation.processATB(atb, withInstallAtb: installAtb, andActivityType: activityType)
+    }
+
+    private func updateUsageSegmentationAfterInstall(activityType: UsageActivityType) {
+        guard let installAtbValue = statisticsStore.atb else { return }
+        let installAtb = Atb(version: installAtbValue, updateVersion: nil)
+        self.usageSegmentation.processATB(installAtb, withInstallAtb: installAtb, andActivityType: activityType)
+    }
```
aataraxiaa: If we wanted we could remove some duplication here with:

```
private func processUsageSegmentation(atb: Atb?, activityType: UsageActivityType) {
    guard let installAtbValue = statisticsStore.atb else { return }
    let installAtb = Atb(version: installAtbValue, updateVersion: nil)
    let actualAtb = atb ?? installAtb
    self.usageSegmentation.processATB(actualAtb, withInstallAtb: installAtb, andActivityType: activityType)
}

private func updateUsageSegmentationWithAtb(_ atb: Atb, activityType: UsageActivityType) {
    processUsageSegmentation(atb: atb, activityType: activityType)
}

private func updateUsageSegmentationAfterInstall(activityType: UsageActivityType) {
    processUsageSegmentation(atb: nil, activityType: activityType)
}
```

Core/UsageSegmentation.swift
```diff
@@ -0,0 +1,100 @@
...
+
+}
+
+protocol UsageSegmenting {
+
+    func processATB(_ atb: Atb, withInstallAtb installAtb: Atb, andActivityType activityType: UsageActivityType)
+
+}
+
+class UsageSegmentation: UsageSegmenting {
```
aataraxiaa: Minor - Maybe could make this `final` if desired.

Core/UsageSegmentation.swift
```diff
@@ -0,0 +1,100 @@
...
+        self.pixelFiring = pixelFiring
+        self.storage = storage
+        self.calculatorFactory = calculatorFactory
+    }
+
+    func processATB(_ atb: Atb, withInstallAtb installAtb: Atb, andActivityType activityType: UsageActivityType) {
+        var atbs = activityType.atbsFromStorage(storage)
+
+        // Fail fast by looking at the end of the list
+        guard !atbs.reversed().contains(where: { $0 == atb }) else { return }
```
aataraxiaa: Unsure how `contains` is implemented internally, but maybe the `reversed` here is not needed?

Core/UsageSegmentation.swift
```diff
@@ -0,0 +1,100 @@
...
+
+        if installAtb != atb {
+           atbs.append(atb)
+        }
+
+        var pixelInfo: [String: String]?
+        let calculator = calculatorFactory.make(installAtb: installAtb)
+        
+        for atb in atbs {
+            pixelInfo = calculator.processAtb(atb, forActivityType: activityType)
```
aataraxiaa, line 75: Is this intended to keep setting the `pixelInfo`? Is the loop required, or would this work:

```
if let lastAtb = atbs.last {
    pixelInfo = calculator.processAtb(lastAtb, forActivityType: activityType)
}
```

Core/UsageSegmentationCalculator.swift
```diff
@@ -0,0 +1,372 @@
...
+final class DefaultCalculatorFactory: UsageSegmentationCalculatorMaking {
+
+    func make(installAtb: Atb) -> any UsageSegmentationCalculating {
+        return UsageSegmentationCalculator(installAtb: installAtb)
+    }
+
+}
+
+/// This implementation is based on https://dub.duckduckgo.com/flawrence/felix-jupyter-modules/blob/master/segments_reference.py and has been written to try and 
+///   resemble the original code as closely as possible.
```
aataraxiaa, line 53: 👍🏼

Core/UsageSegmentation.swift
aataraxiaa: Looking more at the calculator, I think this is intended, as the calculator updates it’s state on every call

Core/UsageSegmentationCalculator.swift
```diff
@@ -0,0 +1,372 @@
...
+        }
+
+        return segments.sorted()
+    }
+
+    /// `count_as_wau`
+    private func countAsWAU(_ atb: Atb) -> Bool {
+        if atb == installAtb {
+            // # Install day - this code should not be running! Report nothing.
+            assertionFailure("See comment")
```
aataraxiaa, line 238: Do we think it would be useful to log any of the `assertionFailure`s?

Core/UsageSegmentationCalculator.swift
```diff
@@ -0,0 +1,372 @@
...
+            return false
+        }
+
+        if previousAtb == nil || previousAtb == installAtb {
+            // # First post-install activity
+            return true
+        }
```
aataraxiaa: We use this check and return logic a lot. Maybe could extract the conditional check to a small method if we want to ensure its consistent across calls.

Core/UsageSegmentationCalculator.swift
```diff
@@ -0,0 +1,372 @@
...
+        static let countAsWAU = "count_as_wau"
+        static let countAsMAUn = "count_as_mau_n"
+
+    }
+
+    let installAtb: Atb
+    var usageHistory = [Atb]()
+    var previousAtb: Atb?
+    var previousWAUSegments = [String]()
+    var previousMAUSegments = Array(repeating: Array(repeating: "", count: 4), count: 4)
```
aataraxiaa: Do all of these properties need to be internal? Could they be `private` or `private (set)`. (Checked `usageHistory` and it could but unsure of the rest).

Core/UsageSegmentation.swift
aataraxiaa: Likewise for all other classes including test cases.

aataraxiaa (commented):
Mostly looks good @brindy , left some comments. Thanks for the useful documentation - it really helped in understanding, particularly for `UsageSegmentationCalculator`.

Also, tested and validated. One unit test is failing but I believe it’s unrelated.

Core/UsageSegmentation.swift
brindy: From just seeing when the `where` block gets called, it appears to start at the beginning and work forward which I guess normally would be fine, but reversing it means it'll fail faster for atbs already in the list.  tbh in the grand scheme of things probably doesn't make much difference from a performance perspective (already managed to optimise this down quite a bit elsewhere).  Will change.

Core/Atb.swift
brindy: Probably using min(self.count, range.upperBound) is better and more like you'd expect from a substring function. Annoying swift just doesn't provide these tbh

Core/StatisticsLoader.swift
brindy: dropped this in directly and tests still pass, thanks

Core/UsageSegmentation.swift
brindy: Done

Core/UsageSegmentation.swift
brindy: Done

Core/UsageSegmentation.swift
brindy: Yep, that's right.  I'll add a comment to make that clearer.

Core/UsageSegmentationCalculator.swift
brindy: Realistically, probably not.  If something is going wrong with this code we'd likely debug it locally anyway.  Also this will occur one time on one day so the chances of us catching logs with this in are pretty remote imo.

Core/UsageSegmentationCalculator.swift
brindy: agreed, done

Core/UsageSegmentationCalculator.swift
brindy: Yeah, they can and probably should be private.

Core/Atb.swift
```diff
@@ -115,7 +115,7 @@ private extension String {
 
     func substring(_ range: ClosedRange<Int>) -> String {
         let startIndex = self.index(self.startIndex, offsetBy: range.lowerBound)
-        let endIndex = self.index(self.startIndex, offsetBy: range.upperBound + 1)
+        let endIndex = self.index(self.startIndex, offsetBy: min(self.count, range.upperBound + 1))
```
aataraxiaa, line 118: 👍🏼

aataraxiaa:
LGTM @brindy . Still a unit test failing - `SyncManagementViewModelTests.testWhenSaveRecoveryPDFPressed_recoveryMethodShown()`, which I assume is unrelated.

aataraxiaa (approved):

PR #3265, opened 2024-08-22
Mitigate Onboarding pixels firing with empty atb

alessandroboron:
Task/Issue URL: https://app.asana.com/0/72649045549333/1208108104325249/f

**Description**:

This PR mitigates the issue that causes onboarding Pixels to fire with an empty ATB.

I split the PR in two commits for easier review:
1. Implement the logic to enqueue and dequeue pixels when needed.
2. Inject the same instance of the OnboardingPixelReporter everywhere it is needed.

**Problem**

There is a race condition where Onboarding pixels are fired as soon the App is launched but the ATB may not be present as it is fetched from an endpoint. 

**Approach**

I modified the `OnboardingPixelReporter` to enqueue Pixels if they need to send the ATB and the ATB is not available.
In the AppDelegate when the `StatisticsLoader` `load` closure completes we ask the OnboardingPixelReporter to fire the enqueued pixels.

**Note**
This is not a bulletproof approach. Currently there’s no check for errors in `StatisticsLoader` `load` closure. If the endpoints fail we will encounter the same issue. Probably this is a broader discussion about what we should do if the ATB/EXTI endpoints fail.

**Screenshot**
After the Fix I can see the atb showing up for the first pixel sent.

Screenshot 2024-08-22 at 12 10 33 PM

**Steps to test this PR**:
1. Repeat this test for `mb` and `ma` variant.
2. Set `Pixel.isDryRun = false` in `AppDelegate` line 133.
3. Replace  requestInstallStatistics(completion: completion) in `StatisticsLoader` line 46 with         
```
DispatchQueue.main.asyncAfter(deadline: .now() + 5.0) {
    self.requestInstallStatistics(completion: completion)
}
```
4. Launch the App.
5. Ensure that Pixel are shown in Kibana and ATB is not empty.
6. Remove changed code
7. Ensure `OboardingPixelReporter` tests are all green.

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPhone 14 Pro
* [ ] iPad

**OS Testing**:

* [ ] iOS 15
* [ ] iOS 16
* [ ] iOS 17

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:

### DIFF
--- a/DuckDuckGo/AppDelegate.swift
+++ b/DuckDuckGo/AppDelegate.swift
@@ -100,6 +100,7 @@ import WebKit
     private var didCrashDuringCrashHandlersSetUp: Bool
 
     private let launchOptionsHandler = LaunchOptionsHandler()
+    private let onboardingPixelReporter = OnboardingPixelReporter()
 
     override init() {
         super.init()
@@ -336,7 +337,7 @@ import WebKit
                                           variantManager: variantManager,
                                           contextualOnboardingPresenter: ContextualOnboardingPresenter(variantManager: variantManager),
                                           contextualOnboardingLogic: daxDialogs,
-                                          contextualOnboardingPixelReporter: OnboardingPixelReporter())
+                                          contextualOnboardingPixelReporter: onboardingPixelReporter)
 
             main.loadViewIfNeeded()
             syncErrorHandler.alertPresenter = main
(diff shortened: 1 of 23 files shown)


brindy (approved):
Works as described! Nice :)

PR #3276, opened 2024-08-26
Add error pixels for Subscription keychain access errors

miasma13:
Task/Issue URL: https://app.asana.com/0/1201037661562251/1208146974665104/f

**Description**:
Add missing pixel reporting for subscription related keychain access errors.

**Steps to test this PR**:
1. Mock keychain failure in `AccountManager` e.g. by replacing [this](https://github.com/duckduckgo/BrowserServicesKit/blob/4a55217003ad7b2d44a1ac616d47596c0bda69dc/Sources/Subscription/Managers/AccountManager.swift#L153) call  with `throw AccountKeychainAccessError.keychainSaveFailure(555)` 
2. Enter the flow to trigger the above call
3. Ensure pixel with appropriate parameters is fired

---
###### Internal references:


DuckDuckGo/Subscription/DefaultSubscriptionManager+AccountManagerKeychainAccessDelegate.swift
```diff
@@ -0,0 +1,31 @@
...
+import Foundation
+import Core
+import Subscription
+
+extension DefaultSubscriptionManager: AccountManagerKeychainAccessDelegate {
+
+    public func accountManagerKeychainAccessFailed(accessType: AccountKeychainAccessType, error: AccountKeychainAccessError) {
+        Pixel.fire(.privacyProKeychainAccessError, withAdditionalParameters: [
+            PixelParameters.privacyProKeychainAccessType: accessType.rawValue,
+            PixelParameters.privacyProKeychainError: error.errorDescription])
```
samsymons: One possible issue here is that we won't be able to fully determine the scope of an issue, which has been an issue for Keychain errors in the past. One user can get into a bad state which then causes a stream of failures to roll in.

With the `PixelKit` API we have the ability to set a `dailyAndCount` frequency type, which can tell us the number of users affected vs. number of total events, though I don't recall if we're set up to use it on iOS. Probably the `DailyPixel` API could be used instead.

samsymons (approved):

PR #3278, opened 2024-08-26
fix back forward swipe gesture in landscape

brindy:
Task/Issue URL: https://app.asana.com/0/392891325557410/1208145984254241/f
Tech Design URL: 
CC:

**Description**:
Fixes the constraints of the uistackview so that the webview uses the full area and so that the back and forward navigation gestures works.

**Steps to test this PR**:
1. Reset your simulator and set the variant manager to only use `mb` (the new onboarding) by commenting out the other variant.
2. Launch the app and browse to some sites.  Switch between landscape and portrait and check that back and forward navigation gestures still work.
3. Confirm that new dax onboarding still works as expected (ie doesn't look broken because of this change)
4. Check on small phone, large phone with notch, iPad


alessandroboron:
@brindy it’s fixed, thanks! I’m approving it! Was the stackView not constrained to the safe area?

While testing I noticed something that I will need to address in the task to fix the Instagram/Youtube dialog.

When navigating back and forward using gesture it seems that the web view delegate method `webView(_ webView: WKWebView, didFinish navigation: WKNavigation!)` doesn't get called?. 

This means that during the onboarding the user can swipe back and the dax dialog doesn’t get updated (Video attached).

https://github.com/user-attachments/assets/63ef2b7d-8862-4920-8d4d-5d15038b0bcf

One simple workaround could be to disable `allowsBackForwardNavigationGestures` during the onboarding. An other solution is to check if `navigationAction.navigationType == .backForward` and perform the logic. I’ll evaluate pros and cons of both.

alessandroboron (approved):

bwaresiak (approved):

brindy:
@alessandroboron 

> Was the stackView not constrained to the safe area?

Yes, that was the problem - the UIStackView was using safe area for constraints, but it needed to use the superview so that it can go to the extents of the screen. The webview takes care of the safe area on its own for rendering content. :)

PR #3284, opened 2024-08-28
[DuckPlayer] 25. Allow www.youtube.com as origin

afterxleep:
Task/Issue URL: https://app.asana.com/0/1204099484721401/1208167725782497/f

**Description**:
- Add additional domain to show overlay on iPad

**Steps to test this PR**:
1. Open App in iPad sim
2. Add custom configuration from https://www.jsonblob.com/api/1276215786342309888
3. Make yourself internal user
4. Set DuckPlayer to 'Always Ask'.  Settings > DuckPlayer
5. Open youtube.com
6. Search for 'Sad but true' and pick a video
7. Confirm the Youtube overlay shows up on iPad. (It's the same as macOS, but that will change in prod)

### DIFF
--- a/DuckDuckGo/DuckPlayer/YoutubeOverlayUserScript.swift
+++ b/DuckDuckGo/DuckPlayer/YoutubeOverlayUserScript.swift
@@ -55,25 +55,6 @@ final class YoutubeOverlayUserScript: NSObject, Subfeature {
             .store(in: &cancellables)
     }
     
-    enum MessageOrigin {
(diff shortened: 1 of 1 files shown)


Bunn:
This is what I'm seeing on iPad
image

Not quite the same as the one in macOS
image

Not sure if it's a problem, just mentioned since you wrote  _(It's the same as macOS, but that will change in prod)_ But other than that it's working.

Bunn (approved):

PR #3319, opened 2024-09-04
fix tab switcher crashes (speculative fix)

brindy:
Task/Issue URL: https://app.asana.com/0/414235014887631/1207531239053735/f
Tech Design URL:
CC:

**Description**:
Use a batch update to queue changes, hopefully to fix the crash.

**Steps to test this PR**:
1. Open and close the tab switcher
2. Open and close one tab
3. Open, close and re-order many tabs
4. Using something to create a slow network (I use ProxyMan's "slow network conditions") load pages and then update the order of tabs in the tab switcher and also close the tab where the request is running

### DIFF
--- a/DuckDuckGo/TabSwitcherViewController.swift
+++ b/DuckDuckGo/TabSwitcherViewController.swift
@@ -518,9 +518,12 @@ extension TabSwitcherViewController: TabObserver {
             return
         }
 
-        if let index = tabsModel.indexOf(tab: tab), index < collectionView.numberOfItems(inSection: 0) {
-            collectionView.reconfigureItems(at: [IndexPath(row: index, section: 0)])
-        }
+        collectionView.performBatchUpdates({}, completion: { [weak self] completed in
+            guard completed, let self = self else { return }
+            if let index = self.tabsModel.indexOf(tab: tab), index < self.collectionView.numberOfItems(inSection: 0) {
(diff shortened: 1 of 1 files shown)


DuckDuckGo/TabSwitcherViewController.swift
```diff
@@ -518,9 +518,12 @@ extension TabSwitcherViewController: TabObserver {
             return
         }
 
-        if let index = tabsModel.indexOf(tab: tab), index < collectionView.numberOfItems(inSection: 0) {
-            collectionView.reconfigureItems(at: [IndexPath(row: index, section: 0)])
-        }
+        collectionView.performBatchUpdates({}, completion: { [weak self] completed in
```
samsymons, line 521: Interesting, so in this case we aren't actually making any batch changes, but rather just waiting for the completion? Not sure I totally follow it, but functionally it works great for me locally.

samsymons (approved):
Approved based on stress-testing the change in a number of ways locally, works as expected - added one comment about the logic just for my own understanding.

DuckDuckGo/TabSwitcherViewController.swift
brindy: yeah my thinking is if we're trying to reconfigure while some other batch is in progress then it's safer to wait. tbh it's just a guess so we'll see how it goes for a week before it gets released.  Thanks.

PR #3326, opened 2024-09-05
defer loading the tab switcher button until view did load

brindy:
Task/Issue URL: https://app.asana.com/0/414709148257752/1208092544762425/f
Tech Design URL: 
CC: @bwaresiak 

**Description**:
Defer loading the tab switcher button until view did load to avoid Lottie trying to do something with the view hierarchy when not available.

**Steps to test this PR**:
1. On the phone, open, close and re-order tabs. Make sure tab switcher button reflects correct state.
3. On the iPad, open, close and re-order tabs. Make sure tab switcher button reflects correct state.
4. Enable "app lock" from settings. Background app. Ensure blank snapshot view is correct.  Foreground app and open / close tabs.
5. Launch the app using a link from cold and while in the background.

### DIFF
--- a/DuckDuckGo/BlankSnapshotViewController.swift
+++ b/DuckDuckGo/BlankSnapshotViewController.swift
@@ -35,7 +35,7 @@ class BlankSnapshotViewController: UIViewController {
     
     let menuButton = MenuButton()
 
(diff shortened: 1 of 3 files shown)


dus7 (approved):
🤞

PR #3346, opened 2024-09-10
C.S.S Bump (Via BSK)

afterxleep:
Task/Issue URL: https://app.asana.com/0/1204099484721401/1208271424696738/f

BSK PR: https://github.com/duckduckgo/BrowserServicesKit/pull/983

**Description**:
- Updates C.S.S
- Adds subfeature flag

**Steps to test this PR**:
1. Make yourself internal
2. Add custom config via https://www.jsonblob.com/api/1276215786342309888
3. Smoke Test Duck Player


Bunn (approved):
LGTM.

I dind't see this "Adds subfeature flag" being used in the actual code, only added to BSK, is this expected?

Also, I found an issue that looks like an edge case. If you open settings and quickly tap back, it will take you back to the Duck Player settings again. Please see the attached video.

I don't think this is caused by this PR; I suggest filing it as a separate task.

https://github.com/user-attachments/assets/30dee4ac-65ea-446b-9528-de705c0034b0

afterxleep:
> I dind't see this "Adds subfeature flag" being used in the actual code, only added to BSK, is this expected?

No.  There was a missing commit.  Just pushed.

> Also, I found an issue that looks like an edge case. If you open settings and quickly tap back, it will take you back to the Duck Player settings again. Please see the attached video.

Correct.  Not caused by this PR.  This is how deeplinks in settings are currently working.  There's not much we can do to fix that for now, but I'll keep track of that.

PR #3357, opened 2024-09-12
Add Marketplace Postback handling

Bunn:
Task/Issue URL: https://app.asana.com/0/72649045549333/1208126219488944/f
Tech Design URL: https://app.asana.com/0/72649045549333/1208274586553401/f

**Description**:
Add Marketplace Postback handling

**Steps to test this PR**:
1. Add a breakpoint [here](https://github.com/duckduckgo/iOS/blob/0425ec0aab17d0adc196a3e5a8a7a5dfe57a8670/Core/MarketplaceAdPostbackUpdater.swift#L52)
2. Either comment out the conditional check [here](https://github.com/duckduckgo/iOS/blob/d271145e88698500b58d4af1a6fe35db56617817/DuckDuckGo/AppDelegate.swift#L763) or use this config (https://jsonblob.com/api/1283828391185604608)
3. Run the app as a new user (never installed before), validate that we call postBack value `0` abd coarse value `low`
4. Do the same as before with a returning user, validate that we call postBack value `1` abd coarse value `high`
5. Validate that we have https://duckduckgo.com set as `Advertising attribution report endpoint URL` in the plist 
---
1 - erase all content and settings
2 - Run app
3 - Validate that the postback is not sent
4 - Set the https://jsonblob.com/api/1283828391185604608 config or remove the feature flag check
5 - Run the app again
6 - Validate that the postback is sent as a new user

1 - Remove the app
2 - Run the app again
3 - Set the https://jsonblob.com/api/1283828391185604608 config or remove the feature flag check (If you had removed the flag on code before, skip steps 4, 5 and 6)
4 - Validate that the postback is not sent
5 - Set the https://jsonblob.com/api/1283828391185604608 config or remove the feature flag check
6 - Run the app again
7 - Validate that the postback is sent as a returning user


Core/MarketplaceAdPostbackUpdater.swift
```diff
@@ -0,0 +1,73 @@
...
+            Logger.general.debug("Attribution: AdAttributionKit postback succeeded")
+        } catch {
+            Logger.general.error("Attribution: AdAttributionKit postback failed \(error.localizedDescription, privacy: .public)")
+        }
+    }
+
+    @available(iOS 16.1, *)
+    private func updateSKANPostback(_ postback: MarketplaceAdPostback, lockPostback: Bool) async {
+        do {
+            try await SKAdNetwork.updatePostbackConversionValue(postback.fineValue,
```
samsymons: At risk of seeming like we're going backwards, what do you think about using the completion handler based API for SKAN? The reason I mention it is that we're seeing rare crashes with this call after moving to the async API. You can see an example [here](https://errors.duckduckgo.com/organizations/ddg/issues/104096/events/ab29c80e711f11efbf32499bdc26619c/).

samsymons (approved):
LGTM, left one optional comment about a rare crash with the SKAN API.

Core/MarketplaceAdPostbackUpdater.swift
Bunn: Since we're seeing this error I believe it makes sense, then, if we still see a problem with these calls, i.e: nothing really changes, we can always revert it back to async. I'll do the change

Core/MarketplaceAdPostbackUpdater.swift
```diff
@@ -0,0 +1,77 @@
...
+        /// https://errors.duckduckgo.com/organizations/ddg/issues/104096/events/ab29c80e711f11efbf32499bdc26619c/
+
+        SKAdNetwork.updatePostbackConversionValue(postback.fineValue,
+                                                  coarseValue: postback.SKAdCoarseValue) { error in
+            if let error = error {
+                Logger.general.error("Attribution: SKAN 4 postback failed \(String(describing: error), privacy: .public)")
+            } else {
+                Logger.general.debug("Attribution: SKAN 4 postback succeeded")
+            }
+        }
```
samsymons, line 79: Since this call doesn't work on the simulator, we could wrap it in a target OS check just to reduce any error logging to the console when debugging. (This was a problem with the previous implementation as well)

samsymons (approved):
LGTM, tested the new/returning check at app launch, and that the postback call is made correctly after the remote config has been refreshed. Left one non-blocking comment about the attribution APIs not working in the simulator.

PR #3358, opened 2024-09-13
Pixel retrying

samsymons:
Task/Issue URL: https://app.asana.com/0/72649045549333/1208300963176861/f
Tech Design URL:
CC:

**Description**:

This PR adds pixel retrying behavior for a small set of pixels.

**Steps to test this PR**:

1. To test this, we should either fire real pixels by disabling `dryRun` mode or simulate a failed error call by adding an error to `Pixel.swift#240` - adjust the following steps based on which path you take (i.e., enabling Airplane Mode to test real pixel call failures, or removing the fake error and rebuilding)
1. Sign into a Privacy Pro account, easiest way to get one is [here](https://app.asana.com/0/0/1208034056804149); once you have it then open the Privacy Pro settings and follow the "I have an account" flow
2. Enable and start the VPN by opening the VPN UI
3. Check that you see log entries that say "Saving persistent pixel" followed by `m_netp_controller_start_attempt` and `m_netp_controller_start_success` pixels (note: even if you're offline, it's expected to see the success pixel, as the success pixel only reports the request we send to the OS to start the VPN, which is not affected by connectivity)
4. Now put your device back into a state that lets it send pixels (disable Airplane Mode, or remove the hardcoded pixel send error) and wait 60 seconds
5. Background and foreground the app and check that the pixels have been retried; look for `Persistent pixel retrying` in the logs and then search for the pixel names (`m_netp_controller_start_attempt`, `m_netp_controller_start_success `)

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPhone 14 Pro
* [ ] iPad

**OS Testing**:

* [ ] iOS 15
* [ ] iOS 16
* [ ] iOS 17

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:


samsymons:
This still needs a bit of clean-up and filled out testing instructions, I'll do that before Monday then mark it ready for review.

DuckDuckGo/NetworkProtectionTunnelController.swift
```diff
@@ -128,16 +130,33 @@ final class NetworkProtectionTunnelController: TunnelController, TunnelSessionPr
     /// Starts the VPN connection used for Network Protection
     ///
     func start() async {
-        Pixel.fire(pixel: .networkProtectionControllerStartAttempt, includedParameters: [.appVersion, .atb])
+        persistentPixel.fire(
+            pixel: .networkProtectionControllerStartAttempt,
+            error: nil,
+            includedParameters: [.appVersion, .atb],
+            withAdditionalParameters: [:],
+            onComplete: { _ in })
```
samsymons, line 138: Error handling here could be improved by sending a write failure pixel

Core/PersistentPixelStoring.swift
```diff
@@ -0,0 +1,157 @@
...
+                pixels.removeFirst()
+            }
+
+            try writePixelDataToFileSystem(pixels: pixels)
+        }
+    }
+
+    func replaceStoredPixels(with pixels: [PersistentPixelMetadata]) throws {
+        try fileAccessQueue.sync {
+            try writePixelDataToFileSystem(pixels: pixels)
```
bwaresiak, line 113: Shouldn't we also check the `pixelCountLimit` here?

Core/PersistentPixel.swift
```diff
@@ -0,0 +1,326 @@
...
+            try self.persistentPixelStorage.append(pixel: pixelMetadata)
+        }
+    }
+
+    private func stopProcessingQueueAndPersistPendingPixels() {
+        pixelProcessingLock.lock()
+        defer { pixelProcessingLock.unlock() }
+
+        for pixel in self.failedPixelsPendingStorage {
+            try? self.persistentPixelStorage.append(pixel: pixel)
```
bwaresiak: Can we append as batch?

Core/PersistentPixelStoring.swift
```diff
@@ -0,0 +1,157 @@
...
+        }
+    }
+
+    func append(pixel: PersistentPixelMetadata) throws {
+        try fileAccessQueue.sync {
+            var pixels = try self.readStoredPixelDataFromFileSystem()
+            pixels.append(pixel)
+
+            if pixels.count > pixelCountLimit {
+                pixels.removeFirst()
```
bwaresiak: That drop does not necessairly remove the oldest - because of how the failed pixel array is behaving.

bwaresiak (changes requested):
I've went through the code, overall it looks good but there are some aspects I'd like to go through with you, so just ping me when you're around. :)

Core/PersistentPixelStoring.swift
samsymons: Done; this function has been replaced with new `append` and `remove` functions, where the `append` one can append an array of any size but will take the last 100 elements before writing them back to disk.

Core/PersistentPixel.swift
samsymons: This is resolved now, we no longer re-append failed pixels as the `failedPixelsPendingStorage` mechanism has been entirely removed.

We now collect a set of successfully-retried pixel UUIDs that get filtered out at the end of the retrying process. This means that incoming pixels can now be written to storage while processing is active, since we're no longer replacing the entire post-processing pixel array.

Core/PersistentPixelStoring.swift
samsymons: The order of the array is now preserved - pixels get written immediately upon failure, and we filter them out when they get successfully retried, so we're no longer juggling two arrays at once and combining them in unexpected ways.

That said, it's still possible for pixels to get a little bit out of order; for example:
1. The app tries to send pixel 1, and then tries to send pixel 2 a second later
2. Pixel 1, for whatever reason, takes longer for its network request to fail
3. Because we capture the timestamp at the time of the pixel send initiation, pixel 1 is now positioned _after_ pixel 2 in the storage despite having a timestamp 1 second earlier

This feels low risk to me so I've opted to leave it, but adding a sorting mechanism to the storage call wouldn't be terribly expensive; let me know how you feel.

Core/PersistentPixel.swift
```diff
@@ -218,46 +217,49 @@ public final class PersistentPixel: PersistentPixelFiring {
...
         let currentDate = dateGenerator()
 
         for pixelMetadata in queuedPixels {
             if let originalSendDateString = pixelMetadata.timestamp,
                let originalSendDate = dateFormatter.date(from: originalSendDateString),
                let date28DaysAgo = calendar.date(byAdding: .day, value: -28, to: currentDate) {
                 if originalSendDate < date28DaysAgo {
+                    pixelIDsAccessQueue.sync {
+                        _ = pixelIDsToRemove.insert(pixelMetadata.id)
+                    }
```
samsymons, line 252: We were getting outdated pixels removed for free because they weren't getting added to the failed pixels array, so now we need to explicitly return them as part of the `pixelIDsToRemove`.

Core/PersistentPixel.swift
```diff
@@ -299,22 +301,13 @@ public final class PersistentPixel: PersistentPixelFiring {
 
         Logger.general.debug("Saving persistent pixel named \(pixelMetadata.pixelName)")
 
-        if self.isSendingQueuedPixels {
-            self.failedPixelsPendingStorage.append(pixelMetadata)
-        } else {
-            try self.persistentPixelStorage.append(pixel: pixelMetadata)
-        }
+        try self.persistentPixelStorage.append(pixels: [pixelMetadata])
```
samsymons, line 304: It's now possible to write pixels to storage even when the queue is actively being processed, since we now only remove pixels based on ID - newly failed pixels will be untouched at the end of processing.

Core/PersistentPixelStoring.swift
```diff
@@ -0,0 +1,176 @@
...
+    var timestamp: String? {
+        return additionalParameters[PixelParameters.originalPixelTimestamp]
+    }
+}
+
+protocol PersistentPixelStoring {
+    func append(pixels: [PersistentPixelMetadata]) throws
+    func remove(pixelsWithIDs: Set<UUID>) throws
+    func storedPixels() throws -> [PersistentPixelMetadata]
+}
```
samsymons, line 46: There's now no way to remove pixels from the array without referencing them by ID, so we don't have to worry about newly-appended pixels getting overwritten somehow, like was possible with the old approach.

Core/PersistentPixel.swift
```diff
@@ -0,0 +1,299 @@
...
+
+        dailyPixelFiring.fireDailyAndCount(
+            pixel: pixel,
+            error: error,
+            withAdditionalParameters: additionalParameters,
+            includedParameters: includedParameters,
+            pixelFiring: Pixel.self,
+            dailyPixelStore: DailyPixel.storage,
+            onDailyComplete: { dailyError in
+                if dailyError != nil {
```
bwaresiak: Shouldn't we ignore `Error.alreadyFired` ? There doesn't seem to be any tests for `fireDailyAndCount`

Core/PersistentPixel.swift
```diff
@@ -0,0 +1,299 @@
...
+                        if let error { additionalParameters.appendErrorPixelParams(error: error) }
+                        Logger.general.debug("Saving persistent count pixel named \(pixel.name)")
+                        try self.persistentPixelStorage.append(pixels: [
+                            PersistentPixelMetadata(
+                                eventName: pixel.name,
+                                pixelType: .count,
+                                additionalParameters: additionalParameters,
+                                includedParameters: includedParameters
+                            )
+                        ])
```
bwaresiak, line 181: Validating assumptions here: typically daily/count combo is used for pixels that may fire more frequently and we want to understand affected devices - in case 'count' pixel floods us with data it can easily overflow the array (discarding other pixels) so we only get one signal.

I'm not saying we need to handle this now, but is seems like a hole in the current approach.

Core/PersistentPixel.swift
```diff
@@ -0,0 +1,299 @@
...
+                }
+            }
+        }
+    }
+
+    public func fireDailyAndCount(pixel: Pixel.Event,
+                                  error: Swift.Error? = nil,
+                                  withAdditionalParameters additionalParameters: [String: String],
+                                  includedParameters: [Pixel.QueryParameters] = [.appVersion],
+                                  completion: @escaping ([Error]) -> Void = { _ in }) {
```
bwaresiak: Maybe we should use tuple and named parameters here?

Core/PersistentPixelStoring.swift
```diff
@@ -0,0 +1,176 @@
...
+        } else if let appSupportDirectory = fileManager.urls(for: .applicationSupportDirectory, in: .userDomainMask).first {
+            self.storageDirectory = appSupportDirectory
+        } else {
+            fatalError("Unable to locate application support directory")
+        }
+    }
+
+    func append(pixels newPixels: [PersistentPixelMetadata]) throws {
+        try fileAccessQueue.sync {
+            var pixels = try self.readStoredPixelDataFromFileSystem()
```
bwaresiak, line 94: I'd consider caching the state in memory so we can limit the number of reads.

Core/PersistentPixel.swift
```diff
@@ -0,0 +1,299 @@
...
+        let dispatchGroup = DispatchGroup()
+
+        let pixelIDsAccessQueue = DispatchQueue(label: "Failed Pixel Retry Attempt Metadata Queue")
+        var pixelIDsToRemove: Set<UUID> = []
+        let currentDate = dateGenerator()
+
+        for pixelMetadata in queuedPixels {
+            if let originalSendDateString = pixelMetadata.timestamp,
+               let originalSendDate = dateFormatter.date(from: originalSendDateString),
+               let date28DaysAgo = calendar.date(byAdding: .day, value: -28, to: currentDate) {
```
bwaresiak: I think `date28DaysAgo` can be computed once outside of the loop.

Core/PersistentPixelStoring.swift
```diff
@@ -0,0 +1,176 @@
...
+        self.pixelType = pixelType
+        self.additionalParameters = additionalParameters
+        self.includedParameters = includedParameters
+    }
+
+    var pixelName: String {
+        switch pixelType {
+        case .daily: return eventName + "_d"
+        case .count: return eventName + "_c"
+        case .regular: return eventName
```
bwaresiak: I recall we wanted to move away from _d and _c - having this hardcoded here will make it harder to do so. Can we just store name along with suffix in `eventName`? I'm not sure there is any real value of storing if given pixel was daily or count other than name handling.

bwaresiak (changes requested):

Core/PersistentPixel.swift
samsymons: There are two tests already (`testWhenDailyAndCountPixels*`), but they miss this case - I forget that DailyPixel can error even in normal usage, good catch. Fixed this and added an additional test for it, thanks!

bwaresiak: Right, I missed these when searching for them. :)

Core/PersistentPixelStoring.swift
samsymons: The `DailyPixel` API has been updated in the following way:
1. It now accepts a tuple of suffixes when sending a pixel
2. Any code using the `fireDailyAndCount` API currently has been updated to use the legacy suffixes
3. Finally, the `fireDailyAndCount` API was given new `_daily` and `_count` suffixes by default, so that no usage of the legacy suffixes gets added to the codebase going forward

The `PersistentPixel` class will now use whatever suffixes are given to it, and stores the final pixel name (including suffix) to disk, so the `PersistentPixelMetadata` struct no longer cares about which type of pixel it's dealing with.

Because of step 2 the PR has grown in size a bit, but for testing it you can remove the default value from `DailyPixel` and confirm that the app still compiles, as all uses have been accounted for and no pixels have been accidentally updated to use the new suffixes.

DuckDuckGoTests/DailyPixelTests.swift
```diff
@@ -349,19 +354,14 @@ final class DailyPixelTests: XCTestCase {
...
 
-    func testThatDailyPixelWithCountWillAppendDToPixelNameForDaily() {
+    func testThatDailyPixelWithLegacyPixelSuffixAndCountWillAppendDAndC() {
         let expectation = XCTestExpectation()
 
-        stub { request in
-            request.url?.absoluteString.contains(Pixel.Event.forgetAllPressedBrowsing.name + "_d") == true
-        } response: { _ in
-            return HTTPStubsResponse(data: Data(), statusCode: 200, headers: nil)
-        }
```
samsymons, line 359: The stubs in this test file weren't actually being used this entire time, since each test is using a mock pixel sender – as a result, these tests were not really testing anything. I've changed them to instead look at the pixel names sent to the mock.

Core/PersistentPixel.swift
samsymons: I've left this alone for the time being as I feel I'll still get some value from the current state, but I'd like to come back to it next week. At the same time, I'd also like to triage and add a storage error pixel, per a previous comment. Let me know if you feel strongly about addressing this one now before it's merged though.

Core/PersistentPixel.swift
```diff
@@ -0,0 +1,292 @@
...
+                        countPixelStorageError = error
+                    }
+                }
+
+                dispatchGroup.leave()
+            }
+        )
+
+        dispatchGroup.notify(queue: .global()) {
+            let errors = [dailyPixelStorageError, countPixelStorageError].compactMap { $0 }
```
bwaresiak: Some leftover? Weird that linter didn't fail...

bwaresiak (approved):
Great job Sam, left one more comment around some warnings, but overall it's approved :)

Core/PersistentPixel.swift
samsymons: Huh, this is surprising. Let me fix it - thanks!

PR #3375, opened 2024-09-20
Show Add Favorite placeholder as last favorite item on New Tab Page

dus7:
Task/Issue URL: https://app.asana.com/0/72649045549333/1208244619690578/f
Tech Design URL:
CC:

**Description**:

Adds a possibility to show an Add button in favorites collection, allowing to create a favorite right from New Tab Page. ([Figma](https://www.figma.com/design/N2GbF5HEvopp5iwmAlMwyD/New-Tab-Page-Customization?node-id=611-147700&m=dev)).
Add button is currently navigating to an empty view, which is expected. The view will be implemented in additional PR.

In order to reduce the amount of duplicated code and separate additional display elements from pure Favorites data,  DataSource for Favorites was defined which is used in common ViewModel logic for Favorites, hence the rename of `NewTabPageModel` -> `NewTabPageViewModel`.

There was also a need to keep some of the items immovable, which required a change in `ReorderableForEach`.

**Steps to test this PR**:

1. Open NTP without Favorites, + button should be visible as a first element.
2. Add a couple of Favorites, + button should be visible as last item.
3. Try reordering Favorites, + button should be kept stationary and it shouldn't be posible to drop favorite after it.

---
###### Internal references:

### DIFF
--- a/DuckDuckGo/FavoritesEmptyStateView.swift
+++ b/DuckDuckGo/FavoritesEmptyStateView.swift
@@ -18,12 +18,11 @@
 //
 
 import SwiftUI
+import DuckUI
 
 struct FavoritesEmptyStateView<Model: FavoritesEmptyStateModel>: View {
-    @Environment(\.horizontalSizeClass) var horizontalSizeClass
-    @Environment(\.isLandscapeOrientation) var isLandscape
-
     @ObservedObject var model: Model
+    @Binding var isAddingFavorite: Bool
 
     let geometry: GeometryProxy?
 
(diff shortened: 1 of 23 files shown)


DuckDuckGo/FavoritesEmptyStateView.swift
```diff
@@ -55,5 +62,17 @@ struct FavoritesEmptyStateView<Model: FavoritesEmptyStateModel>: View {
 }
 
 #Preview {
-    return FavoritesEmptyStateView(model: FavoritesPreviewModel(), geometry: nil)
+    PreviewViewWrapper()
+}
+
+private struct PreviewViewWrapper: View {
```
dus7, line 68: I was using Xcode16 and it gives a warning when `@State` is used inside `Preview` without @Previewable tag, but it's available only from iOS17. I figured wrapper here is the quickest solution for now.

brindy (approved):
LGTM - works as advertised.

PR #3419, opened 2024-10-07
Increases the wait time for a breaking test

diegoreymendez:
Task/Issue URL: https://app.asana.com/0/0/1208389158134600/f

## Description

Increases the wait time for a breaking test.  Locally the test works just fine so I'm guessing it's something with the CI making the test slower.

## Testing

Check the CI is green.

---
###### Internal references:

### DIFF
--- a/DuckDuckGoTests/ContextualDaxDialogsFactoryTests.swift
+++ b/DuckDuckGoTests/ContextualDaxDialogsFactoryTests.swift
@@ -341,13 +341,13 @@ extension ContextualDaxDialogsFactoryTests {
 
     func testDialogDefinedBy(spec: DaxDialogs.BrowsingSpec, firesEvent event: Pixel.Event) {
         // GIVEN
-        let expectation = self.expectation(description: "asd")//#function)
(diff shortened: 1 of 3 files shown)


DuckDuckGoTests/ContextualDaxDialogsFactoryTests.swift
```diff
@@ -353,7 +353,7 @@ extension ContextualDaxDialogsFactoryTests {
         XCTAssertNotNil(host.view)
 
         // THEN
-        waitForExpectations(timeout: 2.0)
+        waitForExpectations(timeout: 10.0)
```
diegoreymendez, line 356: It's not ideal to have to wait this much time, but it's cut short if the test succeeds earlier so it shouldn't have a great impact.

We still should change this test because tests that take this long make no sense IMO, but I also think we should prioritize unblocking Xcode 16 right now.

Locally this test succeeds in around 1.2 seconds.

ayoy (approved):
This change looks reasonable. Thanks for the fix, @diegoreymendez !

PR #3452, opened 2024-10-18
Remove `voiceSearchHelper` from `AppDependencyProvider`

samsymons:
Task/Issue URL: https://app.asana.com/0/414235014887631/1208574561964832/f
Tech Design URL:
CC:

**Description**:

This PR removes `voiceSearchHelper` from `AppDependencyProvider` and updates the test suite to no longer mutate the dependency provider singleton. Tests that use this pattern have been prone to occasional crashes, so I'm simplifying them and working towards removing `makeTestingInstance()` entirely.

**Steps to test this PR**:
1. Check that voice search works as expected

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPhone 14 Pro
* [ ] iPad

**OS Testing**:

* [ ] iOS 15
* [ ] iOS 16
* [ ] iOS 17

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:


DuckDuckGo/SmallOmniBarState.swift
```diff
@@ -30,25 +30,27 @@ struct SmallOmniBarState {
         let showShareButton: Bool = false
         let clearTextOnStart = true
         let allowsTrackersAnimation = false
-        let showSearchLoupe = !AppDependencyProvider.shared.voiceSearchHelper.isSpeechRecognizerAvailable
```
samsymons, line 33: I notice that sometimes this value uses `isSpeechRecognizerAvailable`, and in other places it uses `isVoiceSearchEnabled`. Intentional?

Bunn: Yes, the naming could be better but `isSpeechRecognizerAvailable` is about having the speechRecognizer feature on iOS itself available whereas `isVoiceSearchEnabled` is our feature, if the user turned it on/off.
i.e: `isVoiceSearchEnabled` will always be false if `isSpeechRecognizerAvailable` is false.

In some places (iirc) we have different behaviors depending on system settings or feature settings.

Maybe a better name would be:
* isSystemSpeechRecognizerAvailable
* isSFSpeechRecognizerAvailable

Feel free to change it.

DuckDuckGo/SmallOmniBarState.swift
```diff
@@ -204,24 +216,26 @@ struct SmallOmniBarState {
...
         var onBrowsingStartedState: OmniBarState { return self }
-        var onBrowsingStoppedState: OmniBarState { return HomeEmptyEditingState() }
-        var onEnterPadState: OmniBarState { return LargeOmniBarState.BrowsingTextEditingState() }
+        var onBrowsingStoppedState: OmniBarState { return HomeEmptyEditingState(voiceSearchHelper: voiceSearchHelper) }
+        var onEnterPadState: OmniBarState { return LargeOmniBarState.BrowsingTextEditingState(voiceSearchHelper: voiceSearchHelper) }
         var onEnterPhoneState: OmniBarState { return self }
-        var onReloadState: OmniBarState { return BrowsingTextEditingStartedState() }
+        var onReloadState: OmniBarState { return BrowsingTextEditingStartedState(voiceSearchHelper: voiceSearchHelper) }
+        var showSearchLoupe: Bool { !voiceSearchHelper.isVoiceSearchEnabled }
+        var showVoiceSearch: Bool { !voiceSearchHelper.isVoiceSearchEnabled }
```
Bunn: This is wrong, it should be without the `!`
ex: `var showVoiceSearch: Bool { voiceSearchHelper.isVoiceSearchEnabled }`

Bunn (changes requested):
Overall LGTM, but there's one breaking change there which is why I'm marking as request changes.

DuckDuckGo/SmallOmniBarState.swift
samsymons: Good catch, fixed

DuckDuckGo/SmallOmniBarState.swift
samsymons: That's okay, I'm happy to keep the naming to be honest, just wanted to make sure this was intentional. :+1:

Bunn (approved):

PR #3461, opened 2024-10-21
support third party marketplace URLs

brindy:
Task/Issue URL: https://app.asana.com/0/414709148257752/1207138951730082/f
Tech Design URL:
CC:

**Description**:
Allow marketplace URLs to work.

Note that the URL must be from a 'user' request.  So we can't cancel and re-issue the URL otherwise it doesn't work.  This also means that can't handle the case where it doesn't work unless we want to check the user has an EU apple account, which we don't.

**Steps to test this PR**:
1. With an EU Apple ID, visit altstore.io and install their marketplace. Should cause iOS system prompts and then install successfully.
2. With a non-EU Apple ID, visit altstore.io and try to install their marketplace.  Nothing will happen (unfortunately we are unable to catch this case).

### DIFF
--- a/Core/SchemeHandler.swift
+++ b/Core/SchemeHandler.swift
@@ -51,6 +51,7 @@ public class SchemeHandler {
         case shortcuts
         case shortcutsProduction = "shortcuts-production"
         case workflow
+        case marketKit = "marketplace-kit"
     }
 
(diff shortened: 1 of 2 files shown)


amddg44 (approved):
LGTM!

PR #3519, opened 2024-11-01
Add a debouncer to NavBars animator

afterxleep:
Task/Issue URL: https://app.asana.com/0/1204099484721401/1208671955053442/f
Tech Design URL:
CC:

**Description**:
Fixes a crash related to PDFs and scroll position

This is caused by our navigation bar's hiding animation/visibility triggering a `didScrollEvent`, which causes the app to enter an infinite loop and eventually crash.

The issue does not appear when the navigation bar is at the bottom.  

This adds a debouncer to the method to break the loop while we investigate futher

**Steps to test this PR**:

- Use iOS 18, (preferably on a device) 
- Set the navigation bar to 'top'
- Navigate to https://randomtests.netlify.app/sample3.pdf
- Scroll up and down, zoom, rotate device
- App should not hang or crash

### DIFF
--- a/DuckDuckGo/BarsAnimator.swift
+++ b/DuckDuckGo/BarsAnimator.swift
@@ -87,18 +87,38 @@ class BarsAnimator {
     }
 
     private func transitioningAndScrolling(in scrollView: UIScrollView) {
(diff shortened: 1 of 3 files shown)


brindy (approved):
LGTM on iOS 17

PR #3546, opened 2024-11-05
Switch to free runners for tests that run on Maestro

loremattei:
Task/Issue URL: https://app.asana.com/0/1203301625297703/1208705705667488/f
Tech Design URL:
CC:

**Description**:

**Steps to test this PR**:
1. Ideally, check [this run](https://github.com/duckduckgo/iOS/actions/runs/11690377938) is green. Realistically, check that it's not worse than usual in any way that can be related to this change.  

---
###### Internal references:

### DIFF
--- a/.github/workflows/end-to-end.yml
+++ b/.github/workflows/end-to-end.yml
@@ -68,7 +68,7 @@ jobs:
   end-to-end-tests:
     name: End to end Tests
     needs: build-end-to-end-tests
-    runs-on: macos-14-xlarge
+    runs-on: macos-14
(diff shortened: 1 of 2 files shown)


ayoy (approved):
LGTM, it's reasonable to use the standard runner especially given that it doesn't compile the app. The provided test run is in line with previous runs so there's no clear indication of the influence of reducing runner's compute power on tests stability. Worst case, if we notice tests getting even more unstable, reverting to xlarge runners is simple. Thanks @loremattei!

PR #3562, opened 2024-11-11
fix showing current tab in suggestions

brindy:
Task/Issue URL: https://app.asana.com/0/392891325557410/1208712038329496/f
Tech Design URL:
CC:

**Description**:
Fixes bug showing open tab for current tab when typing in suggestions.

**Steps to test this PR**:
1.  Go to https://news.ycombinator.com/ and open a bunch of pages in new tabs
2. Type in to the address for one of the pages - should show as an open tab.  Select it and switch to tab.
3. Type into the address bar for the same page - should only show as history
4. Validate that the same page open in multiple tabs only shows as a single 'switch to tab' suggestion and does not show at all when on a page with the same URL


alessandroboron (approved):
Works as expected! Perhaps we could add a test for `AutocompleteViewController` -> `candidateOpenTabs` to avoid regressions?

brindy:
Yeah you're right. I'm gonna extract that extension into its own testable class. Bear with me.

DuckDuckGoTests/MockFeatureFlagger.swift
```diff
@@ -21,8 +21,8 @@ import BrowserServicesKit
 import Core
 
 final class MockFeatureFlagger: FeatureFlagger {
+
     var enabledFeatureFlags: [FeatureFlag] = []
-    var enabledFeatureFlag: FeatureFlag?
```
brindy, line 25: is unused by tests and not part of the protocol

DuckDuckGo/AutocompleteSuggestionsDataSource.swift
```diff
@@ -0,0 +1,100 @@
...
+    init(historyManager: HistoryManaging, bookmarksDatabase: CoreDataDatabase, featureFlagger: FeatureFlagger, tabsModel: TabsModel, performSuggestionsRequest: @escaping SuggestionsRequest) {
+        self.historyManager = historyManager
+        self.bookmarksDatabase = bookmarksDatabase
+        self.featureFlagger = featureFlagger
+        self.tabsModel = tabsModel
+        self.performSuggestionsRequest = performSuggestionsRequest
+    }
+
+    func history(for suggestionLoading: Suggestions.SuggestionLoading) -> [HistorySuggestion] {
+        return historyCoordinator.history ?? []
```
alessandroboron, line 71: nip: can remove unnecessary `return` on line. Same applies to below methods:

- `func bookmarks(for suggestionLoading: Suggestions.SuggestionLoading) -> [Suggestions.Bookmark]`

-  `func internalPages(for suggestionLoading: Suggestions.SuggestionLoading) -> [Suggestions.InternalPage] `

alessandroboron (approved):
I tested again and works! Just left a “nip” comment. Thanks for addressing the tests.

DuckDuckGo/AutocompleteSuggestionsDataSource.swift
brindy: I'm gonna leave it.  My thinking is that functions should `return` something (unless they are view builders). I would like to think I'm consistent about that ... but who knows. :) Anyway, since it's just a nit I'm gonna go with what I've got.

PR #3595, opened 2024-11-19
Re-prompt for crash reporting

jdjackson:
Task/Issue URL: https://app.asana.com/0/1208592102886666/1208630312625695/f
Tech Design URL: https://app.asana.com/0/1208592102886666/1208660326715650/f
CC: @loremattei 

**Description**:
- Introduce a feature flag for use with this project overall, currently set to internal only
- Introduce a new UserDefaults value (```crashCollectionShouldRevertOptedInStatusTrigger```) that can be used to force the client to revert the existing ```crashCollectionOptInStatus```  to .unknown, only if the value is currently .optedIn.  This new trigger is implemented as an Int, so that we can ensure the user’s Opt-In state will be reverted only once after the feature is turned on.  In the future if we need this functionality again, we’d simply bump the garget value and the opt-in state would be reset again.

See asana task for specific requirements, and the parent project for more context if necessary.

**Steps to test this PR**:
**Preconditions**
- Start with a fresh install of the iOS browser _**without the changes in this PR**_.
- Crash reporting and submitting works only on a physical device, so testing cannot be done in the simulator

**Testing Steps**
1. Create a crash log: Launch the app, and navigate to “…” -> Settings -> All debug options -> Crash (any one of the 4 crash types)
2. Re-launch the app and see that you are prompted to submit a crash report.  Select “Always Send Crash Reports”.
3. Perform step 1 again to produce another crash.  Do not launch the app again yet.
4. _**Install the new build of the app with this PR’s changes included**_
5. Run the app, and you should be prompted again to submit crash reports (or not), same screen as from step 2.
6. Cause another crash via the settings menu
7. Run the app _again_ and ensure that you are NOT prompted yet another time (your opt-in choice with the new version should now be respected)

* Note: I don’t believe there is any automation around existing crash reporting (I couldn’t find any).  If it exists, or if you think there is a straight foward way to add some, please point me in the right direction :)

**Copy Testing**:

* [N/A] Use of correct apostrophes in new copy, ie `’` rather than `’`

**Orientation Testing**:

Note: I have tested on the devices and configurations I have available at home currently.  These changes are not UI related, nor are they likely to depend on particular OS version functionality, so exhaustive coverage is likely not critical.  If you’re able to test with a different configuration than I used that would be helpful :)

* [X] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPhone 14 Pro
* [X] iPhone 15 Pro
* [ ] iPad

**OS Testing**:

* [ ] iOS 15
* [ ] iOS 16
* [ ] iOS 17
* [X] iOS 18

**Theme Testing**:

* [ ] Light theme
* [X] Dark theme

### DIFF
--- a/DuckDuckGo/CrashCollectionOnboarding.swift
+++ b/DuckDuckGo/CrashCollectionOnboarding.swift
@@ -20,6 +20,7 @@
 import Combine
 import Foundation
 import SwiftUI
+import BrowserServicesKit
 
 enum CrashCollectionOptInStatus: String {
     case undetermined, optedIn, optedOut
@@ -35,10 +36,14 @@ final class CrashCollectionOnboardingViewController: UIHostingController<CrashCo
 }
 
 final class CrashCollectionOnboarding: NSObject {
+    
+    private var featureFlagger: FeatureFlagger
 
-    init(appSettings: AppSettings) {
+    init(appSettings: AppSettings,
+         featureFlagger: FeatureFlagger = AppDependencyProvider.shared.featureFlagger) {
         self.appSettings = appSettings
         self.viewModel = CrashCollectionOnboardingViewModel(appSettings: appSettings)
+        self.featureFlagger = featureFlagger
         super.init()
     }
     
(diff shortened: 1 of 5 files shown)


jdjackson:
@ayoy Assigning to you since I see your name on most of the relevant code I’m working in, but do let me know if you prefer someone else review.  Thanks!

DuckDuckGo/CrashCollectionOnboarding.swift
```diff
@@ -35,16 +36,34 @@ final class CrashCollectionOnboardingViewController: UIHostingController<CrashCo
...
+         featureFlagger: FeatureFlagger = AppDependencyProvider.shared.featureFlagger) {
         self.appSettings = appSettings
         self.viewModel = CrashCollectionOnboardingViewModel(appSettings: appSettings)
+        self.featureFlagger = featureFlagger
         super.init()
     }
+    
+    // If the user's crash report opt in status should be reset to unknown upon next release,
+    // increment this value by 1
+    private let crashCollectionShouldRevertOptedInStatusTriggerTargetValue: Int = 1
```
ayoy, line 52: Nice future-proofing 💪

DuckDuckGo/CrashCollectionOnboarding.swift
```diff
@@ -35,16 +36,34 @@ final class CrashCollectionOnboardingViewController: UIHostingController<CrashCo
...
 
     @MainActor
     func presentOnboardingIfNeeded(for payloads: [Data], from viewController: UIViewController, sendReport: @escaping () -> Void) {
         let isCurrentlyPresenting = viewController.presentedViewController != nil
+        
+        if featureFlagger.isFeatureOn(.crashReportOptInStatusResetting) {
+            if appSettings.crashCollectionOptInStatus == .optedIn &&
+                appSettings.crashCollectionShouldRevertOptedInStatusTrigger < crashCollectionShouldRevertOptedInStatusTriggerTargetValue {
+                debugPrint("Resettting crash report opt in status from \(appSettings.crashCollectionOptInStatus) to .undetermined")
```
ayoy: We try to avoid calls to `print` and `debugPrint` in the code. You could use `Logger.general.debug` instead.

ayoy (approved):
LGTM, I've tested with a physical device and everything worked as expected, got a new prompt after installing the app from the branch. Leaving just one minor comment about debug console logs, and otherwise it's good to go. Thanks @jdjackson!

DuckDuckGo/CrashCollectionOnboarding.swift
jdjackson: Oops, that’s just an oversight on my part from testing.  I’m going to remove the print entirely, but TY for the pointer to our logging system :)

PR #3635, opened 2024-11-27
AI Chat browsing menu

Bunn:
Task/Issue URL: https://app.asana.com/0/1204167627774280/1208794395441049/f

**Description**:
Add AI Chat entry point in the browsing menu

**Steps to test this PR**:
1. Open the browsing menu (the icon at the bottom left).
2. Select "Chat."
3. It should open AI Chat in a full-screen modal. Check if we're correctly sending the "browsingMenuAIChat" pixel
4. Start a conversation with AI Chat.
5. Close the UI, then open it again within 10 minutes. Check if we're correctly sending the "before10min" pixel
6. Validate that your chat is still there.
7. Close the UI, then open it again after 10 minutes. Check if we're correctly sending the "after10min" pixel
8. Validate that your chat is no longer there.
9. Check that the print button in the list is still working.
10. In FeatureFlag.swift, return disabled in aiChatBrowsingToolbarShortcut.
11. Open the browsing menu and validate that the Chat button is not there, and the print button is at the top header and not present in the list.

----
1. Log on a website or set a cookie somehow
2. Open AI Chat
3. Start a conversation
4. Close AI Chat
5. Before 10min, use the fire button
6. Validate that the website doesn't have a cookie or you're logged out
7. Open AI Chat, check if the webview state is still intact (conversation still visible)
----
1. Open AI Chat
2. Inside the page, tap in the top right info button, then tap "Help Pages"
3. Check if the AI Chat page is dismissed, and we open a new tap with the help pages.

----
Testing the new settings page:
1. Set yourself as internal user
2. Check if the feature is enabled by default (it should)
3. Check if you can see the AI Chat button in the browsing menu
4. Open settings -> AI Chat, disable AI Chat
5. Check if the AI Chat button is gone and print is added back to it's place


DuckDuckGo/BrowsingMenu/BrowsingMenuViewController.swift
```diff
@@ -249,8 +249,9 @@ extension BrowsingMenuViewController: UITableViewDelegate {
         
         switch menuEntries[indexPath.row] {
         case .regular(_, _, _, _, let action):
-            dismiss(animated: true)
-            action()
+            dismiss(animated: true) {
```
Bunn, line 252: Having the action called without waiting for the dismissal was causing problems with actions that depended on the view stack, like presenting a ViewController

LocalPackages/AIChat/Sources/AIChat/TimerPixelHandler.swift
```diff
@@ -0,0 +1,56 @@
...
+//
+//  Unless required by applicable law or agreed to in writing, software
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+
+final class TimerPixelHandler {
```
Bunn, line 21: This is temporary. Will be removed once we ship the FE feature

LocalPackages/AIChat/Sources/AIChat/UserText.swift
```diff
@@ -0,0 +1,24 @@
...
+//  distributed under the License is distributed on an "AS IS" BASIS,
+//  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+//  See the License for the specific language governing permissions and
+//  limitations under the License.
+//
+
+import Foundation
+
+public struct UserText {
+    public static let aiChatTitle = NSLocalizedString("aichat.title", value: "DuckDuckGo AI Chat", comment: "Title for DuckDuckGo AI Chat. Should not be translated")
```
samsymons, line 23: We also have `NotLocalizedString` in BSK which could be used here to avoid the string getting accidentally translated, but tbh I've never used it and not sure if it just swaps in seamlessly. 😄

LocalPackages/AIChat/Sources/AIChat/TimerPixelHandler.swift
samsymons: Is this something that should be put behind a subfeature flag and disabled once the FE support ships, or are we happy enough to just remove this code later?

LocalPackages/AIChat/Sources/AIChat/Public API/AIChatViewController.swift
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    public override func didReceiveMemoryWarning() {
+        super.didReceiveMemoryWarning()
+
+        if viewIfLoaded?.window == nil {
+            chatModel.cancelTimer()
+            removeWebViewController()
+        }
+    }
```
samsymons, line 94: Nice 👏

LocalPackages/AIChat/Package.swift
```diff
@@ -0,0 +1,20 @@
+// swift-tools-version: 6.0
```
samsymons, line 1: Swift 6 👀

LocalPackages/AIChat/Sources/AIChat/Public API/AIChatViewController.swift
```diff
@@ -0,0 +1,204 @@
...
+
+        navigationItem.leftBarButtonItem = UIBarButtonItem(customView: stackView)
+
+        let closeButton = UIBarButtonItem(
+            image: UIImage(named: "Close-24"),
+            style: .plain,
+            target: self,
+            action: #selector(closeAIChat)
+        )
+        closeButton.tintColor = .white
```
samsymons: I think this is missing an accessibility identifier, so the asset name will get used as the label.

DuckDuckGo/TabViewControllerBrowsingMenuExtension.swift
```diff
@@ -28,25 +28,32 @@ import PrivacyDashboard
...
 extension TabViewController {
 
-    func buildBrowsingMenuHeaderContent() -> [BrowsingMenuEntry] {
+    private var shouldShowAIChatInMenuHeader: Bool {
+        let settings = AIChatSettings(privacyConfigurationManager: ContentBlocking.shared.privacyConfigurationManager,
+                                      internalUserDecider: AppDependencyProvider.shared.internalUserDecider)
+        return settings.isAIChatBrowsingToolbarShortcutFeatureEnabled
+    }
+
+    private var shouldShowPrintButtonInBrowsingMenu: Bool { shouldShowAIChatInMenuHeader }
```
samsymons: Silly question: is this intended? This'll mean the Print button is hidden when the AI chat option is disabled 😅

DuckDuckGo/SettingsAIChatView.swift
```diff
@@ -0,0 +1,54 @@
...
+
+struct SettingsAIChatView: View {
+    @EnvironmentObject var viewModel: SettingsViewModel
+
+    var body: some View {
+        List {
+
+            VStack(alignment: .center) {
+                Image("SettingsDuckPlayerHero")
+                    .padding(.top, -20) // Change this to AI Chat image
```
samsymons: Looks like a stray TODO here, I assume we're still waiting on an asset for this? (EDIT: saw the comment on the ship review about it, ignore this)

LocalPackages/AIChat/Sources/AIChat/Public API/AIChatViewController.swift
```diff
@@ -0,0 +1,204 @@
...
+            .sink { [weak self] in
+                self?.webViewController?.reload()
+                self?.timerPixelHandler.markCleanup()
+            }
+    }
+
+    @objc private func closeAIChat() {
+        chatModel.startCleanupTimer()
+        dismiss(animated: true)
+    }
```
samsymons, line 197: One note that is probably not a problem: when this VC is dismissed, it triggers `viewDidAppear` in `MainViewController`. This causes various things to happen, like `m.debug.app-did-show-ui-time-more` firing - I'd want to check with @jaceklyp whether this pollutes the data coming from that pixel, or whether this is not an issue.

samsymons:
Added a few comments, only a couple are actionable though. Overall this looks solid, I've spent a bunch of time bashing on it on both iPhone and iPad, trying to break the lifecycle of it in various ways, no issues found. Once the comments are resolved where needed, I think we can approve this.

LocalPackages/AIChat/Sources/AIChat/TimerPixelHandler.swift
Bunn: We're good in just removing the code later, once the new feature is released we'll also pair with a native release, so it should be good

LocalPackages/AIChat/Sources/AIChat/UserText.swift
Bunn: Yep, I was using it but I saw that AI Chat is in our translation memory to no translated it. So decided against copying the NotLocalizedString to the package for now.

DuckDuckGo/TabViewControllerBrowsingMenuExtension.swift
Bunn: It's intended yes. The name clearly needs work then haha. Basically, if AI chat is on, we MOVE the print button from the Browswing menu header to the browsing menu list. So if this flag is ON, it means the print button should be in the list instead of the header.

I'll rename this from `shouldShowPrintButtonInBrowsingMenu` to `shouldShowPrintButtonInBrowsingMenuList`, wdyt?

jaceklyp (approved):
thank you!

LocalPackages/AIChat/Sources/AIChat/Public API/AIChatViewController.swift
Bunn: I talked with @jaceklyp and we decided to prevent this pixel from being fired more than once since it should have been like this in the first place: https://github.com/duckduckgo/iOS/pull/3635/commits/c075d3d07486739343c2f0190d63d68dc14b5fd9

LocalPackages/AIChat/Package.swift
Bunn: Do you see any issues with it? Thought to leave it like that just to be more future-proof, but no strong opinions

DuckDuckGo/TabViewControllerBrowsingMenuExtension.swift
samsymons: Ohh, I get it now - that naming change sounds fine to me, yeah. Thanks for clarifying!

LocalPackages/AIChat/Package.swift
samsymons: Nope, actually I'm happy about it since I want us to begin moving to Swift 6 incrementally 🎉

samsymons (approved):
:shipit:

PR #3654, opened 2024-11-29
fix suggestion scrolling

brindy:
Task/Issue URL: https://app.asana.com/0/392891325557410/1208804646679184/f
Tech Design URL:
CC:

**Description**:
Improve suggestion scrolling in landscape and height calculation on iPad.

See Asana for a sync code for favorites.

**Steps to test this PR**:
1. On phone check suggestions scroll as expected in portrait and landscape.  
2. In landscape suggestions should scroll above the bottom bar when bottom bar is enabled.  Check both bar positions.
4. On iPad sets a more reasonable size for the favorites in overlay mode and flashes the scrollbar
5. On iPad also calculates height of suggestions more correctly to fit the view

### DIFF
--- a/DuckDuckGo/AutocompleteViewController.swift
+++ b/DuckDuckGo/AutocompleteViewController.swift
@@ -204,22 +204,38 @@ class AutocompleteViewController: UIHostingController<AutocompleteView> {
         guard let lastResults else { return }
 
         let messageHeight = model.isMessageVisible ? 196 : 0
-        let cellHeight = 44
-        let sectionPadding = 10
+        let sectionPadding = 12
         let controllerPadding = 20
 
         let height =
-            (lastResults.topHits.count * cellHeight) +
+            sectionHeight(lastResults.topHits) +
             (lastResults.topHits.isEmpty ? 0 : sectionPadding) +
-            (lastResults.duckduckgoSuggestions.count * cellHeight) +
+            sectionHeight(lastResults.duckduckgoSuggestions) +
             (lastResults.duckduckgoSuggestions.isEmpty ? 0 : sectionPadding) +
-            (lastResults.localSuggestions.count * cellHeight) +
+            sectionHeight(lastResults.localSuggestions) +
             (lastResults.localSuggestions.isEmpty ? 0 : sectionPadding) +
             messageHeight +
             controllerPadding
 
         presentationDelegate?
             .autocompleteDidChangeContentHeight(height: CGFloat(height))
+    }
+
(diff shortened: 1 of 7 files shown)


DuckDuckGo/SuggestionTrayViewController.swift
```diff
@@ -326,6 +332,6 @@ extension SuggestionTrayViewController {
 
 private extension SuggestionTrayViewController {
     enum Constant {
-        static let suggestionTrayInitialHeight = 312.0 // ie 52 * 6
+        static let suggestionTrayInitialHeight = 380.0 // ie 52 * 6
```
dus7, line 335: The comment here says 52 * 6 but it's == 312 :)

dus7: The comment here says 52 * 6 (312) but it's == 380 :)

DuckDuckGo/AutocompleteViewController.swift
```diff
@@ -222,6 +221,23 @@ class AutocompleteViewController: UIHostingController<AutocompleteView> {
             .autocompleteDidChangeContentHeight(height: CGFloat(height))
     }
 
+    func sectionHeight(_ suggestions: [Suggestion]) -> Int {
```
dus7, line 224: There's something weird going on with spacing with the autocomplete overlay now when I have bar set to top:

DuckDuckGo/FavoritesHomeViewSectionRenderer.swift
```diff
@@ -89,20 +89,12 @@ class FavoritesHomeViewSectionRenderer {
     }
 
     func sectionMargin(in collectionView: UICollectionView) -> CGFloat {
-        if controller is FavoritesOverlay {
-            return Constants.largeModeMargin
-        }
-        
-        let margin: CGFloat
         if isPad {
```
dus7: I'm afraid this condition is insufficient and we also need to check for a size class in case we're in a Split View:

The same goes for favorites in Bookmarks:

dus7 (changes requested):
The overlay sizing needs fixing, as well as a specific case of favorites on iPad. Fore other cases (different sizes of iPhone, iPad) Favorites worked properly for me.

brindy:
@dus7 I fixed the weird cut off thing, but I decided to roll back the logic around favorites. I think in the long term we should get NTP favorites code running there tbh.

dus7 (approved):
Works good! Tested on a few different iPhones, iPad, iPad with Split View.

PR #3656, opened 2024-12-01
Add `install` pixel

samsymons:
Task/Issue URL: https://app.asana.com/0/1199333091098016/1208873240363373/f
Tech Design URL:
CC:

**Description**:

This PR adds an `m.install.ios.*` pixel that is sent when calling the `exti` URL.

**Steps to test this PR**:
1. Reset your simulator and run the app, look for `m.install` in the console, with the parameters `reinstall` and `locale`, which match the current state of your device - you should expect to see `reinstall=false` if you reset your simulator, then you can delete the app and reinstall it to get `reinstall=true`
2. Now kill and relaunch the app and check that you do not see `m.install` again

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPhone 14 Pro
* [ ] iPad

**OS Testing**:

* [ ] iOS 15
* [ ] iOS 16
* [ ] iOS 17

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:


DuckDuckGoTests/StatisticsLoaderTests.swift
```diff
@@ -270,6 +276,19 @@ class StatisticsLoaderTests: XCTestCase {
         waitForExpectations(timeout: 5, handler: nil)
     }
 
+    func testWhenInstallStatisticsRequestedThenInstallPixelIsFired() {
+        loadSuccessfulExiStub()
+
+        let testExpectation = expectation(description: "refresh complete")
+        testee.refreshAppRetentionAtb {
+            Thread.sleep(forTimeInterval: .seconds(0.1))
+            testExpectation.fulfill()
```
samsymons, line 285: This small delay isn't ideal; it's in place since the `refreshAppRetentionAtb` callback is waiting for the `exti` call, **not** the pixel call, so there is a chance that this could complete before the pixel call has had time to do its thing.

It would be better to wait for the pixel firing mock to have the right value, with a ~1 second timeout. I'll look at updating this.

Bunn (commented):
@samsymons I'm afraid there's an issue here, or I'm missing something when testing this. 
No matter what I do, reset simulator, use a brand new simulator, I always see the pixel 
`Pixel fired m.install ["locale": "en-US", "reinstall": "true"]` being fired.

After some investigation, I found out that the [installCompletedWithATB](https://github.com/duckduckgo/iOS/blob/2d9ec00309ad2e1a4adb19680f9faea76d63a6ef/Core/ReturnUserMeasurement.swift#L51C10-L51C33) is called before the [fireInstallPixel](https://github.com/duckduckgo/iOS/blob/d8afd9e1e1f779d6ccd6aeee9b2277bbcb27c128/Core/StatisticsLoader.swift#L109C18-L109C34) which causes the [ReturnUserMeasurement](https://github.com/duckduckgo/iOS/blob/d8afd9e1e1f779d6ccd6aeee9b2277bbcb27c128/Core/ReturnUserMeasurement.swift#L107) to return true

Bunn (approved):
Working as expected now, I was able to trigger both pixels, only once per install

Pixel fired m.install ["locale": "en-US", "reinstall": "true"]
Pixel fired m.install ["locale": "en-US", "reinstall": "false"]

PR #1116, opened 2022-04-08
Fix Russian fireproofing translations

THISISDINOSAUR:
Task/Issue URL: https://app.asana.com/0/0/1202101260086320/f
Tech Design URL:
CC:

**Description**:
Updates various Russian fireproofing translations to make more sense, see asana task for details

**Steps to test this PR**:
1. Go to settings, and check the fire button animations settings entry, and the title on the page are as specified in the asana task
2. Go to fireproofed websites in settings and do the same. Turn on fireproofing. 
3. Log in to a website so you get the fireproofing dialogue. Check the translations there are correct. Press confirm, and check the toast is correct.

**Copy Testing**:

* [ ] Use of correct apostrophes in new copy, ie `’` rather than `'`

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPad

**OS Testing**:

* [ ] iOS 13
* [ ] iOS 14
* [ ] iOS 15

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:
[Software Engineering Expectations](https://app.asana.com/0/59792373528535/199064865822552)
[Technical Design Template](https://app.asana.com/0/59792373528535/184709971311943)
**When ready for review, remember to post the PR in MM**

PR #3692, opened 2024-12-06
Crash report cohort ID support for iOS

jdjackson:
Task/Issue URL: https://app.asana.com/0/1208592102886666/1208759541597499/f
Tech Design URL: https://app.asana.com/0/1208592102886666/1208660326715650/f

**Description**:
DO NOT MERGE - this is a draft for input, not ready to go live yet.

iOS client support for CRCID send/receive (primarily supported in BSK, with changes under review in [BSK #1116](https://github.com/duckduckgo/BrowserServicesKit/pull/1116)).  This is pretty straightforward, just conforming to CrashCollection’s new init signature, and clearing CRCIDs when the user opts out of crash reporting.  BSK handles everything else.

**Steps to test this PR**:
Note: Must be tested on a physical device, as the simulator does not produce crash logs (and thus doesn’t find and upload them either).

To cause and report a crash:
1. Launch the app and force a crash, which can be done from Settings → All Debug Options → Crash (fatal error) or similar.  Note that Crash (CPU/Memory) does not appear to produce a crash log, and thus won’t trigger crash uploading.
2. Launch the app again (easiest with a debugger)
  1. For the first crash of an app install: You will be prompted to opt in or out of crash reporting when the app is launched.  Opt in and watch logs for “crcid” and you should see logs from CrashReportSender:56, and CrashCollection:95-109.
  2. On subsequent crashes, when opted in, you should see statements confirming the received crcid was sent, and that the server returned either the same matching one, or a new one (in which case the new one should be stored and used on subsequent crash reports)

To test clearing of the crcid when opting out:
1. Navigate to Settings → About and switch “Send Crash Reports” off, then back on again (this step should clear the crcid)
2. Follow steps from “To cause and report a crash” above, and confirm that the crash is submitted without an initial crcid, and that the server assigns one and it is stored (causing and uploading a second crash should confirm this new value is used on send).

**Copy Testing**: N/A

**Orientation Testing**: N/A

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPhone 14 Pro
* [X] iPhone 15 Pro
* [ ] iPad

**OS Testing**:

* [ ] iOS 15
* [ ] iOS 16
* [ ] iOS 17
* [X] iOS 18

**Theme Testing**: N/A

* [ ] Light theme
* [ ] Dark theme

—
###### Internal references:


Core/PixelEvent.swift
```diff
@@ -505,6 +505,9 @@ extension Pixel {
         case dbCrashDetected
         case crashOnCrashHandlersSetUp
 
+        case crashReportCRCIDMissing // crashreporting_crcid-missing
+        case crashReportingSubmissionFailed // crashreporting_submission-failed
```
samsymons: We can omit the comments here for consistency, since the value below will contain the raw name.

jdjackson: Thanks for catching this, it some leftover note taking from when I was working it all out :)  I’ve removed from both iOS and macOS clients and will push shortly.

Core/PixelEvent.swift
```diff
@@ -1367,6 +1370,8 @@ extension Pixel.Event {
             // MARK: debug pixels
             
         case .dbCrashDetected: return "m_d_crash"
+        case .crashReportCRCIDMissing: return "crashreporting_crcid-missing"
+        case .crashReportingSubmissionFailed: return "crashreporting_submission-failed"
```
samsymons: These will need to have the `m_` prefix at the start, otherwise they won't get processed correctly by the pixel endpoint. Something like `m_crashreporting_crcid-missing` is fine. This isn't needed on macOS since the prefixing is handled automatically, but iOS is a bit behind 🙂

samsymons (approved):

PR #3707, opened 2024-12-10
Malicious site protection navigation detection

alessandroboron:
Task/Issue URL: https://app.asana.com/0/1206329551987282/1207151848931030
Tech Design URL: https://app.asana.com/0/1206329551987282/1207273224076495/f
Cc: @not-a-rootkit 

**Description**:

This PR adds the navigation logic for detecting a malicious site and navigating to a special error page if the site is malicious.

The original idea in the tech design was to intercept the Request in `decidePolicyForNavigationAction` and check whether a site is malicious cancelling the request accordingly.

We noticed that the above approach increases the page load time of websites due to the logic check.

I opted for an approach where in `decidePolicyForNavigationAction` we start the detection task in parallel without waiting and in `decidePolicyForNavigationResponse` we evaluate the task’s result.

Another approach I thought of was to perform the logic in the background in `didStartProvisionalNavigation`. The problem with this approach is that is called only for navigation that starts from the main frame so it would not be possible to intercept malicious iFrame URLs.

**NOTE** 
BSK Threat detection is currently mocked. Its integration and tests will follow in an upcoming PR.

**Steps to test this PR**:

**Scenario 1 - Phishing Website should show special error page**
1. Ensure that http://privacy-test-pages.site/security/badware/phishing.html shows the phishing special error page with the copy “This website may be impersonating a legitimate site...".

**Scenario 2 - Malware Website should show special error page**
1. Ensure that http://privacy-test-pages.site/security/badware/malware.html shows the malware special error page with the copy “DuckDuckGo blocked this page because it may be distributing malware...".

**Scenario 3 - Allow visit a malicious website should not show the error again if navigating back to the site**
1. Visit one of the malicious sites above.
2. Tap the “Advanced” button
3. Tap the “Accept Risk and Visit Site” link
4. Wait for malicious site to load.
5. Refresh page. 
6. Ensure the website load without showing the error again.
7. Close tab.
8. Paste the same URL again.
9. Ensure the special error page is shown again.

**Scenario 4 - Leave a malicious site should close the Tab**
1. Visit one of the malicious sites above.
2. Tap the “Leave Site” button.
3. Ensure that the tab is closed.

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPhone 14 Pro
* [ ] iPad

**OS Testing**:

* [ ] iOS 15
* [ ] iOS 16
* [ ] iOS 17

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:


DuckDuckGo/SpecialErrorPage/SpecialErrorPageInterfaces/SpecialErrorPageActionHandler.swift
```diff
@@ -26,11 +27,14 @@ import Foundation
 /// advanced information related to the error.
 protocol SpecialErrorPageActionHandler {
     /// Handles the action of navigating to the site associated with the error page
-    func visitSite()
+    @MainActor
+    func visitSite(url: URL, errorData: SpecialErrorData)
```
alessandroboron, line 34: I’m a bit on the fence with this. 
The idea was to have a generic protocol that all the sub-special error handlers should conform to. 
For `MaliciousSiteProtectionNavigationHandler` I need to pass the URL and SpecialErrorData but I don’t need them for the SSL handler. So I feel I’m violating the Interface Segregation principle. what do you think?

DuckDuckGo/SpecialErrorPage/SpecialErrorPageNavigationHandler+MaliciousSite.swift
```diff
@@ -56,10 +79,53 @@ final class MaliciousSiteProtectionNavigationHandler {
...
+            guard let threatKind = await self.maliciousSiteProtectionManager.evaluate(url) else {
+                return .navigationNotHandled
+            }
+
+            if await navigationAction.isTargetingMainFrame {
+                let errorData = SpecialErrorData.maliciousSite(kind: threatKind, url: url)
+                let response = MaliciousSiteDetectionNavigationResponse(navigationAction: navigationAction, errorData: errorData)
+                return .navigationHandled(.mainFrame(response))
+            } else {
+                // Extract the URL of the source frame (the iframe) that initiated the navigation action
```
alessandroboron, line 107: I will look more into iFrame logic when I integrate with BSK

DuckDuckGo/SpecialErrorPage/SpecialErrorPageNavigationHandler+MaliciousSite.swift
```diff
@@ -81,3 +150,36 @@ extension MaliciousSiteProtectionNavigationHandler: SpecialErrorPageActionHandle
...
 
 }
+
+// MARK: - Private
+
+private extension MaliciousSiteProtectionNavigationHandler {
+
+    @MainActor
+    func handleMaliciousExemptions(for navigationType: WKNavigationType, url: URL) {
+        // TODO: check storing redirects
```
alessandroboron, line 157: Will address this TODO in upcoming PR

DuckDuckGo/SpecialErrorPage/SpecialErrorPageNavigationHandler+SSL.swift
```diff
@@ -90,7 +94,7 @@ extension SSLErrorPageNavigationHandler: SpecialErrorPageActionHandler {
         Pixel.fire(pixel: .certificateWarningLeaveClicked)
     }
 
-    func visitSite() {
+    func visitSite(url: URL, errorData: SpecialErrorData) {
```
alessandroboron: See comment https://github.com/duckduckgo/iOS/pull/3707/files#r1878050527

DuckDuckGo/MaliciousSiteProtection/MaliciousSiteProtectionManager.swift
```diff
@@ -18,38 +18,20 @@
...
-public extension ThreatKind {
 
-    var errorPageType: SpecialErrorKind {
-        switch self {
-        case .malware: .phishing // WIP in BSK
-        case .phishing: .phishing
+        switch url.absoluteString {
+        case "http://privacy-test-pages.site/security/badware/phishing.html":
+            return .phishing
+        case "http://privacy-test-pages.site/security/badware/malware.html":
```
jaceklyp, line 31: should it be http vs https?
also, we should be keeping all urls inside AppURLs

DuckDuckGo/SpecialErrorPage/SpecialErrorPageInterfaces/SpecialErrorPageActionHandler.swift
jaceklyp: could you perhaps create two funcs visitSite, one without params and one with params? Different handlers would be implementing only one function leaving other implementation empty? It's not ideal but the simplest

DuckDuckGo/SpecialErrorPage/SpecialErrorPageInterfaces/WebViewNavigationHandling.swift
```diff
@@ -41,15 +42,24 @@ protocol WebViewNavigationHandling: AnyObject {
     /// - Parameters:
     ///   - navigationAction: Details about the action that triggered the navigation request.
     ///   - webView: The web view from which the navigation request began.
-    /// - Returns: A Boolean value that indicates whether the navigation action was handled.
-    func handleSpecialErrorNavigation(navigationAction: WKNavigationAction, webView: WKWebView) async -> Bool
+    @MainActor
+    func handleDecidePolicyFor(navigationAction: WKNavigationAction, webView: WKWebView)
```
jaceklyp: I think it should be
func handleDecidePolicy(for navigationAction: WKNavigationAction, webView: WKWebView)

DuckDuckGo/SpecialErrorPage/SpecialErrorPageInterfaces/WebViewNavigationHandling.swift
```diff
@@ -41,15 +42,24 @@ protocol WebViewNavigationHandling: AnyObject {
...
+
+    /// Decides whether to to navigate to new content after the response to the navigation request is known or cancel the navigation and show a special error page based on the specified action information.
+    /// - Parameters:
+    ///   - navigationResponse: Descriptive information about the navigation response.
+    ///   - webView: The web view from which the navigation request began.
+    /// - Returns: A Boolean value that indicates whether to cancel or allow the navigation.
+    @MainActor
+    func handleDecidePolicyfor(navigationResponse: WKNavigationResponse, webView: WKWebView) async -> Bool
```
jaceklyp: typo, "for" is lowercased, but I'd refer to my previous comment ^

DuckDuckGo/SpecialErrorPage/SpecialErrorPageNavigationHandler+MaliciousSite.swift
```diff
@@ -22,26 +22,49 @@ import BrowserServicesKit
...
 protocol MaliciousSiteProtectionNavigationHandling: AnyObject {
-    /// Decides whether to cancel navigation to prevent opening the YouTube app from the web view.
+    /// Creates a task for detecting malicious sites based on the provided navigation action.
+    ///
+    /// - Parameters:
+    ///   - navigationAction: The `WKNavigationAction` object that contains information about
+    ///     the navigation event.
+    ///   - webView: The web view from which the navigation request began.
+    @MainActor
+    func createMaliciousSiteDetectionTask(for navigationAction: WKNavigationAction, webView: WKWebView)
```
jaceklyp: I'd go with makeMalicious*

DuckDuckGo/SpecialErrorPage/SpecialErrorPageNavigationHandler+MaliciousSite.swift
```diff
@@ -56,10 +79,53 @@ final class MaliciousSiteProtectionNavigationHandler {
...
+            return
+        }
+
+        handleMaliciousExemptions(for: navigationAction.navigationType, url: url)
+
+        guard !shouldBypassMaliciousSiteProtection(for: url) else {
+            return
+        }
+
+        let threatDetectionTask: Task<MaliciousSiteProtectionNavigationResult, Never> = Task.detached { [weak self] in
```
jaceklyp: why is the Task detached?

DuckDuckGo/SpecialErrorPage/SpecialErrorPageNavigationHandler+MaliciousSite.swift
```diff
@@ -56,10 +79,53 @@ final class MaliciousSiteProtectionNavigationHandler {
...
-        return .navigationNotHandled
+    func createMaliciousSiteDetectionTask(for navigationAction: WKNavigationAction, webView: WKWebView) {
+
+        guard let url = navigationAction.request.url else {
+            return
+        }
+
+        if let aboutBlankURL = URL(string: "about:blank"), url == aboutBlankURL {
+            return
+        }
```
jaceklyp: can we put it under one guard, like so
guard let url = navigationAction.request.url, url != URL(string: "about:blank") else { return }

DuckDuckGo/SpecialErrorPage/SpecialErrorPageInterfaces/SpecialErrorPageActionHandler.swift
jaceklyp: However, if we encounter more cases like this where we’re stretching these protocols, we might need to reconsider whether it’s appropriate to treat all special errors the same way. For example, SSL and malicious site errors already exhibit different behaviors. I’ll leave it up to you, but perhaps we should treat them differently (not under same logic) from the start. We can discuss further in 2025 :)

DuckDuckGo/TabViewController.swift
```diff
@@ -1375,55 +1375,64 @@ extension TabViewController: WKNavigationDelegate {
...
-                decisionHandler(.download)
-                return
-            }
-
-            // 3b. We know the response should trigger the file download prompt
-            self.presentSaveToDownloadsAlert(with: downloadMetadata) {
-                self.startDownload(with: navigationResponse, decisionHandler: decisionHandler)
-            } cancelHandler: {
+        // If the navigation has been handled by the special error page handler cancel navigating to new content as the special error page will be shown.
+        Task { @MainActor in
```
jaceklyp: Whoa :), this is triggering an alert for me. Making a task on this code could have bigger consequences – we’re essentially returning immediately and skipping the run loop, without a guarantee that other tasks spawned elsewhere won’t interfere with this. Are we sure this is the approach we want to take?

jaceklyp (changes requested):

DuckDuckGo/SpecialErrorPage/SpecialErrorPageInterfaces/SpecialErrorPageActionHandler.swift
alessandroboron: I think I will use two functions one without params and one with params as you suggested. I will make them optionals by having an empty implementation in a protocol extension so different handlers will implement only the one needed.

DuckDuckGo/MaliciousSiteProtection/MaliciousSiteProtectionManager.swift
alessandroboron: This is just for testing purposes. We won’t have any hardcoded website URLs in the upcoming BSK integration PR :)

DuckDuckGo/SpecialErrorPage/SpecialErrorPageNavigationHandler+MaliciousSite.swift
alessandroboron: My bad, there’s no reason to have this task detached, I will convert it to a normal task.

DuckDuckGo/TabViewController.swift
alessandroboron: The problem is that here we should be able to retrieve the task and wait on it if the operation hasn’t completed yet. Eg. In some scenarios we need to check with the backend if a website is malicious and this could take few seconds.

What if we convert the `WKNavigationDelegate` [decidePolicyFor navigationResponse](https://developer.apple.com/documentation/webkit/wknavigationdelegate/webview(_:decidepolicyfor:decisionhandler:)-19mn2) to its async counterpart? In this way we can perform the asynchronous operation in a controlled manner.

Happy to jump on a zoom call to discuss the possible approaches.

alessandroboron: I converted [decidePolicyFor navigationResponse](https://developer.apple.com/documentation/webkit/wknavigationdelegate/webview(_:decidepolicyfor:decisionhandler:)-19mn2) to its async counterpart as agreed.

I carefully checked that the values returned in the function are the same ones previously passed to the decisionHandler closure. I also smoked test the [download test sites](https://app.asana.com/0/414709148257752/1208359719493827/f) and everything looks good to me.

jaceklyp (approved):

PR #3716, opened 2024-12-11
re-add settings pixels

brindy:
Task/Issue URL: https://app.asana.com/0/392891325557410/1207818776591137/f
Tech Design URL:
CC:

**Description**:
Add pixels for various actions and views in settings.

**Steps to test this PR**:
1. Go through the list of pixels in the linked Asana task and check they fire at the correct time
2. A pixel fired when a view is opened should not fire again if the user pops the stack back to that view
3. None of the pixels should have any addition parameters other than the default (ie app version)

### DIFF
--- a/Core/PixelEvent.swift
+++ b/Core/PixelEvent.swift
@@ -778,7 +778,6 @@ extension Pixel {
         case settingsAccessibilityOpen
         case settingsAccessiblityTextZoom
 
-        #warning("NEW PIXELS - remove this comment")
         case settingsPrivateSearchOpen
         case settingsEmailProtectionOpen
         case settingsEmailProtectionEnable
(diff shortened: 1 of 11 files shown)


Core/PixelEvent.swift
```diff
@@ -776,6 +778,26 @@ extension Pixel {
...
+        case settingsSyncOpen
+        case settingsAppearanceOpen
+        case settingsThemeSelectorPressed
+        case settingsAddressBarTopSelected
+        case settingsAddressBarBottomSelected
+        case settingsShowFullURLOn
+        case settingsShowFullURLOff
+        case settingsDataClearingOpen
+        case settingsFireButtonSelectorPressed
+        case settingsDataClearingClearDataOpen
```
mallexxx, line 794: `settingsDataClearingClearDataOpen ` pixel is not used anywhere

Core/PixelEvent.swift
```diff
@@ -776,6 +778,26 @@ extension Pixel {
...
+        case settingsAutomaticallyClearDataOn
```
mallexxx, line 795: `settingsAutomaticallyClearDataOn` is fired for every action or timing parameter change in the Automatically Clear Data settings screen

mallexxx (changes requested):
please see 2 issues I commented inline, otherwise looks good

brindy:
Thanks @mallexxx - fixed both of those.

mallexxx (approved):
LGTM 👍

PR #3720, opened 2024-12-12
Fix AI Chat address bar pixel

Bunn:
Task/Issue URL: https://app.asana.com/0/414235014887631/1208965876736143/f

**Description**:
Fire AI Chat address bar pixel
Fixes share button pixel

**Steps to test this PR**:
Set yourself as internal user
1. Do a search, tap on AI Chat address bar, check if we send the m_aichat_addressbar_icon pixel
2. Open a webpage, tap on share on the address bar, check if we send the m_addressbar_share pixel


brindy (approved):
LGTM!

PR #3727, opened 2024-12-13
Refactor AppDelegate (milestone 1)

jaceklyp:
Task/Issue URL: https://app.asana.com/0/0/1208832732122403/f
Tech Design URL: https://app.asana.com/0/481882893211075/1208859623176995/f
CC: @bwaresiak 

**Description**:
- Migrate the app’s startup flow into a state machine with clearly defined states, each specifying its dependencies to ensure the necessary resources are available before execution.
- Keep the flow primarily synchronous as it currently is, with some exceptions for existing async code.
- Define explicit state dependencies to avoid optional objects, ensuring clarity about what exists in each state.

**Steps to test this PR**:
1. Make sure you're assigned to "experimental" state machine group (change appBehavior random value inside AppDelegate)
2. Please go through all smoke tests. https://app.asana.com/0/1202500774821704/1201879741377823/f
3. Run the UI tests.
4. Repeat for old behavior.

**Device Testing**:

* [ ] iPhone 14 Pro
* [ ] iPad

**OS Testing**:

* [ ] iOS 17


DuckDuckGo/AppDelegate.swift
```diff
@@ -110,378 +124,404 @@ import os.log
...
+    private var voiceSearchHelper: VoiceSearchHelper!
 
+    private var marketplaceAdPostbackManager: MarketplaceAdPostbackManager!
     private var didFinishLaunchingStartTime: CFAbsoluteTime?
 
-    private let appStateMachine = AppStateMachine()
+    private let appStateMachine: AppStateMachine = AppStateMachine()
+    private let appBehavior: AppBehavior = {
+        guard !ProcessInfo().arguments.contains("testing") else { return .existing }
+        if let appBehavior = AppDependencyProvider.shared.appSettings.appBehavior {
```
dus7: I'm afraid we can run into an unexpected behavior here, i.e. not being consistent for a single app instance. AppSettings are user defaults and this can be executed before data is unlocked. We are already observing inconsistencies with UserDefaults (https://app.asana.com/0/414235014887631/1208659072736427/f) so maybe it's a good idea to steer away from using those here?

DuckDuckGo/AppDelegate.swift
```diff
@@ -41,24 +41,31 @@ import PixelExperimentKit
...
 import os.log
 
+enum AppBehavior: String {
+
+    case existing
+    case stateMachine
+
+}
+
 @UIApplicationMain class AppDelegate: UIResponder, UIApplicationDelegate {
```
dus7, line 39: I know it'll be a bit more work at this point, but did you consider extracting the code into 2 separate types, say `ExistingAppDelegate` and `StateMachineAppDelegate`, then use them in this class? It'd have a few advantages for the future:
* It'd lower the risk of introducing unwanted changes to existing approach, for example by making changes to instance properties.
* Requirements, dependencies and instance properties for the state machine would be clearly visible and easier to maintain.
* Reasoning about the approach and overall shape will be easier.
* Switching the implementation after the migration is complete could be as easy as replacing `AppDelegate` contents with whatever is inside `StateMachineAppDelegate`.

DuckDuckGo/AppLifecycle/AppStateMachine.swift
```diff
@@ -21,27 +21,30 @@ import UIKit
...
 
-    func apply(event: AppEvent) -> any AppState
+    mutating func apply(event: AppEvent) -> any AppState
 
 }
 
 protocol AppEventHandler {
 
+    @MainActor
     func handle(_ event: AppEvent)
```
dus7, line 43: Marking protocol function declaration as MainActor won't enforce it (based on my tests). It's better to mark the method as async here and defer to the implementation to isolate for specific actor.

DuckDuckGo/AppLifecycle/AppStates/Active.swift
```diff
@@ -17,12 +17,500 @@
...
+                if isEnabled {
+                    subscriptionCookieManager.enableSettingSubscriptionCookie()
+                } else {
+                    await subscriptionCookieManager.disableSettingSubscriptionCookie()
+                }
+            }
+        }
+
+        // onApplicationLaunch code
+        Task { @MainActor [self] in // is capturing self here ok?
```
dus7: > is capturing self here ok?

`self` is a value type so you'll be capturing a copy. If `self` value changes you won't see those changes here.

DuckDuckGo/AppServices/UIService.swift
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    func displayBlankSnapshotWindow(voiceSearchHelper: VoiceSearchHelper,
+                                    addressBarPosition: AddressBarPosition) {
+        guard overlayWindow == nil else { return }
+
+        overlayWindow = UIWindow(frame: window.frame)
+        overlayWindow?.windowLevel = UIWindow.Level.alert
+
+        // TODO: most likely we do not need voiceSearchHelper for BlankSnapshotVC
```
dus7, line 40: It's used to toggle mic's visibility in the omni bar, so I'm afraid it's here to stay.

dus7 (commented):
Posting my comments so far. I'll still need to do some smoke testing.

DuckDuckGo/AppLifecycle/AppStateMachine.swift
jaceklyp: It was added because of a warning - if it is missing, I'm getting complaint about "Main-actor isolated instance method 'handle' cannot be used to..."

DuckDuckGo/AppServices/UIService.swift
jaceklyp: yeah I'm aware I left this note for myself, as I'm pretty sure we don't need the whole voiceSearchHelper object, but rather a simple bool to be passed there (or in the worst case scenario some simple enum)

DuckDuckGo/NewAppDelegate.swift
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    func applicationDidBecomeActive(_ application: UIApplication) {
+        if !ProcessInfo().arguments.contains("testing") {
+            appStateMachine.handle(.activating)
+        }
+    }
+
+    func applicationWillResignActive(_ application: UIApplication) {
+        appStateMachine.handle(.suspending)
```
bwaresiak, line 43: Can this be called during testing? Would it cause inconsistencies?

DuckDuckGo/NewAppDelegate.swift
```diff
@@ -0,0 +1,67 @@
...
+        appStateMachine.handle(.handleShortcutItem(shortcutItem))
+    }
+
+    func application(_ app: UIApplication, open url: URL, options: [UIApplication.OpenURLOptionsKey: Any] = [:]) -> Bool {
+        appStateMachine.handle(.openURL(url))
+        return true
+    }
+
+    func refreshRemoteMessages() {
+        // part of debug menu, let's not support it in the first iteration
```
bwaresiak, line 62: We should be able to easily pass remoteMessagingClient to Settings and Debug view - it's already done for multiple objects.

DuckDuckGo/AppLifecycle/AppStateTransitions.swift
```diff
@@ -24,8 +24,8 @@ extension Init {
 
     func apply(event: AppEvent) -> any AppState {
         switch event {
-        case .launching(let application, let launchOptions):
-            return Launched(application: application, launchOptions: launchOptions)
+        case .launching(let application):
+            return Launched(stateContext: makeStateContext(application: application))
         default:
             return handleUnexpectedEvent(event)
```
bwaresiak, line 33: Just a note: make sure to rename the pixel (if you want to avoid using app version param) before actual tests so we can easily see if things are not working as expected.

DuckDuckGo/AppLifecycle/AppStates/Launched.swift
```diff
@@ -17,12 +17,663 @@
...
+
+    private let application: UIApplication
+
+    // swiftlint:disable:next cyclomatic_complexity
+    init(stateContext: Init.StateContext) {
+
+        application = stateContext.application
+        privacyProDataReporter = PrivacyProDataReporter(fireproofing: fireproofing)
+        vpnWorkaround = VPNRedditSessionWorkaround(accountManager: accountManager, tunnelController: tunnelController)
+        crashReportUploaderOnboarding = CrashCollectionOnboarding(appSettings: AppDependencyProvider.shared.appSettings)
```
bwaresiak, line 95: What is a bit concerning here, is that we change the moment things are instantiated and when data is red from the storage  - and it's potentially earlier than it used to be in the old implementation so things may not be ready yet. E.g. some of the code here for VPN workaround was instantiated lazily in Task {} when did finish launching has already returned to the OS.

DuckDuckGo/NewAppDelegate.swift
jaceklyp: fixed with new Testing state

DuckDuckGo/AppLifecycle/AppStates/Launched.swift
jaceklyp: talked through on 1:1, it shouldn't cause any issues, as there's no work happening in the inits

dus7 (approved):
Went through smoke tests, including data clearing, automatic keyboard, VPN, app lock, sync, opening URL, default browser, context menu items. All seems to be working fine.

Good effort! Looking forward to what comes next.💪

PR #3733, opened 2024-12-17
Bugfix: Fix Youtube Internal links when 'Watching on Youtube' from Duck Player

afterxleep:
Task/Issue URL: https://app.asana.com/0/1204099484721401/1208930843675395/f
Tech Design URL:
CC:

**Description**:
Fixes an issue causing Youtube internal links to open Duck Player

**Steps to test this PR**:
- [ ] 1. Set Duck Player mode to 'Always'
- [ ] 2. Go to Youtube.com and open a video in Duck Player
- [ ] 3. Tap "Watch on Youtube"
- [ ] 4. Tap the Youtube "search" icon or the page's three dot menu
- [ ] 5. Confirm Youtube UI shows up and Duck Player does not open.

Demo:
RocketSim_Recording_iPhone_(18 0)_6 1_2024-12-17_10 02 35

### DIFF
--- a/DuckDuckGoTests/YoutublePlayerNavigationHandlerTests.swift
+++ b/DuckDuckGoTests/YoutublePlayerNavigationHandlerTests.swift
@@ -582,4 +582,24 @@ class DuckPlayerNavigationHandlerTests: XCTestCase {
         XCTAssertNil(tabNavigator.openedURL, "No new tabs should open")
     }
     
+    @MainActor
+    func testHandleDelegateNavigation_YoutubeWatchURLWithDuckPlayerEnabledAndSameVideoNavigation_ReturnsFalse() async {
+        // Arrange
+        let youtubeURL = URL(string: "https://www.youtube.com/watch?v=abc123")!
+        let youtubeInternalURL = URL(string: "https://www.youtube.com/watch?v=abc123&settings")!
+        let request = URLRequest(url: youtubeURL)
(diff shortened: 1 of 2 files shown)


DuckDuckGoTests/YoutublePlayerNavigationHandlerTests.swift
```diff
@@ -582,4 +582,24 @@ class DuckPlayerNavigationHandlerTests: XCTestCase {
         XCTAssertNil(tabNavigator.openedURL, "No new tabs should open")
     }
     
+    @MainActor
```
federicocappelli, line 585: Is @MainActor needed here? I think that Bartek is working thorwards parallelising the unit tests and this could interfeer, In any case I don't think we should run tests on main therad.

federicocappelli (approved):
LGTM, only a question about the test

federicocappelli (approved):
LGTM, only a question about the test

DuckDuckGoTests/YoutublePlayerNavigationHandlerTests.swift
afterxleep: Good call.  Although not required, it was to align with the convention that any code interacting with WebKit or UI components runs on the main thread.   

Like If you replace MockWebView with a real WKWebView or integrate other WKNavigationDelegate-related logic in the future, @MainActor will help avoid threading issues.   Not likely happen in a test though.   I'll make a note to revise/update the class.

PR #3775, opened 2025-01-07
Duck.ai deeplink

Bunn:
Task/Issue URL: https://app.asana.com/0/1204167627774280/1208991515313413/f
Tech Design URL:
CC:

**Description**:

Add deep-link integration to duck.ai. Opening duck.ai from anywhere should open the custom webview instead of a new tab

**Steps to test this PR**:
1. Host this [HTML](https://gist.github.com/Bunn/f89fedbb3ef004cba4c6aee4b7f1bae4) 
2. Change [this](https://github.com/duckduckgo/iOS/blob/fcf09ae9ab75f1260b77c9c62ec8c66ed9dd9838/DuckDuckGo/TabURLInterceptor.swift#L105) to return true
3. Make sure that tapping every button opens the custom AI Chat view instead of a tab
4. Try typing duck.ai directly in the address bar and make sure it opens the custom AI Chat
5. Make sure that there are no changes in behavior without the change done [here](https://github.com/duckduckgo/iOS/blob/fcf09ae9ab75f1260b77c9c62ec8c66ed9dd9838/DuckDuckGo/TabURLInterceptor.swift#L105)


afterxleep (approved):
01-09 at 17 24

Works as expected.  :shipit:

PR #3785, opened 2025-01-09
AI Chat handoff

Bunn:
Task/Issue URL: https://app.asana.com/0/1204167627774280/1209119202522826/f

**Description**:
Add handoff support using the C-S-S communication layer

**Steps to test this PR**:
This feature depends on some FE changes in the SERP, so to test it you'll need to do a bit of initial setup:

1. Use [this](https://jsonblob.com/api/1326573454710988800) privacy config
2. Host [this](https://gist.github.com/Bunn/8071483a64bf82b8f7a24ca26d2d7a43) locally and expose it via ngrok (the config file above is setup to allow the ngrok-free domain, feel free to adjust if you're using something else)
3. Apply [this patch](https://gist.github.com/Bunn/388359a7a5d1ae964cf66d55fcea2d44) which will: Remove the webview callback from within the  AI Chat custom view, setup the ngrok URL
 
----
1.  Open the test website, tap on "open with payload". Validate that the AI chat view displays a large payload
2. Open the test website, tap on "open without payload". Validate that the AI chat view display null payload
3. Open the test website, tap on "open with random payload". Validate that the AI chat view display a random payload each time (Do this a couple times)
4. When doing the above, you can also validate if the payload is correctly consume by tapping on "Update settings".


DuckDuckGo/AIChat/AIChatViewControllerManager.swift
```diff
@@ -0,0 +1,95 @@
...
+    func aiChatViewControllerManager(_ manager: AIChatViewControllerManager, didRequestToLoad url: URL)
+}
+
+final class AIChatViewControllerManager {
+    weak var delegate: AIChatViewControllerManagerDelegate?
+    private var aiChatUserScript: AIChatUserScript?
+    private var payloadHandler = AIChatPayloadHandler()
+
+    @MainActor
+     lazy var aiChatViewController: AIChatViewController = {
```
brindy: nit: indentation

DuckDuckGo/AIChat/UserScript/AIChatUserScript.swift
```diff
@@ -0,0 +1,63 @@
...
+    private(set) var messageOriginPolicy: MessageOriginPolicy
+
+    init(handler: AIChatUserScriptHandling) {
+        self.handler = handler
+        var rules = [HostnameMatchingRule]()
+
+        /// Default rule for DuckDuckGo AI Chat
+        if let ddgDomain = URL.ddg.host {
+            rules.append(.exact(hostname: ddgDomain))
+        }
```
brindy, line 44: Just acknowledging we discussed this.  No action required.

DuckDuckGo/AIChat/AIChatViewControllerManager.swift
```diff
@@ -0,0 +1,95 @@
...
+        let settings = AIChatSettings(privacyConfigurationManager: ContentBlocking.shared.privacyConfigurationManager,
+                                      internalUserDecider: AppDependencyProvider.shared.internalUserDecider)
```
brindy: Can we inject these instead?  Since this is lazy they should be available by this point.

DuckDuckGo/UserScripts.swift
```diff
@@ -66,6 +67,8 @@ final class UserScripts: UserScriptsProvider {
                                                                 properties: sourceProvider.contentScopeProperties,
                                                                 isIsolated: true)
         autoconsentUserScript = AutoconsentUserScript(config: sourceProvider.privacyConfigurationManager.privacyConfig)
+        aiChatUserScript = AIChatUserScript(handler: AIChatUserScriptHandler(featureFlagger: AppDependencyProvider.shared.featureFlagger))
```
brindy: please inject :)

brindy (changes requested):
Nice work.  Please fix the dependency injection but otherwise looks good.

brindy (approved):
LGTM

PR #3572, opened 2024-11-13
pixels for bookmarks > favorites tab

brindy:
Task/Issue URL: https://app.asana.com/0/392891325557410/1208733952836859/f
Tech Design URL:
CC:

**Description**:
Pixels for bookmarks > favorites tab.

**Steps to test this PR**:
1. Change DailyPixel.swift so that `hasBeenFiredToday` returns false
2. Add some favorites
3. Go to the bookmarks UI and tap on the favourites tab
4. Tap on manage -> pixel should fire `m_bookmarks_ui_favorites_manage_daily`
6. Exit edit mode
7. Edit a favorite -> pixel should fire `m_bookmarks_ui_favorites_action_daily`
8. Move a favorite -> pixel should fire `m_bookmarks_ui_favorites_action_daily`
9. Remove a favorite -> pixel should fire `m_bookmarks_ui_favorites_action_daily`
10. Check that actions on the new tab page do not fire the above pixels
11. Check the overlay does not trigger these pixels to fire

PR #3803, opened 2025-01-14
Revert "pixels for bookmarks > favorites tab (#3572)"

brindy:
Task/Issue URL: https://app.asana.com/0/392891325557410/1208736944745857/f
Tech Design URL:
CC:

**Description**:
This reverts commit 95897a71d6c75ed877c1baad334df8a902083f17 as the pixels are no longer required.

**Steps to test this PR**:
1. Build and run the app.
2. Open bookmarks manager and tap on favorites view and smoke test it.

### DIFF
--- a/DuckDuckGo/FavoritesHomeViewSectionRenderer.swift
+++ b/DuckDuckGo/FavoritesHomeViewSectionRenderer.swift
@@ -156,7 +156,6 @@ class FavoritesHomeViewSectionRenderer {
         guard let indexPath = collectionView.indexPath(for: cell),
         let favorite = viewModel.favorite(at: indexPath.row) else { return }
         Pixel.fire(pixel: .homeScreenDeleteFavorite)
-        fireActionPixel()
         viewModel.removeFavorite(favorite)
         WidgetCenter.shared.reloadAllTimelines()
         collectionView.performBatchUpdates {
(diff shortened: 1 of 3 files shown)


DuckDuckGo/FavoritesHomeViewSectionRenderer.swift
```diff
@@ -287,11 +285,10 @@ class FavoritesHomeViewSectionRenderer {
 
         coordinator.drop(dragItem, toItemAt: destinationPath)
 
-        fireActionPixel()
     }
 
     func collectionView(_ collectionView: UICollectionView, itemsForBeginning session: UIDragSession, at indexPath: IndexPath) -> [UIDragItem] {
-        guard let cell = collectionView.cellForItem(at: indexPath) as? FavoriteHomeCell else { return [] }
+        guard let cell = self.collectionView(collectionView, cellForItemAt: indexPath) as? FavoriteHomeCell else { return [] }
```
dus7: This cannot be reverted - I remember the original code was causing an exception on iOS 18 while reordering favorites.

dus7 (changes requested):
Highlighted one thing that can potentially cause a crash.

DuckDuckGo/FavoritesHomeViewSectionRenderer.swift
brindy: Ah, good spot!

brindy: Was able to get it crashing instantly and not reverting this fixed that.  Thanks.

dus7 (approved):

PR #3805, opened 2025-01-15
tds override

SabrinaTardio:
Task/Issue URL: https://app.asana.com/0/72649045549333/1209114783599181/f
Tech Design URL: https://app.asana.com/0/481882893211075/1208547600159199
CC:

**Description**: Makes the tds fetch dependent on the privacy configuration fetch (so that the url for tds can be taken from the configuration). Sets up experiment for tds. Note metrics will be added in a separate PR.

**Steps to test this PR**:
1. Run the Application
2. On the debugger filter for "Stored etag for trackerDataSet" and write down the etag
3. Go to Settings -> All debug Options -> Configuration URLs 
4. Click on privacy config and use this URL: **### https://staticcdn.duckduckgo.com/trackerblocking/config/test/macos-config.json**
5. On the debugger filter for "Stored etag for trackerDataSet" and check the etag is now **2ce60c57c3d384f986ccbe2c422aac44**
6. Check also that "CBR compilation for TrackerDataSet succeeded" has been logged
7. Check “experiment_enroll_tdsNextExperimentBaseline_treatment” has been sent
8.  Click on privacy config and reset to default URL
9. On the debugger filter for "Stored etag for trackerDataSet" and check the etag has gone back to previous one (382541021cbc55b218714db779561a51)
10. Check also that "CBR compilation for TrackerDataSet succeeded" has been logged

<!—
Before submitting a PR, please ensure you have tested the combinations you expect the reviewer to test, then delete configurations you *know* do not need explicit testing.

Using a simulator where a physical device is unavailable is acceptable.
—>

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPhone 14 Pro
* [ ] iPad

**OS Testing**:

* [ ] iOS 15
* [ ] iOS 16
* [ ] iOS 17

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

—
###### Internal references:


DuckDuckGo/Configuration/ConfigurationManager.swift
```diff
@@ -224,3 +244,13 @@ extension ConfigurationManager {
...
+
+#if DEBUG
+extension ConfigurationManager {
+    func setContentBlockingManagers(trackerDataManager: TrackerDataManager,
+                                    privacyConfigurationManager: PrivacyConfigurationManager) {
+        self.trackerDataManager = trackerDataManager
+        self.privacyConfigurationManager = privacyConfigurationManager
+    }
+}
+#endif
```
bwaresiak: Maybe these managers should be init-injected instead of creating this API?

DuckDuckGo/Configuration/ConfigurationManager.swift
```diff
@@ -117,10 +120,27 @@ final class ConfigurationManager: DefaultConfigurationManager {
...
+            try await fetcher.fetch(.privacyConfiguration, isDebug: isDebug)
+            didFetchAnyTrackerBlockingDependencies = true
+            privacyConfigurationManager.reload(etag: store.loadEtag(for: .privacyConfiguration),
+                                               data: store.loadData(for: .privacyConfiguration))
+        } catch {
+            Logger.general.error("Did not apply update to \(Configuration.privacyConfiguration.rawValue, privacy: .public): \(error.localizedDescription, privacy: .public)")
+        }
+
+        // Start trackerDataSet fetch task after privacyConfiguration completes
+        let trackerDataSetTask = Task { try await fetcher.fetch(.trackerDataSet, isDebug: isDebug) }
```
bwaresiak, line 141: I wonder - is there a chance that URL in the config will point to a wrong (not existing) file? Would we catch that with any metrics if that's the case?

Core/AppConfigurationURLProvider.swift
```diff
@@ -20,19 +20,119 @@
...
+        do {
+            if let settingsDict = try JSONSerialization.jsonObject(with: jsonData) as? [String: String],
+               let urlString = cohort == .control ? settingsDict["controlUrl"] : settingsDict["treatmentUrl"] {
+                return URL(string: URL.staticBase + "/trackerblocking/" + urlString)!
+            }
+        } catch {
+            Logger.config.info("privacyConfiguration: Failed to parse subfeature settings JSON: \(error)")
+        }
+        return nil
+    }
```
bwaresiak: Seeing how changes to this file are extremely similar between iOS and macOS, is there a way to extract this somewhere to BSK (which potentially could help us reason about places where properties actually differ, instead of trying to spot them)?

SabrinaTardio: The changes are very similar, probably literally the same, yes I was also thinking about it…
I am reluctant about having a common AppConfigurationURLProvider since the URLs are different and might becoming confusing, but make we could extract:
- a TrackerDataURL which could live in BSK (I think both privacyConfig and Feature Flagger are in the BSK module so there should not be dependency cycle problems)
- We could also move the Tds experiment type and Flag somewhere in BSK

DuckDuckGo/Configuration/ConfigurationManager.swift
SabrinaTardio: Yes, there is a chance that the URL is wrong (anything can be wrong ahah)
I see there is a pixel sent at data validation time if the payload is wrong. If it can’t fetch at all we just throw which would then just log.
Effectively though what would happen is that we would fallback on the embedded one and then we would see in the debug pixel of breakage metrics that we get the wrong etag for the wrong cohort.
We could however having a specific pixel instead of just a log when fetching fails. What do you think?

DuckDuckGo/Configuration/ConfigurationManager.swift
SabrinaTardio: So this are setup like this because ContentBlocking.shared is not available at the time ConfigurationManager is initialised (hence the lazy at the top)
We could have an internal initialiser for test and use ContentBlocking.shared if the ones passed by the initialiser are nill. Again was in two mind about it...

DuckDuckGo/Configuration/ConfigurationManager.swift
SabrinaTardio: It will be surfaced in the compilation pixels (since we will have cohort and tds etag

DuckDuckGo/Configuration/ConfigurationManager.swift
SabrinaTardio: Never mind it was just an issue on macOS and it was just a matter of order of initialisation in the AppDelegate (I thought there was a deeper reason)

Core/AppConfigurationURLProvider.swift
SabrinaTardio: I moved forward with this plan (as agreed)

DuckDuckGo/Configuration/ConfigurationManager.swift
```diff
@@ -71,9 +74,14 @@ final class ConfigurationManager: DefaultConfigurationManager {
...
 
-    override init(fetcher: ConfigurationFetching = ConfigurationFetcher(store: ConfigurationStore(), eventMapping: configurationDebugEvents),
-                  store: ConfigurationStoring = AppDependencyProvider.shared.configurationStore,
-                  defaults: KeyValueStoring = UserDefaults(suiteName: "\(Global.groupIdPrefix).app-configuration") ?? UserDefaults()) {
+    init(fetcher: ConfigurationFetching = ConfigurationFetcher(store: ConfigurationStore(), eventMapping: configurationDebugEvents),
+         store: ConfigurationStoring = AppDependencyProvider.shared.configurationStore,
+         defaults: KeyValueStoring = UserDefaults(suiteName: "\(Global.groupIdPrefix).app-configuration") ?? UserDefaults(),
+         trackerDataManager: TrackerDataManager = ContentBlocking.shared.trackerDataManager,
+         privacyConfigurationManager: PrivacyConfigurationManaging = ContentBlocking.shared.privacyConfigurationManager,
+         contentBlockingManager: ContentBlockerRulesManagerProtocol = ContentBlocking.shared.contentBlockingManager) {
```
bwaresiak: Why the `contentBlockingManager` ?

SabrinaTardio: just a copy paste mistake! I removed it

bwaresiak (approved):

PR #3806, opened 2025-01-15
Validate Webview URL before reloading

afterxleep:
Task/Issue URL: https://app.asana.com/0/1204099484721401/1208801010014872/f
Tech Design URL:
CC:

**Description**:
Attempt to fix [webView Reload crash on a terminated webview](https://errors.duckduckgo.com/organizations/ddg/issues/191825/?project=8&query=is%3Aunresolved+firstSeen%3A-9d&referrer=issue-stream&sort=freq&statsPeriod=14d&stream_index=0).

If the webviewContent is terminated and fails to restart, calling reload on an invalid URL may cause a the crash.  This ensures there’s still a valid URL before attempting a reload.

**Steps to test this PR**:
1. CI should be 🥬

### DIFF
--- a/DuckDuckGo/DuckPlayer/DuckPlayerNavigationHandler.swift
+++ b/DuckDuckGo/DuckPlayer/DuckPlayerNavigationHandler.swift
@@ -813,13 +813,13 @@ extension DuckPlayerNavigationHandler: DuckPlayerNavigationHandling {
         
         // Reset DuckPlayer status
         duckPlayer.settings.allowFirstVideo = false
-        
-        guard isDuckPlayerFeatureEnabled else {
-            webView.reload()
+                
+        guard let url = webView.url else {
             return
         }
         
-        guard let url = webView.url else {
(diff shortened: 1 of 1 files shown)


studiosutara (approved):
I have approved the PR, the code itself looks good. 
However, I noticed that the stack trace seems to be pointing to the this line, unrelated to the fix:
`duckPlayer.settings.allowFirstVideo = false`
Unless I'm missing something, is this line worth another look?

afterxleep:
@studiosutara I’m not sure I follow.   

By looking at the code in this specific app version (7.152.0), It seems to me `webView.reload()` in the `DuckPlayerNavigationHandler.handleReload.825` is the trigger for the crash, as we may be attempting to reload a content-terminated webview.  

Also, `allowFirstVideo` in DuckPlayer Settings does not seem to trigger any side effects that could cause the crash, and this does not look like a deallocated instance.

Am I missing something?

PR #3809, opened 2025-01-15
Remove old app delegate, rename states and introduce Resuming (a.k.a willEnterForeground)

jaceklyp:
Task/Issue URL: https://app.asana.com/0/0/1209022783720243/f
Tech Design URL: https://app.asana.com/0/0/1209144186921382/f
CC: @bwaresiak 

**Description**:
- remove old app delegate code,
- rename states and events, so they better reflect what is going on
- introduce some additional very much needed transitions
- introduce Resuming state

**Steps to test this PR**:
- smoke test the app, make sure handleUnexpectedEvent is never called


DuckDuckGo/AppDelegate.swift
```diff
@@ -45,65 +29,45 @@ protocol DDGApp {
...
     func application(_ application: UIApplication,
                      performActionFor shortcutItem: UIApplicationShortcutItem,
                      completionHandler: @escaping (Bool) -> Void) {
-        realDelegate.application?(application, performActionFor: shortcutItem, completionHandler: completionHandler)
+        appStateMachine.handle(.handleShortcutItem(shortcutItem))
     }
 
     func application(_ app: UIApplication, open url: URL, options: [UIApplication.OpenURLOptionsKey: Any] = [:]) -> Bool {
-        realDelegate.application?(app, open: url, options: options) ?? false
+        appStateMachine.handle(.openURL(url))
```
dus7, line 70: Are we ok with always returning `true` here? I saw there's `false` supposed to be returned when onboarding is present (based on the Old Delegate).

DuckDuckGo/AppDelegate.swift
```diff
@@ -121,12 +85,12 @@ protocol DDGApp {
...
 
     func application(_ application: UIApplication, willContinueUserActivityWithType userActivityType: String) -> Bool {
-        return true
+        true
     }
 
     /// It's public in order to allow refreshing on demand via Debug menu. Otherwise it shouldn't be called from outside.
     func refreshRemoteMessages() {
-        realDelegate.refreshRemoteMessages()
+        // part of debug menu, let's not support it in the first iteration
```
dus7: I think we'll need this now, won't we?

dus7 (approved):
Works fine for me. Didn't hit any unexpected events.

I left two comments. They are not critical, but would be good to plan for resolving those in your project.

DuckDuckGo/AppDelegate.swift
jaceklyp: it doesn't work now, as we already use new app delegate by default, but you are correct, this should be brought back, I'll make it work in next PR

DuckDuckGo/AppDelegate.swift
jaceklyp: I saw this and AFAIU the documentation correctly we can always return true, just like we do inside didFInishLaunching but we do not necessarily have to handle the opening url logic, hence the code inside Foreground state:

guard mainViewController.needsToShowOnboardingIntro() == false else {
             return
         }

DuckDuckGo/AppDelegate.swift
jaceklyp: I fixed it in this PR!

PR #3810, opened 2025-01-15
AI Chat message policy debug

Bunn:
Task/Issue URL: https://app.asana.com/0/1204167627774280/1209160773739107/f

**Description**:
Add a debug settings in the debug UI for people to be able to allow a custom domain in the message policy

**Steps to test this PR**:
1. Add a breakpoint here
2. Open a new tab, make sure the breakpoint is not called
3. Go to the debug UI, AI Chat, then set a custom hostname
4. Open a new tab, make sure the breakpoint is called with the data you set before
5. Go to the debug UI, AI Chat, then reset the custom domain
6. Open a new tab, make sure the breakpoint is not called


brindy (approved):
LGTM, I checked that `rules.append(.exact(hostname: debugHostname))` was only called when a host name had been set.

Couple of SwiftLint things to fix, but approved.

PR #3811, opened 2025-01-15
Remove pre-warm

Bunn:
Task/Issue URL: https://app.asana.com/0/1204167627774280/1209157710205507/f

**Description**:
Remove AIChatWebViewController from memory once the view is dismissed

**Steps to test this PR**:
1. Make sure to set internal user state
2. Smoke test opening AI Chat. Open it from the address bar if you're in the SERP, open it from the browser menu
3. Inspect memory or add a log like
```
    deinit {
        print("BYE")
    }
```
in `AIChatWebViewController` and make sure that every time you dismiss it, the view is called.


brindy (approved):
LGTM

PR #3815, opened 2025-01-16
Inject larger autoconsent bundle containing cosmetic filters

dharb:
Task/Issue URL: https://app.asana.com/0/72649045549333/1209218126526339/f
Tech Design URL: https://app.asana.com/0/481882893211075/1209010810078878/f
CC: @muodov @samsymons @brindy 

**Description**:
This PR is the first step in expanding CPM coverage on mobile by way of using easylist cookie filterlists. There are no user facing changes in this PR: all that's happening here is we are adding a new autoconsent subfeature called `filterlist`, and injecting a different autoconsent bundle if that feature flag is enabled via privacy configuration.

**Steps to test this PR**:
1. Checkout this branch, build locally
2. Point app at this config: https://jsonblob.com/api/1329908032888561664. The only change here is the added `filterlist` autoconsent subfeature:
``` JSON
"filterlist": {
    "state": "enabled"
}
```
3. Ensure that CPM is on, then visit a website that we have an existing CPM rule for (eg ecosia.org). Confirm that the cookie popup is hidden.
4. Visit a website that we don't have an existing CPM rule for, but there is an easylist rule for (eg rijksmuseum.nl). Confirm that the cookie popup is not hidden.

To fully confirm that the larger bundle is being injected when the feature flag is enabled, you may do so by enabling the feature as follows:
1. Around https://github.com/duckduckgo/iOS/pull/3815/files#diff-cc6c354af70301c8b10e53aed7c5539b9fa7816ccd7779971469c684ed87caedR263, add
`let isFilterListExperimentEnabled = config.isSubfeatureEnabled(AutoconsentSubfeature.filterlist)`
2. Pass this param by editing the `config` key in `replyHandler` to add `"enableFilterList": isFilterListExperimentEnabled`
3. Rebuild app
4. Double check that config is still pointed at the jsonblob url above
5. Ensure that CPM is on, then visit a website that we have an existing CPM rule for (eg ecosia.org). Confirm that the cookie popup is hidden.
6. Visit a website that we don't have an existing CPM rule for, but there is an easylist rule for (eg rijksmuseum.nl). Confirm that the cookie popup is hidden.
7. As a final check, use the fire button to clear all local data, then point privacy config back at the production endpoint where this feature isn't enabled. Then visit ecosia.org and rijksmuseum.nl again and confirm that the cookie pop _is_ hidden on ecosia.org, and is _not_ hidden on rijksmuseum.nl

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPhone 14 Pro
* [ ] iPad

**OS Testing**:

* [ ] iOS 15
* [ ] iOS 16
* [ ] iOS 17

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:


DuckDuckGo/Autoconsent/userscript-exp.js
```diff
@@ -0,0 +1,19 @@
+import AutoConsent from '@duckduckgo/autoconsent/extra';
```
dharb, line 1: This file is just a copy of `DuckDuckGo/Autoconsent/userscript.js` with one change - in line 1 we use `'@duckduckgo/autoconsent/extra'` instead of `'@duckduckgo/autoconsent'`, which generates the larger bundle including easylist filters.

dus7 (approved):
Works as advertised. 
I tested the described scenario, confirmed larger bundle is being loaded. I understand we don't want to enable `filterList` just yet, only inject the bundle. 

What I'm worried about is memory consumption caused by this change. With small bundle `AutoconsentUserScript` takes ~450kB **per each tab**, while when the large bundle is loaded it 4.13 MB each. That's more than 800% increase. I think we need to observe app vitals very carefully after this is released.

brindy (changes requested):
I think this needs some explicit instrumentation to measure the impact so we can compare it against where it hasn't been enabled.  

But I've seen we're disabling it for desktop so actually probably best not to merge this at all, so just marking this as requesting changing to prevent accidental merging before we've had chance to discuss it further, as per https://app.asana.com/0/0/1209233669401395/1209227765756701/f

Thanks.

brindy (approved):
Approved, as per discussion here: https://app.asana.com/0/0/1209233669401395/f

brindy:
@dharb just FYI I updated the branch from `main` for you because there was some structural changes to deal with.

dharb:
Closing as we decided not to go this route for the time being.

PR #3821, opened 2025-01-17
better guards on access tab index

brindy:
Task/Issue URL: https://app.asana.com/0/392891325557410/1209158320724215/f
Tech Design URL:
CC:

**Description**:
Speculative fix for a crash. Checks index is available in indices before accessing the array.

**Steps to test this PR**:
Smoke test tab manager on iPhone and iPad
1.  Grid and list mode
2. Add new tabs 
3. Delete tabs
4. Reorder tabs
5. Close and re-open the app

### DIFF
--- a/DuckDuckGo/TabsModel.swift
+++ b/DuckDuckGo/TabsModel.swift
@@ -75,7 +75,7 @@ public class TabsModel: NSObject, NSCoding {
 
     var currentTab: Tab? {
         let index = currentIndex
-        return tabs[index]
+        return tabs.indices.contains(index) ? tabs[index] : nil
     }
 
     var count: Int {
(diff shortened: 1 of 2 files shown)


DuckDuckGo/TabsModel.swift
```diff
@@ -75,7 +75,7 @@ public class TabsModel: NSObject, NSCoding {
 
     var currentTab: Tab? {
         let index = currentIndex
-        return tabs[index]
+        return tabs.indices.contains(index) ? tabs[index] : nil
```
dus7, line 78: There's also an extension with `safe` subscript for `RandomAccessCollection` in BSK, but that will work just fine.

dus7 (approved):
Went through tabs management and smoke tested around this. Looks good. Thanks!

PR #3823, opened 2025-01-17
use instance var for TLD in text zoom coordinator

brindy:
Task/Issue URL: https://app.asana.com/0/392891325557410/1209139290185318/f
Tech Design URL:
CC:

**Description**:
Speculative crash fix.  Crashes are coming from init of TLD but there's no need for it to do that so moving to an instance variable.

**Steps to test this PR**:
1. Smoke test setting the text zoom
2. Specifically go to a website, set the zoom level.  Open a new tab and visit the same site.
3. Fire button should clear zoom levels.  Visit the same site and check that it's not applied again.

### DIFF
--- a/DuckDuckGo/TextZoomCoordinator.swift
+++ b/DuckDuckGo/TextZoomCoordinator.swift
@@ -51,7 +51,7 @@ protocol TextZoomCoordinating {
     func onTextZoomChange(applyToWebView webView: WKWebView)
 
     /// Shows a text zoom editor for the current webview. Does nothing if the feature is disabled.
-    func showTextZoomEditor(inController controller: UIViewController, forWebView webView: WKWebView)
+    func showTextZoomEditor(inController controller: UIViewController, forWebView webView: WKWebView) async
 
     /// Creates a browsing menu entry for the given link.  Returns nil if the feature is disabled.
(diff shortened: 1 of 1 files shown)


DuckDuckGo/TextZoomCoordinator.swift
```diff
@@ -51,7 +51,7 @@ protocol TextZoomCoordinating {
     func onTextZoomChange(applyToWebView webView: WKWebView)
 
     /// Shows a text zoom editor for the current webview. Does nothing if the feature is disabled.
-    func showTextZoomEditor(inController controller: UIViewController, forWebView webView: WKWebView)
+    func showTextZoomEditor(inController controller: UIViewController, forWebView webView: WKWebView) async
```
miasma13, line 54: What is the reason for that change?

miasma13 (approved):
LGTM 👍 , tested it and the text zoom works correctly.

One comment for potentially unrelated change.

DuckDuckGo/TextZoomCoordinator.swift
brindy: fix warning about concurrency and isolation that would be an error in swift 6

PR #3816, opened 2025-01-17
update tab manager non selection mode UI

brindy:
Task/Issue URL: https://app.asana.com/0/392891325557410/1208685664909479/f
Tech Design URL:
CC:

**Description**:
Integrates feature flag for tabs multi-selection and update UI when feature is enabled. Edit button doesn't work yet.

**Steps to test this PR**:
1. Smoke test tab switcher on iPhone and iPad when internal user is off. Functionality should be unchanged from production.
2. Check matches Figma on iPhone and and iPad when internal user is on. Add all bookmarks is gone and the selection mode button has been moved, edit button has been added.
3. Check dark / light mode

PR #3832, opened 2025-01-20
tab manager multi select edit menu

brindy:
Task/Issue URL: https://app.asana.com/0/1206767823185684/1207969519536944
Tech Design URL:
CC:

**Description**:
Adds the new tab manager edit menu.  Close all tabs work, but select tabs doesn't yet. 

Depends on #3816 

**Steps to test this PR**:
1. Smoke test the tab manager with internal user on and off
2. Confirm that Close all tabs menu item closes all tabs but does not delete data


DuckDuckGo/UserText.swift
```diff
@@ -217,7 +217,12 @@ public struct UserText {
     
     public static let openHomeTab = NSLocalizedString("tab.open.home", value: "Open home tab", comment: "Accessibility label on tab cell")
     public static let closeHomeTab = NSLocalizedString("tab.close.home", value: "Close home tab", comment: "Accessibility label on remove button")
-    
+
+    public static func closeTabs(withCount count: Int) -> String {
+        let message = NSLocalizedString("tab.close.with.count", value: "Close All %@ Tabs", comment: "Close tabs with a count")
+        return message.format(arguments: "\(count)")
```
dus7, line 224: It'll need to be `Int` argument instead of `String` for plurals to work properly.

DuckDuckGo/TabSwitcherTopBarView.swift
```diff
@@ -41,8 +41,23 @@ struct TabSwitcherTopBarView: View {
...
-        Button {
-            model.onEditPressed()
+        Menu {
+            Button {
+                print("*** select tabs")
+                model.transitionToMultiSelect()
+            } label: {
+                Label("Select tabs", image: "Check-Circle-16")
+            }.onAppear {
+                print("*** select tabs button appeared")
```
dus7: Let's remove `print` statements before merging to main.

dus7 (approved):
Tested with and without internal flag. Two things to consider - can be done in upcoming parts.

DuckDuckGo/UserText.swift
brindy: yep, will fix this up later, thanks, made a note.

PR #3859, opened 2025-01-23
Refactor app startup code - extracting services

jaceklyp:
Task/Issue URL: https://app.asana.com/0/0/1208832732122404/f

**Description**:
- Extract service-related code from the app’s launching logic to improve maintainability and clarity.
- Establish clear patterns for synchronizing crucial services, ensuring consistent behavior during app lifecycle events.
- Design an intuitive API to simplify usage for developers interacting with the app lifecycle code.

**Steps to test this PR**:
1. Go through our smoke tests list.
2. Make sure the authentication and auto-clearing works.
3. Trackers are being blocked.


DuckDuckGo/AppServices/AutofillService.swift
```diff
@@ -0,0 +1,77 @@
...
+
+    func onBackground() {
+        autofillLoginSession.endSession()
+    }
+
+    private func registerForAutofillEnabledChanges() {
+        NotificationCenter.default.addObserver(forName: AppUserDefaults.Notifications.autofillEnabledChange,
+                                               object: nil,
+                                               queue: nil) { _ in
+            self.autofillPixelReporter.updateAutofillEnabledStatus(AppDependencyProvider.shared.appSettings.autofillCredentialsEnabled)
```
dus7: Closure is prolonging the life of `AutofillService`. Is it expected?

DuckDuckGo/AppDelegate.swift
```diff
@@ -33,6 +33,13 @@ import Core
 
     var window: UIWindow?
 
+    override init() {
+        super.init()
+        NotificationCenter.default.addObserver(forName: .databaseDidEncounterInsufficientDiskSpace, object: nil, queue: .main) { _ in
+            self.application(UIApplication.shared, willTerminateWithReason: .insufficientDiskSpace)
```
dus7: AppDelegate is retained here. It should be present throughout app lifecycle regardless but I'd probably advise to not keep it strongly via closure.

DuckDuckGo/AppLifecycle/AppStates/Launching.swift
```diff
@@ -235,19 +235,18 @@ struct Launching: AppState {
...
         }
 
-        unService = UNService(window: window, accountManager: accountManager)
         uiService = UIService(window: window)
 
         // Task handler registration needs to happen before the end of `didFinishLaunching`, otherwise submitting a task can throw an exception.
         // Having both in `didBecomeActive` can sometimes cause the exception when running on a physical device, so registration happens here.
         AppConfigurationFetch.registerBackgroundRefreshTaskHandler()
 
-        UNUserNotificationCenter.current().delegate = unService
```
dus7, line 485: I'm not sure it's the best solution to encapsulate `UserNotificationCenter` inside VPN. Given VPNService conforms to `UNUserNotificationCenterDelegate` there can be `NotificationService` which holds an array of conforming instances, where one of those will be `VPNService`. This way those two are decoupled from each other and we're ready for easier handling of other cases outside of the VPN purpose. It's only a suggestion, since it may increase already quite big scope of the project.

dus7 (commented):
Great effort 💪 ! I'm not approving yet because I haven't smoke tested it and there are several features to test thoroughly, but overall I have no major concerns. I left a few suggestions though.

I think you can also consider `*LifecycleController` naming instead of `*Service` for those newly created objects. In fact they're controlling the encapsulated logic based on the app events, similarly to how `ViewController` controls view based on UI events.

DuckDuckGo/AppConfiguration/PixelConfiguration.swift
```diff
@@ -0,0 +1,71 @@
...
+        Pixel.isDryRun = false
+#endif
+
+        var dryRun = false
+#if DEBUG
+        dryRun = true
+#endif
+        let isPhone = UIDevice.current.userInterfaceIdiom == .phone
+        let source = isPhone ? PixelKit.Source.iOS : PixelKit.Source.iPadOS
+        PixelKit.setUp(dryRun: dryRun,
```
bwaresiak: This looks a bit weird, can we just use `Pixel.isDryRun` here?

DuckDuckGo/AppServices/RemoteMessagingService.swift
```diff
@@ -0,0 +1,80 @@
...
+        BGTaskScheduler.shared.getPendingTaskRequests { tasks in
+            let hasRemoteMessageFetchTask = tasks.contains { $0.identifier == RemoteMessagingClient.Constants.backgroundRefreshTaskIdentifier }
+            if !hasRemoteMessageFetchTask {
+                RemoteMessagingClient.scheduleBackgroundRefreshTask()
+            }
+        }
+    }
+
+    // TODO:  It's public in order to allow refreshing on demand via Debug menu. Otherwise it shouldn't be called from outside.
+    func refreshRemoteMessages() {
```
bwaresiak, line 73: We could add debug_ prefix to these methods.

DuckDuckGo/AppServices/ConfigurationService.swift
```diff
@@ -0,0 +1,83 @@
...
+            if !hasConfigurationTask {
+                AppConfigurationFetch.scheduleBackgroundRefreshTask()
+            }
+        }
+    }
+
+    func onForeground() {
+        if AppConfigurationFetch.shouldScheduleRulesCompilationOnAppLaunch {
+            ContentBlocking.shared.contentBlockingManager.scheduleCompilation() // TODO: should compilation happen in line here
+            // or should other service be responsible for it? e.g. ContentBlockingService
```
bwaresiak: I think this place is correct - after all only background task knows if things were supposedly downloaded

DuckDuckGo/MainCoordinator.swift
```diff
@@ -0,0 +1,271 @@
...
+
+    func shouldProcessDeepLink(_ url: URL) -> Bool {
+        // Ignore deeplinks if onboarding is active
+        // as well as handle email sign-up deep link separately
+        !controller.needsToShowOnboardingIntro() && !handleEmailSignUpDeepLink(url)
+    }
+
+    func onForeground() {
+        controller.showBars()
+        controller.didReturnFromBackground() // TODO: why was it like this?
```
bwaresiak: Can you elaborate the question? :)

jaceklyp: it is called on every foreground, no matter if coming from background or starting app, it was like this in the old app delegate

bwaresiak: From what I recall, yes, we don't want to run the SERP flows for the first interaction with omni bar upon returning to the app.

jaceklyp: isn't the name of this method a bit misleading in context of app transitions? should I rename it to onForeground() or something?

DuckDuckGo/AppConfiguration/PersistentStoresConfiguration.swift
```diff
@@ -0,0 +1,105 @@
...
+                case .some(let error):
+                    Pixel.fire(pixel: .dbInitializationError,
+                               error: error,
+                               withAdditionalParameters: parameters)
+                    if error.isDiskFull {
+                        NotificationCenter.default.post(name: .databaseDidEncounterInsufficientDiskSpace, object: nil)
+                        return
+                    } else {
+                        Thread.sleep(forTimeInterval: 1)
+                        fatalError("Could not create daetagtabase stack: \(error.localizedDescription)")
```
dus7:
```suggestion
                        fatalError("Could not create database stack: \(error.localizedDescription)")
```

DuckDuckGo/MainCoordinator.swift
bwaresiak: Yup, feel free to,

DuckDuckGo/AppConfiguration/PixelConfiguration.swift
```diff
@@ -0,0 +1,71 @@
...
+                       appVersion: AppVersion.shared.versionNumber,
+                       source: source.rawValue,
+                       defaultHeaders: [:],
+                       defaults: UserDefaults(suiteName: "\(Global.groupIdPrefix).app-configuration") ?? UserDefaults()) { (pixelName: String, headers: [String: String], parameters: [String: String], _, _, onComplete: @escaping PixelKit.CompletionBlock) in
+
+            let url = URL.pixelUrl(forPixelNamed: pixelName)
+            let apiHeaders = APIRequestV2.HeadersV2(additionalHeaders: headers)
+            guard let request = APIRequestV2(url: url, method: .get, queryItems: parameters.toQueryItems(), headers: apiHeaders) else {
+                assertionFailure("Invalid request Pixel request")
```
dus7:
```suggestion
                assertionFailure("Invalid Pixel request")
```

DuckDuckGo/AppConfiguration/UserAgentConfiguration.swift
```diff
@@ -0,0 +1,35 @@
...
+import Networking
+import Core
+
+final class UserAgentConfiguration {
+
+    static func configureAPIRequestUserAgent() {
+        APIRequest.Headers.setUserAgent(DefaultUserAgentManager.duckDuckGoUserAgent)
+    }
+
+    // Called at the end of launching due to some race condition in the IPC layer when spawning a WebView for content blocking compilation.
```
dus7, line 30: Wondering if this could be somehow verified by unit tests?

DuckDuckGo/AppLifecycle/AppStates/Background.swift
```diff
@@ -35,97 +31,50 @@ import Core
 ///   - Background tasks, such as saving data or refreshing content, should be handled in this state.
 ///   - Use this state to ensure that the app's current state is saved and any necessary cleanup is performed
 ///     to release resources or prepare for a potential termination.
+@MainActor
```
dus7: This looks weird to me, given the fact it's a Background state. Is there a case where one of the methods of this instance will be called on a background thread? If not I think adding a sentence of doc clarifying that would be beneficial.

DuckDuckGo/AppLifecycle/AppStates/Launching.swift
```diff
@@ -51,612 +38,152 @@ import PixelExperimentKit
...
             let launchTime = CFAbsoluteTimeGetCurrent() - didFinishLaunchingStartTime
             Pixel.fire(pixel: .appDidFinishLaunchingTime(time: Pixel.Event.BucketAggregation(number: launchTime)),
                        withAdditionalParameters: [PixelParameters.time: String(launchTime)])
         }
+        reportingService = ReportingService(fireproofing: fireproofing)
+        KeyboardConfiguration.configure()
+        PixelConfiguration.configure(featureFlagger: featureFlagger)
+        ContentBlockingConfiguration.configure()
+        UserAgentConfiguration.configureAPIRequestUserAgent()
+        NewTabPageIntroMessageConfiguration().configure() // TODO: @Mariusz can it be moved up here?
```
dus7: Yes. It should be done before UI is displayed.

DuckDuckGo/AppLifecycle/AppStates/Launching.swift
```diff
@@ -51,612 +38,152 @@ import PixelExperimentKit
...
 
-        if shouldPresentInsufficientDiskSpaceAlertAndCrash {
-            window = UIWindow(frame: UIScreen.main.bounds)
-            window.rootViewController = BlankSnapshotViewController(addressBarPosition: appSettings.currentAddressBarPosition,
-                                                                    voiceSearchHelper: voiceSearchHelper,
-                                                                    featureFlagger: AppDependencyProvider.shared.featureFlagger)
-            window.makeKeyAndVisible()
-            application.setWindow(window)
+        WidgetCenter.shared.reloadAllTimelines() // TODO: should it be moved to some service?
+        PrivacyFeatures.httpsUpgrade.loadDataAsync() // TODO: should it be moved to some service?
```
dus7: Taking into account our most recent concept I'd say that MOST of the things should be extracted somehow - either to be a service or a configuration. We should optimize to make State objects as simple as possible.
Having said that, transitioning into multiple granular services just for calling a single line will just add boilerplate code, so I think defining a `LaunchService` for such things can be one of alternatives.
If you not decide to do more changes I'd definitely extract all those lines out of init to a separate private method.

DuckDuckGo/AppServices/AutofillService.swift
```diff
@@ -0,0 +1,104 @@
...
+
+final class AutofillService {
+
+    private let autofillLoginSession = AppDependencyProvider.shared.autofillLoginSession
+    private let autofillUsageMonitor = AutofillUsageMonitor()
+    private var autofillPixelReporter: AutofillPixelReporter?
+
+    var syncService: SyncService?
+
+    init() { // TODO: this was always done after DB was initialized, so looks like this code was never being executed, to confirm
```
dus7: Since it's a part of a protocol, I'd assume the bool property (`isDatabaseFileInitialized`) can be false in some corner cases, so I'd say it's reasonable to have such check.

DuckDuckGo/AppServices/AutofillService.swift
```diff
@@ -0,0 +1,104 @@
...
+                    break
+                }
+            },
+            installDate: StatisticsUserDefaults().installDate ?? Date()
+        )
+    }
+
+    func onForeground() {
+        guard let syncService else {
+            assertionFailure("SyncService must be injected before calling onForeground.")
```
dus7, line 78: I think we need a pixel for this case.

DuckDuckGo/Favicons.swift
```diff
@@ -67,34 +64,6 @@ public class Favicons {
         NotificationCenter.default.removeObserver(Constants.tabsCache)
     }
 
-    public func migrateFavicons(to size: CGSize, afterMigrationHandler: @escaping () -> Void) {
```
dus7, line 70: It was agreed to remove this, right?

jaceklyp: yes

jaceklyp: see: https://app.asana.com/0/0/1209225275889527/f

DuckDuckGo/AppServices/AutofillService.swift
jaceklyp: turns out it can be removed too:
https://app.asana.com/0/0/1209194443865280/f

DuckDuckGo/AppConfiguration/UserAgentConfiguration.swift
jaceklyp: Outside of scope in this project

DuckDuckGo/AppLifecycle/AppStates/Launching.swift
jaceklyp: I agree, however, I don't want to build a more generic solution for this in this project. If you don't mind I'd leave it in VPN, and leave to a person who wants to start displaying UNNotifications.

DuckDuckGo/AppServices/AutofillService.swift
jaceklyp: this service lives throughout the whole app lifecycle, I don't mind capturing it here

DuckDuckGo/AppServices/AutofillService.swift
dus7: We discussed this with @jaceklyp and I agree it's not needed. The assertion here is enough.

dus7 (approved):
Solid work! 💪

PR #3866, opened 2025-01-24
Duckplayer: Show video in Full Screen when rotating the phone

afterxleep:
Task/Issue URL: https://app.asana.com/0/1142021229838617/1209253933548230/f
Tech Design URL:
CC:

**Description**:
Renders the video in “Full Screen” when rotating your phone to landscape.

Given that there’s no straightforward way to render a Youtube Video in Full screen without user interaction this emulates it by hiding the Browser Chrome when you rotate the phone.

**Steps to test this PR**:
1. Find a video and watch in Duck Playe
2. Rotate to landscape
3. Confirm the Browser Chrome is hidden
4. Tap the screen
5. Confirm the Browser Chrome appears
6. Wait 4 seconds, and confirm the browser chrome hides again.
7. Rotate to Portrait mode
8. Confirm Browser Chrome is always visible.
9. Recommended:  Test with both URL bar at the top and at the bottom

<!—
Before submitting a PR, please ensure you have tested the combinations you expect the reviewer to test, then delete configurations you *know* do not need explicit testing.

Using a simulator where a physical device is unavailable is acceptable.
—>

* [X] Does this PR satisfy our [Definition of Done](https://app.asana.com/0/1202500774821704/1207634633537039/f)?

---
  - https://app.asana.com/0/0/1209253933548230
  - https://app.asana.com/0/0/1209272240764363


THISISDINOSAUR:
In testing it mostly works, but I did manage to find a case where the chrome doesn't disappear. If I pull down the Lock Screen (either fully or partially) the chrome never goes 

https://github.com/user-attachments/assets/7523873c-d5f0-4e72-9663-dce4479f5ba9

DuckDuckGo/BrowserChromeManager.swift
```diff
@@ -36,6 +36,16 @@ protocol BrowserChromeDelegate: AnyObject {
     var tabBarContainer: UIView { get }
 }
 
+extension BrowserChromeDelegate {
+    func setBarsHidden(_ hidden: Bool, animated: Bool) {
```
THISISDINOSAUR: Apologies for my lack of context on the iOS codebase at this point, but what's the reason for this extension versus just giving the original function default values? 
(`func setBarsHidden(_ hidden: Bool, animated: Bool, animationDuration: CGFloat? = nil)`)

DuckDuckGo/DuckPlayer/DuckPlayer.swift
```diff
@@ -217,13 +217,16 @@ final class DuckPlayer: DuckPlayerControlling {
...
+        static let chromeShowHideAnimationDuration: CGFloat = 0.4
     }
     
     
     private(set) var settings: DuckPlayerSettings
-    private(set) weak var hostView: UIViewController?
+    private(set) weak var hostView: TabViewController?
     
     private var featureFlagger: FeatureFlagger
+    private var hideTimer: Timer?
```
THISISDINOSAUR: I think this would benefit from a more specific name. What's it hiding? (obviously I know because of the context of the PR, but just from looking at it I wouldn't)

DuckDuckGo/DuckPlayer/DuckPlayer.swift
```diff
@@ -255,20 +258,66 @@ final class DuckPlayer: DuckPlayerControlling {
...
+        tapGesture.delegate = self
+        vc.view.addGestureRecognizer(tapGesture)
     }
     
     /// Removes the host view controller.
     public func removeHostView() {
+        // Remove the tap gesture recognizer
+        hostView?.view.gestureRecognizers?.forEach { recognizer in
+            if recognizer is UITapGestureRecognizer {
+                hostView?.view.removeGestureRecognizer(recognizer)
```
THISISDINOSAUR: Could this end up making life harder if we ever want a tap recognizer for something else on the TabViewController?

DuckDuckGo/DuckPlayer/DuckPlayer.swift
```diff
@@ -371,7 +474,7 @@ final class DuckPlayer: DuckPlayerControlling {
         public func telemetryEvent(params: Any, message: WKScriptMessage) async -> Encodable? {
             // Not currently accepting any telemetry events
             return nil
```
THISISDINOSAUR: Looks like the rest of the indentation needs to change too

DuckDuckGo/DuckPlayer/DuckPlayer.swift
```diff
@@ -489,3 +591,78 @@ final class DuckPlayer: DuckPlayerControlling {
...
+        let scrollViewBackground: UIColor?
+    }
+    
+    // Key for associated object
+    private static var originalSettingsKey: UInt8 = 0
+    
+    // Getter/Setter for original settings using associated objects
+    private var originalSettings: OriginalSettings? {
+        get {
+            return objc_getAssociatedObject(self, &Self.originalSettingsKey) as? OriginalSettings
```
THISISDINOSAUR: I'm struggling to follow the purpose other this. Is there a reason besides avoiding having a stored param on TabViewController?

DuckDuckGo/BrowserChromeManager.swift
afterxleep: Yea, we could do that.  Just went for an extension as it gives us more flexibility to add shared behavior later and keeps the protocol interface cleaner.

DuckDuckGo/DuckPlayer/DuckPlayer.swift
afterxleep: Yea, good point. I updated it so the gesture recognizer is only added when you’re in DuckPlayer and rotate the phone.  Also updated removal to be explicit, instead of wiping out everything.

DuckDuckGo/DuckPlayer/DuckPlayer.swift
afterxleep: Decided to do it here as this is exclusive to DuckPlayer.-  A stored param in Tabviewcontroller will continue to couple things further

afterxleep:
> In testing it mostly works, but I did manage to find a case where the chrome doesn't disappear. If I pull down the Lock Screen (either fully or partially) the chrome never goes

@THISISDINOSAUR this is likely because the app is pushed to the background when you do, and timers are cancelled.   We could restart the timers again when apps move to foreground but that’d require some effort as DuckPlayer is not aware of backgrounding or anything like that.

If you tap the screen again after coming back, the timers will be restarted and controls will go away as expected.   I think we can live with that.

DuckDuckGo/DuckPlayer/DuckPlayer.swift
THISISDINOSAUR: This seems like quite the act of coupling to me already, for me I'm not sure this is a worthwhile evil. I see TabViewController already has two DuckPlayer related properties, so two thoughts:
- Perhaps the settings could be saved on DuckPlayer or DuckPlayerNavigationHandler instead
- We consolidate things so TabViewController only needs to know about a single managing object

DuckDuckGo/BrowserChromeManager.swift
THISISDINOSAUR: I'm not sure I agree it makes the protocol interface cleaner, to me this feels like a layer of indirection confusing things. If we want it to be possible to specify no parameter, why not use the language feature for it?

It's also going to be hard to know which to call in future if they both do different things, so I don't think we should use this to add a default behavior (particularly since it's speculative). (e.g. we add some stuff to `setBarsHidden(_ hidden: Bool, animated: Bool)`, but someone comes along and innocently calls the more visible `func setBarsVisibility(_ percent: CGFloat, animated: Bool, animationDuration: CGFloat?)`)

afterxleep: Not sure I follow.  

`BrowserChromeDelegate` is a protocol and I can't use default parameters, so adding the extension is the cleanest way to preserve backwards compatibility with existing code and keep existing usage clean.

Otherwise, I'd have to replace all occurrences of `func setBarsHidden(_ hidden: Bool, animated: Bool)` to add the animation param and pass nil, which would be noisy and unnecessary.

DuckDuckGo/DuckPlayer/DuckPlayer.swift
afterxleep: Just to clarify, your proposal would be to add a prop to  TabviewController and move the settings ‘saving/restoring’ there?

I don’t see shy this is coupling.  This is an extension of TabViewController itself.  The only thing is that I’ve added it to a DuckPlayer file, because it’s the only usage ocurrence.

DuckPlayerNavigationHandler is not aware of ViewControllers (and shouldn't be) - it only knows about the webView. Moving this storage to DuckPlayer or DuckPlayerNavigationHandler isn't the best option since DuckPlayerNavigationHandler is view/nav focused and DuckPlayer shouldn't store properties about TabVC internally.

So we have two options:
1. Add these properties/methods directly to TabViewController (which is already quite large and bloated)
2. Keep them in a private extension in DuckPlayer (the only component that uses them)

I went with #2 since it:
- Keeps the code with the only component that uses it
- Avoids bloating TabViewController
- Uses associated objects to handle temporary state without need to use stored properties
- Maintains proper separation of concerns

DuckDuckGo/BrowserChromeManager.swift
THISISDINOSAUR: Sorry was trying to do too much stuff yesterday, ignore my first sentence. 

I am opposed to us ending up with two different functions someone could call and one of them being the wrong one, so I really wouldn't want us to add shared behavior in this extension, but given that there isn't any shared behavior there at the moment I'm willing to let it go.

That said, there aren't that many call sites, so I'm not sure it's worth making any decisions based on avoiding updating them

DuckDuckGo/DuckPlayer/DuckPlayer.swift
THISISDINOSAUR: I don't have a specific proposal I'm particularly attached to, I'd just rather we do almost anything else.

IMO the bar to adding this needs to be quite high, we don't use objc_getAssociatedObject anywhere else in iOS and it has quite the downsides along with obviously not being very swifty. At the very least I think the comment(s) should be more informative as to the why, at the moment the comments basically just say what the code already says, which is really only helpful to someone that has no understanding of Swift

DuckDuckGo/BrowserChromeManager.swift
afterxleep: fine, I’ll update the calls

DuckDuckGo/DuckPlayer/DuckPlayer.swift
afterxleep: I don’t really agree.  This is just a way to move this (only used by DuckPlayer) out of TabViewController, (which is already bloated).  The fact that we don’t use it does not mean we shouldn’t, but if that is a concern or is not clear enough, I’ll rather make the changes and move to TabVC props.

afterxleep:
@THISISDINOSAUR I’ve moved the landscape/portrait customization to TabViewController and removed the protocol extension, adding a `customAnimationDuration` param to all function calls.

THISISDINOSAUR (commented):
I don't know if this is expected behavior, but if I start in landscape when I load a video, the chrome never disappears, is this expected?

afterxleep:
> I don't know if this is expected behavior, but if I start in landscape when I load a video, the chrome never disappears, is this expected?

Yes, when rotating from portrait to landscape the reason is usually you want the video in full screen, but if you’re starting in landscape, you may not necesarily want that, especially considering the native side does not know if the video is playing or not.

Same as in portrait, in this case you can easily use the player’s full screen button.

THISISDINOSAUR (approved):

PR #3894, opened 2025-01-29
AI Chat debug improvement

Bunn:
Task/Issue URL: https://app.asana.com/0/1204167627774280/1209263437216309/f

**Description**:
Bypass delegate logic from AIChatWebViewController in case we have a custom host settings in the debug UI.
This was done to allow FE devs to use custom hosts while debugging without having to use Xcode

**Steps to test this PR**:
1. There are no behavior change for normal users.
2. Make sure you’re set as internal
3. Search something, then tap in the AI Chat button
4. From the AI Chat view, tap on the top right setting button then Help Pages. The View should be dismissed and the help pages opened in a tab
5. Go to settings -> Debug -> AI Chat -> Type anything in the host field
6. Do the same step as # 4, this time nothing should happen. No page is loaded because inside AI Chat every link goes to _blank, you can validate this by adding this in the decide policy method:

```
if navigationAction.targetFrame == nil {
            webView.load(navigationAction.request)
            return .cancel
        }

```
7. Go back to the AI Chat debug menu, reset the data, and try again. It should do the same as step # 4


DuckDuckGo-iOS.xcodeproj/project.pbxproj
```diff
@@ -1566,7 +1565,6 @@
 		311BD1AE2836BB4200AEF6C1 /* AutofillItemsLockedView.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = AutofillItemsLockedView.swift; sourceTree = "<group>"; };
 		311BD1B02836C0CA00AEF6C1 /* AutofillLoginListAuthenticator.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = AutofillLoginListAuthenticator.swift; sourceTree = "<group>"; };
 		31206F6F2D3804E800A95D76 /* AIChatDebugView.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = AIChatDebugView.swift; sourceTree = "<group>"; };
-		31206F712D38071000A95D76 /* AIChatDebugSettingsHandling.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = AIChatDebugSettingsHandling.swift; sourceTree = "<group>"; };
```
Bunn, line 1572: File moved to the AI Chat package.

jotaemepereira (approved):
Works as expected, and great job adding the `AIChatRequestAuthorizationHandlerTests` tests 👏🏼

PR #3918, opened 2025-02-03
Update setup script for latest devices

Bunn:
Task/Issue URL: https://app.asana.com/0/1203301625297703/1209305186837554/f

**Description**:
Update setup script for latest devices

**Steps to test this PR**:
1. Run setup script, check if it’s working
2. Run release maestro tests, check if they are passing with the new maestro version

### DIFF
--- a/.maestro/setup_ui_tests.sh
+++ b/.maestro/setup_ui_tests.sh
@@ -11,8 +11,9 @@ target_device="iPhone-16"
 target_os="iOS-18-2"
 
 # Convert the target_device and target_os to the format required by the -destination flag
-destination_device="${target_device//-/' '}"
-destination_os="${target_os//-/.}"
+destination_device="${target_device//-/ }"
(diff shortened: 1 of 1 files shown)


.maestro/setup_ui_tests.sh
```diff
@@ -8,14 +8,19 @@ source $(dirname $0)/common.sh
 
 # The simulator command requires the hyphens
 target_device="iPhone-16"
-target_os="iOS-18-1"
+target_os="iOS-18-2"
```
loremattei, line 11: FWIW, we can't select the minor version of iOS on Maestro Cloud, so we may end up being out of sync between local and CI. 
Anyway, this can happen in any case given the limited control we have, so I think it's good to keep selecting the minor version locally. It will help for consistency in out local development environment. We should just be aware we can't expect that consistency with cloud.

loremattei (approved):
LGTM!
Thanks @Bunn !

PR #3921, opened 2025-02-03
 tab manager multi select mode

brindy:
Task/Issue URL: https://app.asana.com/0/392891325557410/1209303765372372/f
Tech Design URL:
CC:

**Description**:
Adds UI and menus for multi-tab select.

Minor deviation from Figma - long press menu items are only specific to the current tap and make more sense.

**Steps to test this PR**:

Phone:

1. Internal user off
3. Go to tab switcher ensure it looks like production 
4. Internal user on
5. Open various combinations of new tabs / web pages, single, and multiple.
8. Check that long pressing a tab and choosing select tab enters select mode and selects the tab
9. Tap anywhere on screen to exit select mode
10. Tap anywhere on screen again to go back to previous screen
6. Back in tab switcher tap edit and check menu items
7. Use Select Tabs to enter select mode
8. Check bottom menu and button are enabled when appropriate and show correct menu items
9. Check long press menu items are appropriate

Repeat test on iPad:

1. Bottom bar (and therefore menu) doesn't exist, check Figma for iPad details


DuckDuckGo/TabSwitcherViewController+MultiSelect.swift
```diff
@@ -345,8 +251,15 @@ extension TabSwitcherViewController {
         }.count
     }
 
+    var allPagesCount: Int {
+        return tabsModel.tabs.compactMap {
+            $0.link
+        }.count
```
dus7: You could also do a tidy one-liner
```suggestion
        tabsModel.tabs.compactMap(\.link).count
```
and skip `return`.

DuckDuckGo/TabSwitcherViewController+MultiSelect.swift
```diff
@@ -630,6 +510,18 @@ extension TabSwitcherViewController {
 // MARK: UIAction factories
 extension TabSwitcherViewController {
 
+    func action(title: String, _ handler: @escaping () -> Void) -> UIAction {
+        return UIAction(title: title) { _ in
+            handler()
```
dus7, line 583: Using `handler` here causes strong retain cycle, leaving instances of `TabSwitcherViewController` and related classes in memory. You should capture weak self in all usage cases of `action` function.

dus7 (commented):
Good job! Works as advertised. I checked both iPhone and iPad. Leaving one suggestion and one place to fix which is causing retain cycles.

dus7 (approved):
Looks good! Checked described scenarios and verified there are no retain cycles.

PR #3933, opened 2025-02-06
[Draft] shared web tests

jonathanKingston:
Task/Issue URL: https://app.asana.com/0/72649045549333/1208763070373634
Tech Design URL:
CC:

**Description**:

**Draft as dependent repo isn't open yet.**
Code for that is here: https://github.com/duckduckgo/shared-web-tests/compare/main...jkt/webdriver

**Steps to test this PR**:
1.
2.

**Copy Testing**:

**Orientation Testing**:

* [ ] Portrait
* [ ] Landscape

**Device Testing**:

* [ ] iPhone SE (1st Gen)
* [ ] iPhone 8
* [ ] iPhone X
* [ ] iPhone 14 Pro
* [ ] iPad

**OS Testing**:

* [ ] iOS 15
* [ ] iOS 16
* [ ] iOS 17

**Theme Testing**:

* [ ] Light theme
* [ ] Dark theme

---
###### Internal references:


DuckDuckGo/MainViewController.swift
```diff
@@ -502,8 +502,7 @@ class MainViewController: UIViewController {
     }
     
     func startOnboardingFlowIfNotSeenBefore() {
-        
-        guard ProcessInfo.processInfo.environment["ONBOARDING"] != "false" else {
```
jonathanKingston, line 512: @brindy  Just going to validate if we need this, iirc I think it's not actually used properly. If it is needed I think my preference would be to move it into launchOptionsHandler code anyway to turn it into an `or` statement.

brindy: [spam message deleted]

DuckDuckGo/AppLifecycle/AppStates/Launching.swift
```diff
@@ -650,6 +652,13 @@ struct Launching: AppState {
                                dataStoreIDManager: dataStoreIDManager)
     }
 
+    private func startAutomationServer() {
```
brindy: Please call this `startAutomationServerIfNeeded`

DuckDuckGo/AutomationServer.swift
```diff
@@ -0,0 +1,340 @@
...
+    func receive(from connection: NWConnection) {
+        connection.receive(
+            minimumIncompleteLength: 1,
+            maximumLength: connection.maximumDatagramSize
+        ) { content, _, isComplete, error in
+            switch connection.state {
+            case .ready:
+                break // Connection is valid, continue
+            case .cancelled, .failed:
+                print("Connection is no longer valid \(connection.state) \(String(describing: error)) \(String(describing: content)).")
```
brindy, line 57: Please replace this print with logging (other places too)

DuckDuckGo/MainViewController+Automation.swift
```diff
@@ -0,0 +1,34 @@
...
+//
+
+import UIKit
+
+extension MainViewController {
+    
+    func executeScript(_ javaScriptString: String,
+                       args: [String: Any] = [:]) async -> Result<Any, any Error>? {
+        var result = await currentTab?.executeScript(javaScriptString, args: args)
+        return result! // TODO fix !
```
brindy, line 27: Please fix :)

.maestro/setup_ui_tests.sh
```diff
@@ -4,17 +4,6 @@
 
 source $(dirname $0)/common.sh
 
-## Constants
```
brindy, line 7: Did you mean to delete all this or was it temporary?  Likewise for the build function below.

DuckDuckGo/MainViewController.swift
brindy: I think it's OK to remove this yeah, looks like LaunchOptionsHandler is the correct approach

DuckDuckGo/TabManager.swift
```diff
@@ -162,7 +162,7 @@ class TabManager {
         }
     }
     
-    private func controller(for tab: Tab) -> TabViewController? {
+    public func controller(for tab: Tab) -> TabViewController? {
```
brindy: Just removing `private` should be enough for what you need. Public is only necessary if you want to access it from a different package.

DuckDuckGo/AutomationServer.swift
```diff
@@ -0,0 +1,340 @@
...
+                    Logger.automationServer.error("Invalid URL: \(match.2)")
+                    return // Or handle the error appropriately
+                }
+                switch url.path {
+                case "/navigate":
+                    self.navigate(on: connection, url: url)
+                case "/execute":
+                    self.execute(on: connection, url: url)
+                case "/getUrl":
+                    self.respond(on: connection, response: self.main.currentUrl() ?? "")
```
brindy: Since this the only place this is used I'd consider just accessing the webView url directly.

    var url = self.main.currentTab?.webView.url?.absoluteString
    self.respond(on: connection, response: url ?? "")

Then remove those accessors you added.  Mainly because they will be visible to other code and might cause confusion about what the URL is the code should be using.

DuckDuckGo/AutomationServer.swift
```diff
@@ -0,0 +1,340 @@
...
+        Logger.automationServer.info("Handling request!")
+        let stringContent = String(decoding: content, as: UTF8.self)
+        // Log first line of string:
+        if let firstLine = stringContent.components(separatedBy: CharacterSet.newlines).first {
+            Logger.automationServer.info("First line: \(firstLine)")
+        }
+
+        // Get url parameter from path
+        // GET / HTTP/1.1
+        if #available(iOS 16.0, *) {
```
brindy: Rather than nested if elses... use a guard and return faster.

In this case
    guard #available(iOS 16.0, *) else { return }

This will remove 'pyramid of doom'

DuckDuckGo/AutomationServer.swift
```diff
@@ -0,0 +1,340 @@
...
+                    self.switchToWindow(on: connection, url: url)
+                case "/newWindow":
+                    self.newWindow(on: connection, url: url)
+                case "/getWindowHandle":
+                    self.getWindowHandle(on: connection, url: url)
+                default:
+                    self.respondError(on: connection, error: "unknown")
+                }
+            } else {
+                self.respondError(on: connection, error: "unknown method")
```
brindy: Change corresponding if to a guard.

DuckDuckGo/AutomationServer.swift
```diff
@@ -0,0 +1,340 @@
...
+            }
+        } else {
+            self.respondError(on: connection, error: "unhandled")
```
brindy: Change corresponding if to a guard

DuckDuckGo/AutomationServer.swift
```diff
@@ -0,0 +1,340 @@
...
+                    let jsonDecoder = JSONDecoder()
+                    args = try jsonDecoder.decode([String: String].self, from: argsData)
+                } catch {
+                    self.respondError(on: connection, error: error.localizedDescription)
+                }
+            } else {
+                self.respondError(on: connection, error: "Unable to decode args")
+            }
+        }
+        Task {
```
brindy, line 178: This task will still run even if you've responded error above - is that intentional?  If not, change the above code to use guards and return out.

DuckDuckGo/AutomationServer.swift
```diff
@@ -0,0 +1,340 @@
...
+//
+import Foundation
+import Network
+
+extension Logger {
+    static var automationServer = { Logger(subsystem: Bundle.main.bundleIdentifier ?? "DuckDuckGo", category: "Automation Server") }()
+}
+
+
+class AutomationServer {
```
brindy: Please make this final (surprised there wasn't a SwiftLint tbh)

DuckDuckGo/AutomationServer.swift
```diff
@@ -0,0 +1,340 @@
...
+//  limitations under the License.
+//
+import Foundation
+import Network
+
+extension Logger {
+    static var automationServer = { Logger(subsystem: Bundle.main.bundleIdentifier ?? "DuckDuckGo", category: "Automation Server") }()
+}
+
+
```
brindy, line 26: Generally - can you try and flatten the if else pyramids where possible using guards, please?  I added a few comments to this effect already.

DuckDuckGo/AutomationServer.swift
```diff
@@ -0,0 +1,340 @@
...
+    @MainActor
+    func newWindow(on connection: NWConnection, url: URLComponents) {
+        self.main.newTab()
+        let handle = self.main.tabManager.current(createIfNeeded: true)
+        guard let handle else {
+            self.respondError(on: connection, error: "no window")
+            return
+        }
+        // Response {handle: "", type: "tab"}
+        let response: [String: String] = ["handle": String(UInt(bitPattern: ObjectIdentifier(handle))), "type": "tab"]
```
brindy: Do you have to use an ObjectIdentifier?  Tabs already have a UUID allocated to them. You can access it like

`tab.tabModel.uid`

DuckDuckGo/AutomationServer.swift
```diff
@@ -0,0 +1,340 @@
...
+                )
+            }
+        } catch {
+            Logger.automationServer.error("Got error encoding JSON: \(error)")
+        }
+    }
+    
+    @MainActor
+    func handleConnection(_ connection: NWConnection) {
+        // connection.start(queue: .global())
```
brindy: stray comment

DuckDuckGo/MainViewController+Automation.swift
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    func currentUrl() -> String? {
```
brindy: As commented elsewhere remove this, just access the webView url directly in the one place you need it

DuckDuckGo/AutomationServer.swift
jonathanKingston: I'd missed this, that's great thank you! So much simpler!

